PR #15, opened 2017-09-25
Added ca translation

joamuran:
Just added catalan/valencian translation tyo translations folder.


lupoDharkael:
Thank you for your help!
Looks good!

PR #31, opened 2017-12-14
Added first version of bash_completion for flameshot

cmolinaord:
I wrote a bash completion to autocomplete with the basic commands when using flameshot from a bash terminal. I would like you to test it and do some corrections if needed.
Regards,
Carlos.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1,0 +1,59 @@
+# bash-completion for flameshot command
+# To be installed in "/usr/share/bash-completion/completions/flameshot"
+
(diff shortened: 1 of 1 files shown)


lupoDharkael:
Seems to be ok.
Thanks for the contribution!

cmolinaord:
Thank you! Now you'll have to include this file in the package to be installed properly in the directory for bash completions:
`/usr/share/bash-completion/completions/flameshot`

PR #40, opened 2017-12-26
AppImage package & Github Release

ZetaoYang:
Please set environment variable GITHUB_OAUTH_TOKEN on travis-ci project setting page for Github release and github oauth token:
flameshot-2017-12-26-22 36 18


lupoDharkael:
Why do you need the token?

ManuelLR:
Using the token like this is insecure because anyone can make a pull request where he has modified the travis file and can do whatever he wants with the private token

ZetaoYang:
@ManuelLR Some environment variables (such as usernames and passwords) can not be made public, and are available on the Travis website at each repository's settings page, and Travis automatically adds them to environment variables. As a result, these environment variables can still be used inside the script, but **only the administrator can see the value of the variable**. Please see https://docs.travis-ci.com/user/environment-variables .

@lupoDharkael need token to deploy AppImage to github release. When `git tag` and `push`, it will be auto deployed to github release. like this:

image

ManuelLR:
@ZetaoYang Yes, I know. I tried to reproduce the attack but I can because travis say:

> Travis CI makes encrypted variables and data available only to pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them.
Pull requests sent from forked repositories do not have access to encrypted variables or data.

Source: https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

And that is the `travis.yml` file that I try (and you can reproduce in you local repository):
```yaml
os: linux
dist: trusty
addons:
  apt:
    packages:
      - gnupg
script:
  # Download and save my gpg public key
  - gpg --keyserver pgp.mit.edu --recv-key 0x89c3a1d8c4421d67
  # Save your token to a file
  - echo "Eii your personal github token is '${GITHUB_OAUTH_TOKEN}'" > your_token.txt
  # Encrypt the file with your token to prevent make public to everyone
  - gpg --trust-model always --output your_token.gpg --encrypt --recipient manuellr.git@gmail.com your_token.txt
  # Transfer to be accessible
  - curl --upload-file ./your_token.gpg https://transfer.sh/your_token.gpg
```

But anyway, there is a more logical and entendible way, I refer to https://docs.travis-ci.com/user/encryption-keys
Alike they say:
> Please note that encrypted environment variables are not available for pull requests from forks.

So both are secure (aparently) but I think that If we use `secure` instance of variables configurable in travis would be more readable and understandable.

ZetaoYang:
@ManuelLR  How is Encrypted token instead of original token？

ManuelLR:
The encrypted token must be encrypted in your local machine with the travis app. The easy way is generate an oauth token as indicated [here](https://docs.travis-ci.com/user/deployment/releases/#Authenticating-with-an-OAuth-token) that is the recommended option.

You can view a example of the final result here:
https://github.com/AgoraUS-G1-1617/Deliberations/blob/master/.travis.yml#L26-L37

ZetaoYang:
@lupoDharkael using  encrypted token  instead of original token is OK. here is [tutorial](http://lthr.io/deploy-to-gh-pages).

ZetaoYang:
@ManuelLR Could you help us make the .deb package in travis-ci environment？

ManuelLR:
@ZetaoYang yes, of course but first I'm improving the .travis.yml file to support new features and make it more understandable.

Take a look in my master branch of the project and feel free to comment anything https://github.com/ManuelLR/flameshot/blob/master/.travis.yml

lupoDharkael:
I like @ManuelLR's travis as it's structured and concise and reuses some concepts of @ZetaoYang's travis.

I think it should be done this way:
- Start with @ManuelLR's travis as a base. 
- Update the ./debian directory
- Generate the AppImage as @probonopd pointed here https://github.com/lupoDharkael/flameshot/issues/39#issuecomment-353594921

I think it would be important to distribute signed appImages or binaries too, but first lets get a consistent base. Another important aspect of @ManuelLR's travis is that he uses the min compiler and QT versions  supported by Flameshot so every build will tell if I'm breaking the build in Debian 8 (as that's the oldest target system I'm supporting)

PR #42, opened 2017-12-30
General improvements of travis

ManuelLR:
The main goals of this change are:
- Improve code comprehensibility
- Remove unnecessary dependencies
- Improve design (with stages inclusion) to make easily to maintain and adding new "packaging" stages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,183 +1,190 @@
-language: cpp
-#speed compile process
-cache: ccache 
-compiler: gcc
-os:
-  - linux
-#sudo: required
+os: linux
+sudo: false
 dist: trusty
-#Ubuntu Trusty 14.04 (LTS)
-#Xenial still not out in 2017!
+stages:
+  - Build & Test
+  - Packaging
+  - name: Deploy
+#    if: branch = master
+ 
 env:
   global:
-    - QT_VERSION_MAJOR=5
-    - TRAVIS_DIST_NAME=trusty
-#    - CLANG_VERSION=4.0
-  matrix:
-#    - QT_VERSION_MINOR=8 QT_VERSION_REVISION=0
-#    - QT_VERSION_MINOR=9 QT_VERSION_REVISION=1
-    - QT_VERSION_MINOR=9 QT_VERSION_REVISION=3
-addons:
-  apt:
-    sources:
-      - ubuntu-toolchain-r-test
-      #clang4
-#      - llvm-toolchain-trusty-4.0
-#      - sourceline: 'ppa:beineri/opt-qt58-trusty'
-#      - sourceline: 'ppa:beineri/opt-qt591-trusty'
-      - sourceline: 'ppa:beineri/opt-qt593-trusty'
-    packages:
-###################################### GCC
-      - g++-6
-      - gcc-6
-###################################### Clang
-#      - clang-4.0
-###################################### Qt5
-      - qt593d
-      - qt59base
-      - qt59canvas3d
-      - qt59connectivity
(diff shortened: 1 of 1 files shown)


ManuelLR:
I would like to know the opinion of @ZetaoYang and I appreciate and I'd be very grateful if someone would check the AppImage.

ManuelLR:
I recommend creating a script that automatically compiles the AppImage for the future and use it in the `.travis.yml`

.travis.yml
```diff
@@ -1,183 +1,190 @@
...
+        ##### Clean cache folder
+        - rm -rf $DEB_DST_PATH/*
+
+      script:
+        - ls -alhR $BUILD_DST_PATH
+        - echo "Do .deb" > $DEB_DST_PATH/dev.txt
+        - ls -alhR $DEB_DST_PATH
+
+      after_script:
+        #### Wait preventing cache doesn't update
```
lupoDharkael: Wait preventing a cache update

.travis.yml
```diff
@@ -1,183 +1,190 @@
...
+        #### Packaging
+        - ./squashfs-root/usr/bin/linuxdeployqt appdir/flameshot -verbose=2 -bundle-non-qt-libs
+        - ./squashfs-root/usr/bin/linuxdeployqt appdir/flameshot -verbose=2 -appimage
+        #- ./squashfs-root/usr/bin/linuxdeployqt appdir/flameshot -verbose=2 -appimage -no-translations
+        - ls -alhR ./*.AppImage
+        - cp *.AppImage $APPIMAGE_DST_PATH/
+
+      after_script:
+        #### Wait preventing cache doesn't update
+        - sleep 10
```
lupoDharkael: isn't enough a shorter sleep?

.travis.yml
```diff
@@ -1,183 +1,190 @@
...
+      script:
+        #### Prepare packing
+        - mkdir appdir
+        - cp $BUILD_DST_PATH/flameshot appdir/
+        - cp docs/desktopEntry/package/* appdir/
+        - cp img/flameshot.png appdir/
+        - ls -alhR appdir/
+
+        #### Configure env vars
+        - unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
```
lupoDharkael: You can do `unset QTDIR QT_PLUGIN_PATH  LD_LIBRARY_PATH`

.travis.yml
ManuelLR: I don't know what is the correct sleep time to prevent cache errors. I use this arbitrary value that can be adapted according to our feed back. If you prefer i can reduce or omit it but i think that isn't significant (10s * 2 = 20s of 10m aprox 3% of total time)

.travis.yml
ManuelLR: I will modify it. Actually I don't remember why i put that, i'm going to try to omit it or comment why i execute it

.travis.yml
ManuelLR: I'm going to put `Wait for the cache to update`

ZetaoYang:
@ManuelLR Why you put `doc` folder into appimage? It is not needed.
image

ManuelLR:
I didn't put that folder, It has been created automatically when the AppImage was generated. Only contains the copyright of the libraries so the folder is needed.

ManuelLR:
Otherwise, we must include the flameshot copyright, right @lupoDharkael ?

lupoDharkael:
Seems ok for me as an initial step towards improvement.

PR #44, opened 2018-01-04
.deb package 

ZetaoYang:
Because the current travis environment is Ubuntu 14.04, it supports Qt version 5.2 (`qt5-qmake`  `qt5-default`  `qttools5-dev-tools` , more details, look at this https://pkgs.org/download/qt-default ) does not meet the minimum compile requirements, if compiled, there will be two errors about "New Signal Slot Syntax". But the good news is that travis is working on Ubuntu 16.04. What we have to do for the deb package is to wait for travis to release the 16.04 image.
I made changes can be used in the 16.04 environment, but now commented out.

P.S.  I want to achieve “pin capture to desktop/paste images as topmost floating windows” and already have ideas on how to achieve it. I will take the time to complete this work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 os: linux
 sudo: false
 dist: trusty
+#dist: xenial  ### wait ubuntu 16.04's coming from travis-ci
 stages:
   - Build & Test
   - Packaging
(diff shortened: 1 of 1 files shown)


ManuelLR:
Hi @ZetaoYang, I promised I'm going to help automatize the .deb package so I'm here. I think that a good approach would be create a bash script that run in debian and create the deb package.
I think that would be a good option do the same (create a bash script) to automatize appImage creation.

PR #54, opened 2018-01-05
Update README.md

shibby:
Avoid forced installation for dependencies before compiling.

Since required packages are for development, it is good idea to see what will be installed to our system. For instance, it has tried to install 49 new package into my system.  
screenshot from 2018-01-05 17-02-45

I suggest to remove `-y` parameters and than everyone can see what will be installed into system.


lupoDharkael:
Seems reasonable

PR #57, opened 2018-01-05
Make packaging for Flatpak easier

Alexander-Wilms:
This would make the patch at https://github.com/Alexander-Wilms/flathub/tree/new-pr superfluous.

### DIFF
--- a/flameshot.pro
+++ b/flameshot.pro
@@ -198,36 +198,38 @@ RESOURCES += \
 
 # installs
 unix:!macx {
-    packaging {
-        USRPATH = /usr
-    } else {
-        USRPATH = /usr/local
+    isEmpty(PREFIX) {
+      packaging {
+        PREFIX = /usr
+      } else {
+        PREFIX = /usr/local
+      }
     }
 
-    target.path = $${BASEDIR}$${USRPATH}/bin/
+    target.path = $${BASEDIR}$${PREFIX}/bin/
 
-    qmfile.path = $${BASEDIR}/usr/share/flameshot/translations/
+    qmfile.path = $${BASEDIR}$${PREFIX}/share/flameshot/translations/
     qmfile.files = $${TRANSLATIONS_FILES}
 
-    dbus.path = $${BASEDIR}/usr/share/dbus-1/interfaces/
+    dbus.path = $${BASEDIR}$${PREFIX}/share/dbus-1/interfaces/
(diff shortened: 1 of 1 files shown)


flameshot.pro
```diff
@@ -198,24 +198,26 @@ RESOURCES += \
...
-    qmfile.path = $${BASEDIR}/usr/share/flameshot/translations/
+    qmfile.path = $${BASEDIR}$${USRPATH}/share/flameshot/translations/
     qmfile.files = $${TRANSLATIONS_FILES}
 
-    dbus.path = $${BASEDIR}/usr/share/dbus-1/interfaces/
+    dbus.path = $${BASEDIR}$${USRPATH}/share/dbus-1/interfaces/
     dbus.files = dbus/org.dharkael.Flameshot.xml
     
-    icon.path = $${BASEDIR}$${USRPATH}/share/icons/
+    icon.path = $${BASEDIR}$${USRPATH}/share/icons/hicolor/scalable/apps
```
Alexander-Wilms: Not sure if scalable is the correct directory. But when I put the icon in .../hicolor/100x100/apps/, it didn't show up under KDE.

flameshot.pro
```diff
@@ -198,36 +198,38 @@ RESOURCES += \
...
-    qmfile.path = $${BASEDIR}/usr/share/flameshot/translations/
+    qmfile.path = $${BASEDIR}$${PREFIX}/share/flameshot/translations/
     qmfile.files = $${TRANSLATIONS_FILES}
 
-    dbus.path = $${BASEDIR}/usr/share/dbus-1/interfaces/
+    dbus.path = $${BASEDIR}$${PREFIX}/share/dbus-1/interfaces/
     dbus.files = dbus/org.dharkael.Flameshot.xml
     
-    icon.path = $${BASEDIR}$${USRPATH}/share/icons/
+    icon.path = $${BASEDIR}$${PREFIX}/share/icons/hicolor/scalable/apps
```
Alexander-Wilms: Not sure if scalable is the correct directory. But when I put the icon in .../hicolor/100x100/apps/, it didn't show up under KDE.

lupoDharkael: could you leave the icon path as it was before?
I'm considering the option of having multiple sized png icons and that will require to revisit where to save them.

Alexander-Wilms: Yes, I've changed it back

PR #68, opened 2018-01-08
Added AppData manifest file for Gnome Software.

xvitaly:
Added AppData manifest file for Gnome Software.


lupoDharkael:
I didn't know about AppData files, thanks!

PR #69, opened 2018-01-08
Fixed errors in desktop files

xvitaly:
Fixed warnings and errors in desktop files:
 * removed shebang;
 * removed encoding tag as it no longer required;
 * added missing GenericName;
 * changed chmod from 0755 to 0644.


lupoDharkael:
Good to know about it 👍

PR #71, opened 2018-01-09
appdata: Install appdata into /usr/share/metainfo

hosiet:
While examining the source code, I found that the appdata description
file was installed into the legacy path /usr/share/appdata/ .

This commit would install appdata info into the canonical dir
/usr/share/metainfo/ .

Refs:
 https://lintian.debian.org/tags/appstream-metadata-in-legacy-location.html


lupoDharkael:
It seems that freedesktop also confirms this https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

Thanks.

PR #89, opened 2018-01-16
Add Traditional Chinese translation

PeterDaveHello:
This is a quick Traditional Chinese based on the Simplified Chinese, hope that there are no big problems.


lupoDharkael:
Thanks for your contribution

PR #101, opened 2018-01-21
Fix case statement (so we build with clang).

scode:
As encountered in https://github.com/NixOS/nixpkgs/pull/34096 we fail
to build under clang:

```
src/utils/screengrabber.cpp:56:16: error: case value is not a constant expression
```


lupoDharkael:
Thanks for the fix

PR #111, opened 2018-02-04
 fix bug

ZetaoYang:
"setWindowFlags": Can not find identifier


lupoDharkael:
I noticed it right after I did the commit but I decided to fix it today as The windows build is still incomplete, thanks anyway.

PR #114, opened 2018-02-16
Add function: pin captures on the desktop

ZetaoYang:
Paste images as topmost floating windows/pin images on the desktop.

#### Features:
1. Pin capture on the desktop
2. Drag & drop
3. Zoom & move

gif

the Pin icon from [icons8](https://icons8.com/icon/set/pin/all). 
[Icon License](https://icons8.com/license/)


lupoDharkael:
I love the concept but I don't like how coupled is the logic of the capture widget having to add a lot of code and a global static state in this specific case (it's not your fault).
Requiring a bigger and bigger "god" class is a code smell which implies the need of a code refactor.
I'll have to think about the design for the next days and make implementations of tools like this more clean (and make the implementation of text tool easier tool too).

ZetaoYang:
@lupoDharkael When I finished with the new API, I will close this PR and open a new one.

PR #118, opened 2018-02-18
Travis: fcitx support for AppImage

ZetaoYang:
travis add fcitx support for AppImage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,9 @@ services:
 
 language: cpp
 
-os:
-  - linux
+branches:
+  only:
+    - master
 
 cache:
     directories:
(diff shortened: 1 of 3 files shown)


lupoDharkael:
MacOS? I don't have a mac so I can't test it there. Does it work?

ZetaoYang:
The current flaemshot code is mainly for Linux, so I suggest you focus on the Linux platform. flameshot is a great and unique screenshot application on the Linux platform. While MacOS platform, there are many similar applications. The purpose of this Pull Request is to provide the possibility of compiling MacOS in the future.

MacOS compilation is  unsuccessful in continuous integration. So you can comment out the travis part about MacOS. It is difficult to support MacOS at present, and may support it in the future.

lupoDharkael:
I see. Thanks for the changes.

PR #147, opened 2018-03-03
Small fix in Spanish translation and cleanup

AlfredoRamos:
Just a small punctuation fix, a couple of string marked as done and translations cleanup done with `lupdate`.

### DIFF
--- a/translations/Internationalization_ca.ts
+++ b/translations/Internationalization_ca.ts
@@ -97,14 +97,6 @@ Utilitzeu la roda del ratolí per a canviar el gruix de l&apos;eina.</translatio
         <translation>Configuració</translation>
     </message>
     <message>
-        <source>Button Selection</source>
-        <translation type="vanished">Selecció de botó</translation>
-    </message>
-    <message>
-        <source>Select All</source>
-        <translation type="vanished">Selecciona-ho tot</translation>
-    </message>
-    <message>
         <source>Interface</source>
         <translation>Interfície</translation>
     </message>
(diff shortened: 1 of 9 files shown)


ZetaoYang:
@AlfredoRamos Please submit the translated files to this [repo](https://github.com/flameshotapp/translations).

AlfredoRamos:
Mmm... I thought it was a personal project...

AlfredoRamos:
@lupoDharkael Adding translations into a separate repository will make it harder for contributors. You can't simply run `lupdate` because the translations won't be synchronized with the sources (this repo) and the translations repository.

You need `lupdate` to get the translations (and cleanup), however you also need the *.pro file to properly get new strings.

The translations strategy needs more planing, like creating tools to sync translations, use submodules or services like Crowdin.

lupoDharkael:
@AlfredoRamos that makes sense. I think we should keep the translation dynamics in the main repo as a different repo implies more work to be done.

ZetaoYang:
@AlfredoRamos If the project is built using cmake, is `lupdate` useful? My intention is this, to distinguish between code and translation contributions.

AlfredoRamos:
@ZetaoYang The project uses `qmake` to build and install the files, I don't see any `CMakeLists.txt` file.

Even if this was a `cmake` project, you still need `lupdate` to search new strings, that's a core functionality in Qt, see [Using lupdate](https://doc.qt.io/qt-5/linguist-manager.html#using-lupdate).

What would be the benefit to split them? I can see it would help when there are a lot of files per language, however right now it only creates a single TS file per language.

@lupoDharkael I will open a new pull request, pull requests can't be reopened if the fork is deleted, which I did.

ZetaoYang:
@AlfredoRamos I am sorry for the inconvenience caused to you, it has been corrected to [translation-instruction](https://github.com/flameshotapp/translation-instruction). And it contains contributor list.
Thank you for your contribution.

Issue #140, opened 2018-02-28
Providing 3 menu entries together might look weird (wishlist issue)


hosiet:
As reported in downstream (Debian https://bugs.debian.org/887466):

```
The display name in the system menu looks odd.  No other program has "Launch
XXX" as its entry, just "XXX".

There's also that "Configure Flameshot" which is redundant, as this is
something the user can do from the program -- no other program has such
separate configuration (at least not in the menu).
```

Personally I have the same feeling. Three menu entries altogether occupies lots of spaces, especially in GNOME Shell, where icons are big.

Example screenshot attached below (system is under `zh_CN` locale; not a big problem so never mind)

<img width="584" alt="default" src="https://user-images.githubusercontent.com/8443926/36798299-30e0286c-1ce5-11e8-8fbe-7f5361d50aaf.png">

My personal wish is to keep only one `.desktop` file distributed during installation. @lupoDharkael  I'm wondering the opinion you hold towards this issue.

lupoDharkael:
I agree. It doesn't look right.
Launch Flameshot could be deleted as the process should be launched with every command if it isn't active.
In the future I'd like to add a menu launcher similar to the main menu of Spectacle (as a command and a menu item in the trayicon) and a new mode to edit images using Flameshot as a classical windowed program.

The question is: what desktop files should be installed?

They are usefull to ease the use of non cli proficient users. As an example, if the user disables the icon accidentally, they need a desktop file o the command line to re-enable it.

AlfredoRamos:
I think [desktop actions](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#extra-actions) would help with this issue, I'll see what I can do about it, to create a pull request.

PR #149, opened 2018-03-04
Merge desktop files

AlfredoRamos:
These changes try to address #140, so my proposal is to keep only one `*.desktop` file, and add the same functionality of the other desktop entries as Actions, see [Additional applications actions](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#extra-actions).

Even if in a near future a new menu launcher is created ([reference](https://github.com/lupoDharkael/flameshot/issues/140#issuecomment-369302364)), the functionality will remain as this is a official especification.

This however, has a downside, and is that not every desktop environment support extra application actions. Right now the only one I'm aware that support this is GNOME Shell.

The menu is shown if you right-click the icon. Here's a preview:

screenshot_20180303_204827


lupoDharkael:
wow, this is amazing and a clean solution for the multiple desktop files issue!
Thank you!

PR #154, opened 2018-03-05
 travis_linux make adjustments

ZetaoYang:
Make adjustments, because the desktop file changes.


hosiet:
Hmmm. Only an advice, could you please squash those changes and rebase them against current git HEAD?

ZetaoYang:
@hosiet Thank you for your advice, I paid attention to this problem: the extra commit information.

PR #162, opened 2018-03-09
README - point out flameshot must be running.

scode:
Motivation: At least twice I have forgotten about this and become
confused by the fact that commands like `flameshot gui` will just exit
without doing anything.

Ideally an error should be printed to indicate that flameshot isn't
running, but this README tweak should make it more likely to be
realized by someone double checking usage instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@
 If you want to contribute check the [CONTRIBUTING.md](./CONTRIBUTING.md)
 
 ## Usage
-Example commands:
+Example commands (all assuming flameshot is already running in the background):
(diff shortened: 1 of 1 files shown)


hosiet:
That is not correct.

With proper installation, `flameshot gui` will automatically bring flameshot up if it is not running. If not, please consider it a bug and file a bug report.

scode:
Ok. If I have time I'll look into understanding the mechanism used;  I've been using a manual build and later a nix package and was not aware that was expected behavior.

Closing PR.

Issue #121, opened 2018-02-20
Blurring doesn't hide text


kramer65:
I love Flameshot and use it on a daily basis. But one thing is a real problem for me. Blurred sensitive text can still be read. Since I work in the medical industry, blurring patient information is extremely important. Have a look at the screenshot below though. If you look closely, you can still read the first part of the text. 

When I used to work on a mac I used Monosnap, which does an awesome job at blurring. Is it somehow possible to make the blurring resemble that of Monosnap?

Is this blurring a setting somewhere or can this be marked as a bug?

![no secret](https://user-images.githubusercontent.com/596581/36417391-143e72b4-162c-11e8-88f2-b88fa34d335c.png)

vendforce:
go over it twice. it will blur it out

lupoDharkael:
I've made it more blurry with the last commit. The actual implementation is not ideal as it requires 3 iterations of the Qt's built in blur. The perfect solution would be to implement a specific blur and be able to control the level of blur with the mouse wheel. I think Deepin's screenshot tool has an implementation, I should check it.
I hope the blur is enough with the last change while I investigate the implementation of a specific blur algorithm.

kramer65:
Hi @lupoDharkael - Thanks for the extra blur. Unfortunately I don't think it's enough for me. As you already suggest I think you need a different kind of blurring algorithm. I would really suggest to have a look at the blurring of Monosnap. Maybe someone on the interwebs knows what kind of algorithm that is.

mbirth:
Could you make the blurring according to the selected pencil width? Because blurring out big headings often needs multiple passes. It would be awesome if we could just crank up the mouse wheel and only blur it once.

reitzig:
Seconded. With 0.5.1 (trunk in Ubuntu 18.04 repos):

![image](https://user-images.githubusercontent.com/4246780/46693869-62420300-cc0b-11e8-8ca9-27ba3d345570.png)

### Workaround

Get more area under the blur and/or repeat in different orientations.

Single pass, higher box: 
![image](https://user-images.githubusercontent.com/4246780/46693913-8a316680-cc0b-11e8-883c-cd7db0cc8877.png)

Four passes, once along each diagonal:
![image](https://user-images.githubusercontent.com/4246780/46693942-9caba000-cc0b-11e8-848e-dc361d8cbc33.png)

reitzig:
[...]

PR #201, opened 2018-04-26
Add option to configure blur radius

the-saif-ahmad:
Fixes issue #121 

Adds configuration slider to determine modify the blur radius (I still haven't determined a decent range for it, I assume it will be fixed later)


lupoDharkael:
I've been testing with the blur and the only good solution is to write a new algorithm, increasing the radius ends up being ineffective and a low radius is not enough for hiding huge text.

the-saif-ahmad:
Which algorithm is it using currently?

PR #210, opened 2018-05-06
Add delete Imgur image button

### LINKED ISSUE #180: Add Imgur Warning
Please add a warning when uploading an image directly to Imgur, at-least until personal Imgur account support is added. It's very easy to accidentally upload a sensitive image and not be able to subsequently remove it.

On that note, a log file with the links of past uploads would be pretty nice...

AlfredoRamos:
This would allow the user to delete the uploaded image.

The auto generated `deletehash` is unique for each image and and can't be guessed, just like the Image `id`.

Fixes #180


lupoDharkael:
your contribution is appreciated.

PR #212, opened 2018-05-07
debian: Update deb packaging instructions

hosiet:
With this commit, I would expect users able to build deb packages with a cleaner packaging instruction that is compatible from down to Ubuntu 16.04.x LTS.

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -1,16 +1,365 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: flameshot
-Source: <url://example.com>
+Source: https://github.com/lupoDharkael/flameshot/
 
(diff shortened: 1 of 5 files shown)


debian/copyright
```diff
@@ -1,16 +1,365 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: flameshot
-Source: <url://example.com>
+Source: https://github.com/lupoDharkael/flameshot/
 
 Files: *
-Copyright: 2016 lupoDharkael
-License: GPL-3.0+
+Copyright: 2016-2017 lupoDharkael <izhe@hotmail.es>
```
lupoDharkael: The correct copyright would be 1017-2018

Issue #223, opened 2018-05-15
flameshot is not runnig in ubuntu 18.04


Lvaskz:
**Describe the bug**
The flameshot is not running in Ubuntu 18.04
Icon not appear in dock, no fuctions, nothing at all

**To Reproduce**
Download flameshot.appimage.
Right click, go permission, check make exexutable.
doble click flameshot.appimage

**Expected behavior**
Icon must to appear in dock ubuntu 18.04, and the app must running normal.

**Desktop **
 - OS: [ Ubuntu 18.04 Gnome]
 - Version [3.28]

ningacoding:
Installing as documented [here](https://github.com/lupoDharkael/flameshot#installation)
with `apt install flameshot`, works extremely good.
![image](https://user-images.githubusercontent.com/17342546/40147331-e50e636e-592e-11e8-910c-28b1861d5bd4.png)

Lvaskz:
Thanks for suggest. I will proof but actually I prefer a more agnostic solution. The use of debs packages generates some distrust to me. I think you could provide a flatpak package in the future, the updates arrive in real time, they are loaded in flathub or anywhere else you want, you could provide a flatpakref. And it could be better integrated into the desktop.

Lvaskz:
Hi, it works really great. Are you thinking provide a snap? It would be nice, I see some advance in the main page. Will you?

hosiet:
The problem is not related to Ubuntu 18.04. In fact, I could confirm that current AppImage binary downloaded from the Release page cannot run under my Debian Unstable. Terminal output is:

```
% ./flameshot_x86_64_0.5.1.AppImage 
QXcbConnection: XCB error: 9 (BadDrawable), sequence: 274, resource id: 73400325, major code: 55 (CreateGC), minor code: 0
QXcbConnection: XCB error: 9 (BadDrawable), sequence: 275, resource id: 73400325, major code: 130 (Unknown), minor code: 
```

Not sure if it is to be fixed or not but the issue do exist.

omkarnathsingh:
Please release a update for Ubuntu package, it was last updated in February.

ZetaoYang:
@omkarnathsingh You can try Travis CI daily build version.

omkarnathsingh:
@ZetaoYang Where to download those .deb file?

ZetaoYang:
@omkarnathsingh  https://travis-ci.org/lupoDharkael/flameshot
[...]

PR #236, opened 2018-06-01
Travis CI tests improvements

### LINKED ISSUE #213: Travis error notification
**Describe the bug**
The RPM build fails in travis and it isn't notified as a failed build.
```
PM build errors:
    Macro expanded in comment on line 59: %{name}.svg
    Macro expanded in comment on line 60: %{name}.svg
    File not found: /build/BUILDROOT/flameshot-0.5.1-1.fc27.x86_64/usr/share/applications/flameshot-init.desktop
    File not found: /build/BUILDROOT/flameshot-0.5.1-1.fc27.x86_64/usr/share/applications/flameshot-config.desktop
```
**Expected behavior**
A notification from Travis about a failed build.

AlfredoRamos:
This change aims to improve some logic in bash scripts, as well to make easier to maintain the installation of dependencies, build and deploy process.

Some highlights:

- Add `set -e` to bash scripts to exit immediately if a command exits with a non-zero status (Fixes #213)
- Move `.travis_linux.sh` to `travis/linux_script.sh`
- Move `.travis_macos.sh to travis/osx_script.sh`
- Use variable syntax `${var}` consistently
- Quote directory paths
- Split long commands into multiple lines
- Remove unneeded characters (`;`, `./`) and whitespaces
- Move `before_install` step instructions into a separate file
- Move `install` step instructions into a separate file

To maintain consistency, scripts are named using the following format:

```
<OS>_<step>.sh
```


lupoDharkael:
I don't really understand why this one fails https://travis-ci.org/lupoDharkael/flameshot/jobs/386451095

OSX's build fails because I don't really support that platform but I could fix the compilation errors after the merge.

lupoDharkael:
I don't think we should have osx in the CI, I refuse to maintain a version for Apple products. could you disable the build for that platform? maybe not delete the script but just disable the build job.

ZetaoYang:
@AlfredoRamos @lupoDharkael  just comment them out.
```
 #      - os: osx
 #        compiler: clang
 #        osx_image: xcode9.2
```

AlfredoRamos:
The only step that keeps failing is when running `source /opt/qt53/bin/qt53-env.sh` At this point I'm not sure if there's something wrong with that build or with the scripts.

I will investigate about that script.

ZetaoYang:
@AlfredoRamos It succeed.

AlfredoRamos:
@ZetaoYang Finally! :joy: 

Apparently there's a problem using `source`, I had to download `qt53base` deb to export the same environment variables.

https://github.com/travis-ci/travis-ci/issues/7472

@lupoDharkael @ZetaoYang Sorry if you got spammed with emails about failing tests, it won't happen again.

P.S.: I don't think is related, but it would be good if somebody could check if #223 was somehow related with the failing test, I don't use Ubuntu.

PR #243, opened 2018-06-04
Store settings colors in hexadecimal format

### LINKED ISSUE #142: Store config colors in RGB format
It would be useful to store configuration color in standard RGB format instead of just plain hex qcolor format. There is a easy way to do so in QT: 
http://doc.qt.io/qt-5/qcolor.html#name and http://doc.qt.io/qt-5/qcolor.html#setNamedColor

I use automated installation of the whole desktop and some colors are gotten from configuration templates - and there is no easy way in automation tools like Ansible or Puppet to convert QColor hex values to RGB values. Also it would be useful in case you want to have Flameshot ui color to be similar with main QT theme color - but there no easy way to convert theme RGB color to Flameshot Hex one.

AlfredoRamos:
This will make easier to see and edit colors in the settings file.

The format is defined by `QColor::HexRgb`

Fixes #142

https://doc.qt.io/qt-5/qcolor.html#NameFormat-enum

### DIFF
--- a/src/utils/confighandler.cpp
+++ b/src/utils/confighandler.cpp
@@ -74,8 +74,11 @@ void ConfigHandler::setButtons(const QVector<CaptureButton::ButtonType> &buttons
 
 QVector<QColor> ConfigHandler::getUserColors() {
     QVector<QColor> colors;
+
     if (m_settings.contains("userColors")) {
-        colors = m_settings.value("userColors").value<QVector<QColor> >();
+        for (const QString &name : m_settings.value("userColors").toStringList()) {
+            colors.append(QColor(name));
+        }
     } else {
         colors = {
             Qt::darkRed,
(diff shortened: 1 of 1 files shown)


lupoDharkael:
I guess this is going to break color configurations for some users. 
I've been waiting to modify the serialization of the tools to use strings instead of enums. The enums require casting after the serialization and it's kind of awful to handle (and not very friendly for a plugin based tool system in the future).
The next release will be kind of a big one, maybe this would be a good opportunity to add some breakage in the configuration.
I don't like to break things but the reason Flameshot isn't 1.0 yet is because this is my first big project and I knew I would need to refactor multiple times and redesign parts from time to time.

AlfredoRamos:
@lupoDharkael I just tested the old config (until commit dfef834) with these changes and I haven't noticed any errors, even the colors were respected (using the old `QVariant` value).

It would be good if somebody could test it.

lupoDharkael:
I'll test it later

PR #253, opened 2018-06-13
Set dynamically Imgur client ID

AlfredoRamos:
This will allow users to set their own Imgur `client_id` passing the `IMGUR_CLIENT_ID` variable to `qmake`.

If the users does not set this variable, it will fallback to the default value.

P.S.: This will create a new project file (`src/imgur.pri`) to avoid pollution of the main project file and to keep data specifically about Imgur.

### DIFF
--- a/flameshot.pro
+++ b/flameshot.pro
@@ -248,13 +248,13 @@ unix:!macx {
 
     dbus.path = $${BASEDIR}$${PREFIX}/share/dbus-1/interfaces/
     dbus.files = dbus/org.dharkael.Flameshot.xml
-    
+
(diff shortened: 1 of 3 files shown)


ZetaoYang:
@AlfredoRamos I don‘t agree with this approach, the best way is to provide a graphical interface to modify the imgur `IMGUR_CLIENT_ID` settings for users.

lupoDharkael:
I agree with this method as withour the variable it works as before this PR and also enables the overwrite of the key. The Imgur related implementation needs a lot of improvements and this is a good first step similar to what other projects do and not incompatible with a graphical configuration.

PR #264, opened 2018-06-30
Support BSDs OSs

daniloegea:
Use defined(Q_OS_LINUX) || defined(Q_OS_UNIX) where applicable to support BSD OSs.

We maintain the flameshot in the FreeBSD ports tree. The last version (0.5.1) added some checks for Linux and broke the app on FreeBSD.


lupoDharkael:
Thanks for the contribution! I'll keep in mind to keep using this and don't break the BSD build. FreeBSD looks like a cool system and I'd like to give it a try someday.

daniloegea:
Thank you! :)

PR #277, opened 2018-07-22
screengrabber: fix devicePixelRatio detection

PeterCxy:
At least on two of my devices, `QApplication::desktop()->devicePixelRatio()` is always 0. This patch fixes detection of the `devicePixelRatio` by fetching the correct value from the screen that the current desktop is on, and also corrects the calculation of the selection area in `screengrabber` when the pixel ratio of the pixmap is set correctly.

The bug causes no problems on non-HiDPI devices or devices with integer scale factors for some reason that I haven't investigated yet (maybe it has a default value of 1 if 0 is fed into `setDevicePixelRatio`?). However, on my laptop with a scale factor of `1.5` (set through KDE System Settings on Xorg), flameshot breaks and the background is scaled improperly.

For reference, this is how my fractionally-scaled desktop should look like:
screenshot_20180722_205120

And here is how it looks through flameshot before this patch:

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -85,7 +85,9 @@ QPixmap ScreenGrabber::grabEntireDesktop(bool &ok) {
                   geometry.width(),
                   geometry.height())
               );
-    p.setDevicePixelRatio(QApplication::desktop()->devicePixelRatio());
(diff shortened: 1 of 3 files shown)


lupoDharkael:
Thank you!
I don't own a hidpi monitor so trying to solve this without testing it made it a problem. 
The code looks fine!

PR #309, opened 2018-08-12
flameshot.pro: Install icons into $prefix/share/icons/hicolor/ instead of $prefix/share/pixmaps

hosiet:
While #293 somehow solves the icon problem, it is not the best solution. According to Freedesktop.org Icon Theme Specification
(https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html):

    Directory Layout

    [...]
    By default, apps should look in $HOME/.icons (for backwards compatibility),
    in $XDG_DATA_DIRS/icons and in /usr/share/pixmaps (in that order).
    [...]
    In order to have a place for third party applications to install their icons
    there should always exist a theme called "hicolor".

While using /pixmaps/ directory ensures the least priority in search icons, the directory name itself "pixmaps" does not suite our actual situation (provided that "flameshot.svg" is a scalable icon, not a **pixmap**).

I suggest that we place this icon into the "hicolor" theme (aka fallback theme) in /icons/ directory under "scalable/apps/" subdirectory.

Proposed changes:

  1. Install flameshot.svg under `$prefix/share/icons/hicolor/scalable/apps/` instad of `$prefix/share/pixmaps/`.

  2. Rename qmake "icon" target into "appsvgicon" target so that the name "icon" may still be used in the future for other tasks.

  3. The embedded rpm spec file is updated accordingly.

* * *

cc @AlfredoRamos


AlfredoRamos:
I agree that the `pixmap` is not the best solution, just was the easiest.

From the specification:
> [...] Minimally you should install a 48x48 icon in the hicolor theme. This means installing a PNG file in $prefix/share/icons/hicolor/48x48/apps. Optionally you can install icons in different sizes. For example, installing a svg icon in $prefix/share/icons/hicolor/scalable/apps means most desktops will have one icon that works for all sizes.

> [...] PNG is the recommended bitmap format, and SVG is for vectorized icons. [...] Support for SVGs is optional.

Maybe now that is going to be changed, a 48x48 PNG image should be added at least.

flameshot.pro
```diff
@@ -270,7 +270,7 @@ unix:!macx {
     }
 
     INSTALLS += target \
-        icon \
+        appsvgicon \
```
AlfredoRamos: Why not simply `iconsvg`? I think is more descriptive that way.

Maybe `icon` can be used for a 48x48 PNG image.

hosiet:
Thanks, I'd agree with that proposal. However:

```
% file ./img/app/flameshot.png
./img/app/flameshot.png: PNG image data, 100 x 100, 8-bit/color RGBA, non-interlaced
```

The current embedded png icon is not of a standard size, which made it a little bit awkward.

Anyway, I believe the PNG icon problem should be tracked in a separate issue and that does not affect current PR.

flameshot.pro
hosiet: If there's another icon in the future but it is not an app icon, say it has to be put into `/$theme_name/scalable/actions/` dir, the `iconsvg` target is not applicable.

I'm not really familiar with `qmake`; is there an option to install the whole directory? In that way we may generate directory tree in advance and install all files in one target.

AlfredoRamos: That would be better (install whole directory)

I think using a wildcard would allow you to install a whole directory, but I'm not sure, I'll try.

AlfredoRamos:
Yes, you can use a wildcard to install everything (directories included).

I tried the following with success:

**flameshot.pro**
```qmake
    hicolor.path = $${PREFIX}/share/icons/hicolor/
    hicolor.files = img/hicolor/*
```

The tree for the `hicolor` directory looks like this:

```shell
img/hicolor/
├── 48x48
│   └── apps
│       └── flameshot.png
└── scalable
    └── apps
        └── flameshot.svg
```

It installs the files properly.

hosiet:
Looks great. I'll convert this PR and use the prebuilt directory structure.

@lupoDharkael what do you think?

lupoDharkael:
Makes more sense with this change.

PR #332, opened 2018-08-31
use m_thickness in blur_tool for blur radius

mefest:
allows you to change the radius of the blur depending on the thickness of the brush

### DIFF
--- a/src/tools/blur/blurtool.cpp
+++ b/src/tools/blur/blurtool.cpp
@@ -58,7 +58,7 @@ void BlurTool::process(QPainter &painter, const QPixmap &pixmap, bool recordUndo
     QRect selectionScaled = QRect(p0 * pixelRatio, p1 * pixelRatio).normalized();
 
     QGraphicsBlurEffect *blur = new QGraphicsBlurEffect;
-    blur->setBlurRadius(10);
+    blur->setBlurRadius(m_thickness);
     QGraphicsPixmapItem *item = new QGraphicsPixmapItem (
                 pixmap.copy(selectionScaled));
     item->setGraphicsEffect(blur);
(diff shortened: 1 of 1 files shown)


AlfredoRamos:
Shouldn't a minimum/maximum and default value should be specified? It doesn't have much sense to allow a blur radius of `0`

mefest:
of course.
what if value transform to range [10..50]?

```
...
// (m_thickness - in_min) * (out_max - out_min) / (in_max - in_min) + out_min
// (m_thickness - 0) * (50 - 10) / (100 - 0) + 10
    int blurRadius = m_thickness * 0.4 + 10;
...
```

AlfredoRamos:
I was thinking on something simpler:

```cpp
int minThickness = 5;
int maxThickness = minThickness * 10;
m_thickness = m_thickness < minThickness ? minThickness : m_thickness;
m_thickness = m_thickness > maxThickness ? maxThickness : m_thickness;
```

I remember I saw some issues with blur radius greater than 30, but I'm not sure if it was just me.

I think this is a great addition to the tool.

lupoDharkael:
I agree there should be a way to modify the intensity of the blur but when I tried modifying the radius it wasn't scaling well. The purpose of the blur is hiding (most of the time) is hiding text, including a potentially a big font.
I'm using the default algorithm of Qt, which is good for a soft blur but not for things such as hiding big text.
The proper solution is to find a better blur algorithm implemented for Qt which would be ideal, the last resort is implementing one from scratch.

mefest:
I added a check for the maximum and minimum radius values. Just slightly changed the algorithm for a better blur near the edges.

blur radius 50


blur radius 50

PapiCZ:
I think it would be better to have something like `m_thickness * 10` or `m_thickness * 5`. Because arrow with `m_thickness = 50` is huge arrow, but blur 50 is only slightly changed blur.

mefest:
With a large radius there will be problems at the edges of the screen.

PR #343, opened 2018-09-10
Update README.md

SeedPuller:
just added a installation command. just like other distro's. :)


lupoDharkael:
thanks for the contribution!

PR #346, opened 2018-09-14
Save on copy.

vlakam:
Configurable feature to save image automatically when copying into clipboard.

### DIFF
--- a/src/config/geneneralconf.cpp
+++ b/src/config/geneneralconf.cpp
@@ -17,7 +17,6 @@
 
 #include "geneneralconf.h"
 #include "src/utils/confighandler.h"
-#include "src/utils/confighandler.h"
 #include "src/core/controller.h"
 #include <QVBoxLayout>
 #include <QHBoxLayout>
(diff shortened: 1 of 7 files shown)


dgw:
If I had developed this feature it would have been "Copy to clipboard on save", but it's the same effect: file on disk + image in clipboard.

:+1: from me (from a functional perspective; haven't reviewed the code or anything)

vlakam:
Obsolete as https://github.com/lupoDharkael/flameshot/pull/516 implements exactly the same functionality

ElijahLynn:
Virgin user here (20 minutes ago, first install on Arch). This is the deal breaking feature for me and I was just going to open an issue to see if I was somehow not understanding. Thanks for linking to #516 saying it is duplicate of this as I saw it first but didn't realize it was for this as it mostly talks about saving the path to the clipboard but not the actual image. I just asked for a clarification here https://github.com/lupoDharkael/flameshot/pull/516#issuecomment-505523566.

Issue #350, opened 2018-09-20
Pin tool not pinning to original position on Ubuntu 16.04


elonliu:
**Flameshot version**

v0.6.0

**Describe the bug**

Copy an area of the screen and pin it using the pin tool. The pinned picture doesn't appear at the same position as it was, but moved a little.

Not sure this is a bug, but if it stays exactly where it was, it could be more helpful in some cases.

![image](https://user-images.githubusercontent.com/5549178/45808233-e9c8e000-bcf7-11e8-8eef-961ccc11df9d.png)

**To Reproduce**

On Ubuntu 16.04 screen, begin screenshot to copy a region and click the pin tool button.

**Expected behavior**

The pinned picture stays exactly where it was copied

**System Information**

Ubuntu 16.04 LTS

ZetaoYang:
@elonliu Yes, I know this situation, it's  not a bug.

ZetaoYang:
@elonliu Please see #352.

elonliu:
Cool!

PR #352, opened 2018-09-20
Some adjustment for Pin Tool

ZetaoYang:
- Change `adjusted_pos`, pin pictures to their original position. issue #350 
- Change the mouse wheel scroll delta, because `5` is too small, I have to scroll much times before I can adjust the pinned picture to the size that I want.
- Add `Qt::X11BypassWindowManagerHint` again.

All of above was tested normally, on **Fedora 28, Gnome 3.28.x, Qt 5.11.1**.

### DIFF
--- a/src/tools/pin/pintool.cpp
+++ b/src/tools/pin/pintool.cpp
@@ -45,11 +45,7 @@ QString PinTool::description() const {
 QWidget* PinTool::widget() {
     PinWidget *w = new PinWidget(m_pixmap);
     const int &&m = w->margin();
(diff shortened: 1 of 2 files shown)


lupoDharkael:
Why do you add Qt::X11BypassWindowManagerHint? is there something problematic about the window manager handling the widgets?
In my system, with that flag enabled, if you click a pinned image it doesn't go to the front, so the newer created pinned images are always over the older ones. I don't think that's the desired behavior.

I agree with the other changes.

ZetaoYang:
@lupoDharkael I have remove that flag.

PR #361, opened 2018-10-04
Add japanese translation

sicklylife-jp:
I translated into Japanese.


lupoDharkael:
Thanks for the contribution!

PR #363, opened 2018-10-04
Fixes #362

### LINKED ISSUE #362: Enter key doesn't copy content to clipboard
**Flameshot version**
0.6.0

**Describe the bug**
The return key is commonly called as _enter key_ in many times during our day to day usage, as opposed to _return key_, at least from my experience. So, when we use a big keyboard with a dedicated numeric keyboard on the side, we usually expect the _return key_ to have the same behavior as the _enter key_.

On flameshot, _return key_ saves the selection on GUI mode to the clipboard, but the _enter key_ does not replicate this behavior. I stumbled upon this a couple of times when I reached the numeric enter key, as it is closer to the mouse.

**To Reproduce**

- Start GUI mode
- Make a selection
- Press "Enter" on  numeric keyboard
- Nothing Happens 

**Expected behavior**
- Start GUI mode
- Make a selection
- Press "Enter" on  numeric keyboard
- Content should be copied to clipboard 

**System Information**
Ubuntu 16.04

This is nothing critical, so please feel free to reply in case you don't agree

btw much love from brazil <3 flameshot is great

Streppel:
This fixes #362, making the enter key copy the content of the GUI selection copy the content to the clipboard

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -741,6 +741,7 @@ void CaptureWidget::initShortcuts() {
     new QShortcut(Qt::Key_Space, this, SLOT(togglePanel()));
     new QShortcut(Qt::Key_Escape, this, SLOT(deleteToolwidgetOrClose()));
     new QShortcut(Qt::Key_Return, this, SLOT(copyScreenshot()));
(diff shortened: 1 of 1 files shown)


lupoDharkael:
thanks!

PR #364, opened 2018-10-04
Adds Brazilian Portuguese translations 

Streppel:
Adds Brazilian Portuguese translation file and reference on flameshot.pro file


lupoDharkael:
thanks for your contribution!

PR #374, opened 2018-10-09
[WIP] Add line thickness to side panel

### LINKED ISSUE #373: If I do not have a mouse, I can not adjust the line width and color
Require add button to increase line width and select color.
If I do not have a mouse, I can not adjust the line width and color.

AlfredoRamos:
This adds a slider in the side panel to change the line thickness to allow devices to change this value without the use of the scroll of a mouse.

This is working in the current state, however I'm marking it as work in progress because I would like to receive feedback (@lupoDharkael) on the filename of the widget `colorpickerwidget`, now it does not just allows to change the color.

I was thinking in renaming those files to `sidepanelwidget` or `tooloptionswidget`. Should I leave it as is?

Fixes #373

P.S.: This also replaces the deprecated `Q_ENUMS` macro with `Q_ENUM` (since [Qt 5.5](https://doc.qt.io/qt-5/whatsnew55.html)), and some other minor fixes.

### DIFF
--- a/flameshot.pro
+++ b/flameshot.pro
@@ -140,7 +140,7 @@ SOURCES += src/main.cpp \
     src/tools/text/textwidget.cpp \
     src/core/capturerequest.cpp \
     src/tools/text/textconfig.cpp \
-    src/widgets/panel/colorpickerwidget.cpp
+    src/widgets/panel/sidepanelwidget.cpp
 
 HEADERS  += src/widgets/capture/buttonhandler.h \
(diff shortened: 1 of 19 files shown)


lupoDharkael:
I like the name sidepanelwidget, once you change the class name and the file name I'll merge this

AlfredoRamos:
Just for the record, tests didn't pass because transfer.sh seems to be offline right now.

PR #375, opened 2018-10-09
Tweak tools' tooltip phrasing to be consistent

dgw:
Some tooltips were written as imperatives ("Set foo to bar"), others were written as simple present ("Performs foo action"). All should now be consistently written in imperative form (unless I missed any…).

Running `lupdate flameshot.pro` yielded "unfinished" translations for the edited strings, and marked the existing translations as "vanished", which would mean having to re-translate every string I touched in all supported languages before the next release… I can simply commit and push the updated files in this PR if desired. I can also just update the source strings manually, leaving the translations as-is and accepting that the verb tenses might not agree between English and the localizations, or take some other action after receiving direction in this thread.

Do let me know how you'd like me to handle the translations, and I'll be happy to update this PR. :smile_cat: This is my first time working with translations in a Qt project, so I'm hoping to learn something.


lupoDharkael:
I think it's better to have separated PRs one with the source modifications and other with translations updated. That way changes are more atomic as they should be.

This looks good so far.
Do you need help about the workflow of the translation process?

dgw:
If you're happy with this PR as-is, please remove the `[WIP]` from the title. It should probably be merged soon, as any translation updates will depend on having the current strings. Opening another PR based on this one, to get the updated strings, will just duplicate this commit in that PR. (I've tried to do dependent PRs before… it's messy.)

The translation workflow itself seems easy enough. I already figured out how to update the TS files with `lupdate`, and the Linguist documentation came up when I was looking for how to do that. As far as the actual editing of translation files goes, I should be set—especially if I'm correct that I needn't mess with `lrelease` or QM files at all, just the TS files stored in the repo. Linguist was even already installed on my system. Couldn't be easier to get started.

What I do need help with is how much I should wing it when updating these. Since I don't actually know/speak most of flameshot's supported languages, how far down this list would you like me to go?

  1. Run `lupdate`
  2. Take any existing "vanished" translations for the old strings and use them for the new strings
      * Marked as "unfinished"?
      * "vanished" translations removed?
  3. Properly retranslate the new strings in languages I know or can get help with from friends
  4. Retranslate remaining strings using machine translation tools

Everything I've looked up since starting this indicates that "vanished" translations aren't included when running `lrelease`, meaning the application would drop back to English for any string I changed that had a translation before. Probably not ideal… For what it's worth, I'd prefer to keep the old strings in my own hypothetical project until new translations are made—but it's your call here.

I'm less clear on what happens to "unfinished" translations. [Qt Linguist `lrelease` docs say](https://doc.qt.io/qt-5/linguist-manager.html#using-lrelease) that only finished translations are included, but `lrelease` has an option `-nounfinished` to exclude unfinished strings—implying that unfinished strings are otherwise included, in contradiction to the docs. flameshot's project file only says `command = $$LRELEASE -removeidentical $$tsfile -qm $$qmfile` in the translation-building loop, so I'm not sure if unfinished translations are included. Your input would be helpful here, too.

Think I've gone on long enough about stuff I only sort-of understand for one comment. :joy_cat: I'll poke around Linguist a bit more to make sure I know what I'm doing, and await direction from you.

AlfredoRamos:
I would change the `<source>` text with `grep` or an editor like Notepadqq, then run `lupdate` with absolute locations to update the line numbers, add the new translations and avoid marking those strings as vanished.

From the docs and description, I think you're right, when a translation is marked as unfinished is included but not used, maybe adding that flag and `-compress` would help to generate a smaller file size.

dgw:
First I missed that there was a new translation (Serbian), then I forgot to update the commit message… but #380 is now open to update the corresponding strings in the translation files.

@lupoDharkael I hope this was OK. Not getting any input from you, I decided to go with what @AlfredoRamos suggested. (More or less—I didn't commit the results of running `lupdate`, because it touched a bunch of unrelated stuff. That should probably be done after this PR is merged.)

dgw:
Edited this PR's title to indicate it's no longer WIP.

PR #397, opened 2018-11-06
Remove transfer.sh

### LINKED ISSUE #396: transfer.sh will shutdown on 30th of November.
"transfer.sh will shutdown 30th of November 2018
After running transfer.sh for 4 years, it has to come to an end. I started this project, because I needed something quick and simple to share files between from the command line. Now it has grown out of control, with millions of files being shared each month, and no longer sustainable and affordable to keep it running for free. From day one the code has been opensource, of which you can run your own instance. Special thanks to Andrea Spacca for maintaining the opensource version, the amazing community for all efforts, the patrons for all support and to Gopher  for being you. 
So Long, and Thanks for All the Fish. 
Remco Verhoef."
Can be seen on https://transfer.sh/

AlfredoRamos:
Transfer.sh is now discontinued, however it has only been commented out in case a similar service is used in a future.

Fixes #396

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,8 +72,8 @@ script:
 # - if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then source travis/osx_script.sh; fi
   - pwd && ls
 
-after_success:
-  - if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then source travis/linux_after_success.sh; fi
+#after_success:
+#  - if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then source travis/linux_after_success.sh; fi
 
 # deploy:
 #    # Deploy packages to Github Release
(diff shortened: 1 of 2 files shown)


ZetaoYang:
@AlfredoRamos The opensuse build service has been finished, we can transfer current package build & storage to it. Here it is: https://build.opensuse.org/package/show/home:Vitzy/flameshot . Also current CI building service should be keepped for daily building & testing.

AlfredoRamos:
@ZetaoYang I don't know how the OBS works however from what I've seen it does not build nor package the application on each commit/pull request, does it allow to do so?

Travis CI (or another CI service) indeed should be kept, because I don't think OBS can replace the functionality of a CI service.

I was thinking in file.io or 0x0.st as a replacement, but the OBS looks good, if it supports at least daily builds.

JavierGelatti:
It seems that transfer.sh isn't discontinued (anymore)!

See this announcement on [their webpage](https://transfer.sh/):
captura de pantalla de 2018-11-09 17-50-59

AlfredoRamos:
@JavierGelatti That's a good news!

So this isn't needed after all.

PR #422, opened 2018-12-15
README: add repology badge inside collapsed div

luzpaz:
Using https://repology.org/metapackage/flameshot/badges it's possible for users to see what version of flameshot are available in the package ecosystem. This badge is nested in an expandable div so it doesn't overwhelm the README with its size. This commit also fixes a minor typo
FYI HTML was used instead of markdown due to the markdown not being displayed.

Example of how this looks see https://github.com/luzpaz/flameshot/tree/repology#installation

Edit:fixed incorrect url

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,14 @@ There are packages available for a few distros:
 - [Docker](https://github.com/ManuelLR/docker-flameshot)
 - Fedora: `dnf install flameshot`
 - [Snap/Flatpak/AppImage](https://github.com/flameshotapp/packages)
-- Besides, generic packages available via [opensuse software  repertory](https://software.opensuse.org//download.html?project=home%3AVitzy&package=flameshot)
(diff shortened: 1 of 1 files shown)


luzpaz:
TravisCI is complaining about an unrelated [issue](https://travis-ci.org/lupoDharkael/flameshot/jobs/468405022#L1452-L1462):
```
[FAILED] cups-libs-2.2.8-6.fc29.x86_64.rpm: No more mirrors to try - All mirrors were already tried without success
(11-12/91): libdrm-d  4% [-                   ] 417 kB/s | 4.1 MB     03:41 ETA
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
Error: Error downloading packages:
  Cannot download Packages/c/cups-libs-2.2.8-6.fc29.x86_64.rpm: All mirrors were tried
sudo: yum-builddep: command not found
make: *** [/pack//rpm.mk:124: package] Error 1
make: Leaving directory '/source'
```

lupoDharkael:
The CI problem should be fixed now. Thanks for the contribution!

Issue #431, opened 2018-12-25
support for more image format outputs


deathtrip:
Would it be possible to add support for saving a screenshot in more than just png format?
As formats such as jpeg are more efficient when it comes to file sizes.
Maybe using ImageMagick or some image library to convert them on the fly, if they are installed.

Issue #278, opened 2018-07-22
No image extension while saving gives error


arslanfarooq:
**Describe the bug**
When saving the screenshot, if no file extension is giving it throws error.

**To Reproduce**
- take a screenshot
- file save dialog will open
- start giving a custom name for the screenshot to describe it
- in above step, when you start typing it removes the auto generated file name, along with the png extension, of course
- now if you try to save the screenshot, it generates an error [(screenshot)](https://imgur.com/a/b8ywRvk)
- you have to re-type the custom name you gave, **with** an extension to save the file

**Expected behavior**
If no file extension is giving it may default to png

**Desktop:**
 - OS: Arch Linux
 - Version: bit

PR #436, opened 2019-01-01
Add support for more images formats

b00m-b00m:
Now when user wants to save screenshot it is possible to save it not only as png but with support of several image formats.
Resolves [#431](https://github.com/lupoDharkael/flameshot/issues/431)

### DIFF
--- a/src/utils/screenshotsaver.cpp
+++ b/src/utils/screenshotsaver.cpp
@@ -60,14 +60,11 @@ bool ScreenshotSaver::saveToFilesystemGUI(const QPixmap &capture) {
         QString savePath = QFileDialog::getSaveFileName(
                     nullptr,
                     QString(),
-                    FileNameHandler().absoluteSavePath() + ".png");
+                    FileNameHandler().absoluteSavePath() + ".png",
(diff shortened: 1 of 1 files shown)


src/utils/screenshotsaver.cpp
```diff
@@ -60,16 +60,13 @@ bool ScreenshotSaver::saveToFilesystemGUI(const QPixmap &capture) {
...
                     QString(),
-                    FileNameHandler().absoluteSavePath() + ".png");
+                    FileNameHandler().absoluteSavePath() + ".png",
+					QLatin1String("Portable Network Graphic file (PNG) (*.png);;BMP file (*.bmp);;JPEG file (*.jpg)"));
 
         if (savePath.isNull()) {
             break;
         }
 
-        if (!savePath.endsWith(QLatin1String(".png"))) {
```
AlfredoRamos: Wouldn't be better to verify if the file has a valid extension instead of removing that check?

That piece of code was added to address #278 this would introduce a regression.

b00m-b00m: Well... I have tested my code in Windows, and it's works fine even in described case (extension adding automatically to string) and I thought shoud be the same behavior for ahother OSs but its not true indeed. I have added mentioned verification for all supported images format.

Issue #437, opened 2019-01-02
A simpler way to draw horizontal/vertical/45° lines/arrow


dantonini:
I'm using 0.6.0 version. The tool is very simple to use, fast and quite complete. Thanks for developing this tool.

Now I'm a little bit struggled with drawing some horizontal/vertical arrows. Line tool allows to draw horizontal line easily (pin automatically) but vertical lines do not have the same facility, Arrow tool doesn't pin horizontally. 
A nice to have feature is a more simpler way to draw straight lines/arrows. I found libreoffice draw interaction is simple and effective: the SHIFT key allow to draw, horizontal/vertical/45° lines/arrow.

What do you think about?

Cheers

borgmanJeremy:
This works with the CTRL Key now.

PR #439, opened 2019-01-03
Add 45-multiple degree adjustment for line, arrow and marker tools

magnickolas:
This PR adds a feature proposed in #437. More specifically, "Line", "Arrow" and "Marker" tools' lines are now adjusted not only in a horizontal direction but also vertical and those of the bisectors' of the angles between them.

### DIFF
--- a/src/tools/abstracttwopointtool.cpp
+++ b/src/tools/abstracttwopointtool.cpp
@@ -16,6 +16,29 @@
 //     along with Flameshot.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "abstracttwopointtool.h"
+#include <cmath>
+
+namespace {
+
+const double ADJ_UNIT = std::atan(1.0);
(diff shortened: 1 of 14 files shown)


lupoDharkael:
I think it would be good to have this feature only effective while pressing Ctrl (KolourPaint does this) it can be kind of frustrating if you need a specific angle and you can't draw it because of the auto-adjustment.
If you add a check for Ctrl pressed for the automatic adjustment I'll happily merge this.
Sorry for the delay of the review, my university has no mercy this time of the year :P

magnickolas:
I agree with you. I've added a check for pressed Ctrl and generalized an adjustment for the two-point tools (the same functionalily for a line/arrow/marker + an ability to draw a symmetric selection/rectangle/circle).

lupoDharkael:
thanks!

PR #443, opened 2019-01-07
utils/confighandler.cpp: Enable Pin and Text tool by default

hosiet:
This commit enables all the available tools by default so that users may use them out-of-box without bothering with modifying configuration.


lupoDharkael:
Seem reasonable.

PR #448, opened 2019-01-15
Check if filename contains invalid characters

### LINKED ISSUE #409: 1 line Bug Fix - April 23 commit prevents using : in ISO 8601 extended time formats
While #196  makes a case for changing the time default, the associated 17ac6f7a352f136b7470b66f536d372861fc6d8d completely prevents using colons often used in [ISO Extended Time formats](https://www.cl.cam.ac.uk/~mgk25/iso-time.html). 

It also creates an unexpected user experience when colons can not be entered and are changed to dashes before the user's eyes.

As colons are not a problem in Linux file names, and *all* my image files have them ;-), I suggest keeping the new default, but allowing users to enter colons by undoing the second part of 17ac6f7a352f136b7470b66f536d372861fc6d8d as follows:

https://github.com/lupoDharkael/flameshot/blob/94daa4fecdfaf6188853625d4798b4c5804a20a3/src/utils/filenamehandler.cpp#L48 
```cpp
48>     res = res.replace(QLatin1String("/"), QStringLiteral("⁄"));
```
Thank you for such a well designed utility!

AlfredoRamos:
Currently it only have conditions for Windows and Linux/Unix, checks are done though regular expressions in the filename and invalid characters are replaced with underscores in the filename pattern.

Fixes #409

### DIFF
--- a/src/utils/filenamehandler.cpp
+++ b/src/utils/filenamehandler.cpp
@@ -24,6 +24,11 @@
 
 FileNameHandler::FileNameHandler(QObject *parent) : QObject(parent) {
     std::locale::global(std::locale(""));
+#ifdef Q_OS_WIN
+    invalidCharacters = QRegularExpression("[<>:\"/\\\\|?*]+");
+#else //Q_OS_LINUX
+    invalidCharacters = QRegularExpression("[/]+");
+#endif
 }
 
 QString FileNameHandler::parsedPattern() {
@@ -32,9 +37,11 @@ QString FileNameHandler::parsedPattern() {
 
 QString FileNameHandler::parseFilename(const QString &name) {
     QString res = name;
+
     if (name.isEmpty()) {
         res = QLatin1String("%F_%H-%M");
     }
+
     std::time_t t = std::time(NULL);
 
     char *tempData = QStringTocharArr(res);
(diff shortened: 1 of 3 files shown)


AlfredoRamos:
This is a proof of concept and work in progress, please do not merge it yet.

Feedback is welcome.

@jr81 @tastytea

P.S.: MacOSX was not added because I've read that the colon (`:`) is valid in recent versions.

tastytea:
According to [this comment from 2015](https://apple.stackexchange.com/a/173621), GUI programs on macOS will map `:` to `/` and `/` to `:`. That could be a bit confusing if that's still the case.

src/utils/filenamehandler.cpp
```diff
@@ -24,6 +24,11 @@
 
 FileNameHandler::FileNameHandler(QObject *parent) : QObject(parent) {
     std::locale::global(std::locale(""));
+#ifdef Q_OS_WIN
+    invalidCharacters = QRegularExpression("[<>:\"/\\\\|?*]+");
```
tastytea: Because of the '+' at the end of the expression, consecutive invalid characters will be replaced by a single underscore. `:::` -> `_` for example. Not sure if that's a problem.

AlfredoRamos: Yep, I added precisely for that, otherwise `<*\` it will generate multiple `___`.

src/utils/filenamehandler.cpp
```diff
@@ -44,8 +51,11 @@ QString FileNameHandler::parseFilename(const QString &name) {
     res = QString::fromLocal8Bit(data, (int)strlen(data));
     free(tempData);
 
-    // add the parsed pattern in a correct format for the filesystem
-    res = res.replace(QLatin1String("/"), QStringLiteral("⁄")).replace(QLatin1String(":"), QLatin1String("-"));
+    // replace invalid characters with underscores
+    if (res.contains(invalidCharacters)) {
```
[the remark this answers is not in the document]
AlfredoRamos: Right

src/utils/filenamehandler.cpp
AlfredoRamos: @Rmano yeah, is hard to fix the problem in that kind of scenario, a user always has the option to manually copy the file generated in Linux/Unix to Windows, and in that case there's nothing the application can do.

PR #449, opened 2019-01-15
Fix Exec in .desktop file

SiqingYu:
https://github.com/lupoDharkael/flameshot/issues/90


docs/desktopEntry/package/flameshot.desktop
```diff
@@ -17,7 +17,7 @@ Keywords=flameshot;screenshot;capture;shutter;
 Keywords[zh_CN]=flameshot;screenshot;capture;shutter;截图;屏幕;
 Keywords[fr]=flameshot;capture d'écran;capter;shutter;
 Keywords[ja]=flameshot;screenshot;capture;shutter;スクリーンショット;キャプチャー;
-Exec=flameshot
+Exec=flameshot gui
```
ZetaoYang: I disagree with the above change. It doesn't need to change flameshot `Exec` command.

PR #462, opened 2019-02-04
shortcut description

noraj:
see https://github.com/lupoDharkael/flameshot/issues/271


lupoDharkael:
Thanks for the contribution!

PR #469, opened 2019-02-12
Update russian translation

alok1111:
default
By the way, it would be nice to have an ability to translate "SPACEBAR" text.

### DIFF
--- a/translations/Internationalization_ru.ts
+++ b/translations/Internationalization_ru.ts
@@ -401,7 +401,7 @@ Press Space to open the side panel.</source>
     <message>
         <location filename="../src/widgets/infowindow.cpp" line="63"/>
         <source>Right Click</source>
-        <translation>ПКМ</translation>
+        <translation>Правый клик</translation>
     </message>
     <message>
         <location filename="../src/widgets/infowindow.cpp" line="64"/>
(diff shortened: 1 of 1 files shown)


DaVukovic:
> By the way, it would be nice to have an ability to translate "SPACEBAR" text.

That's possible. At least I got it managed at my PR. Please take a look at:

https://github.com/DaVukovic/flameshot/blob/6d278c0fc7281de998e2f8a17b11659282b8b9d6/src/widgets/infowindow.cpp#L62

and then:

https://github.com/DaVukovic/flameshot/blob/6d278c0fc7281de998e2f8a17b11659282b8b9d6/translations/Internationalization_de_DE.ts#L401-L405

That should be exactly what you are searching for.

alok1111:
> Hi! IMHO, it's shorter to make such translation.
> Показать панель с настройками выбранного инструмента
Дословно sidebar это всё-таки боковая панель

greno4ka:
Yes, you're right, but first of all it doesn't matter sidebar it is or just panel. The second thing is, that it's quite long for that small window. Even english variant is not fully readable in default window, so you need to expand it to read about this feature. That takes a time. I don't insist on this change. Just suggested to think about it.

lupoDharkael:
Thanks for the contribution!

PR #473, opened 2019-02-21
Allow notification to be dragged and dropped

magnus-gross:
The x-kde-urls field in the hints map allows us to provide a direct URL to the saved screenshot. This will allow KDE to show a thumbnail and will also enable drag and drop functionality.

img

### DIFF
--- a/src/utils/screenshotsaver.cpp
+++ b/src/utils/screenshotsaver.cpp
@@ -49,7 +49,7 @@ bool ScreenshotSaver::saveToFilesystem(const QPixmap &capture,
         saveMessage = QObject::tr("Error trying to save as ") + completePath;
     }
 
-    SystemNotification().sendMessage(saveMessage);
+    SystemNotification().sendMessage(saveMessage, ok ? completePath : "");
     return ok;
 }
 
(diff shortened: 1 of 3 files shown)


magnus-gross:
I just realized that somehow the drag and drop context only offers a symbolic link option, while Spectacle offers options for moving, copying and linking.

Please do not merge yet, while I investigate what they are doing differently.

Related source code:
https://github.com/KDE/knotifications/commit/0092619e40b922e3c551fc26d225fa1635dd441f
https://github.com/KDE/spectacle/commit/945548e8c63f02c39aef492d31d8942ee6063bd6

magnus-gross:
The KDE foks were kind enough to help me out, I was actually missing the `file:///` prefix.
Now everything is working properly.

You can consider this PR for review now.

src/utils/screenshotsaver.cpp
```diff
@@ -49,7 +49,7 @@ bool ScreenshotSaver::saveToFilesystem(const QPixmap &capture,
         saveMessage = QObject::tr("Error trying to save as ") + completePath;
     }
 
-    SystemNotification().sendMessage(saveMessage);
+    SystemNotification().sendMessage(saveMessage, ok ? completePath : "");
```
lupoDharkael: Maybe this could be changed to prevent double checking `ok` for branching?

magnus-gross:
I have fixed the double branching by using an extra variable for the notification url. Theoretically, I could just have used `completePath` and set it to `""`, but I don't think that this would fit really well and future code might want to use `completePath` after the branching.
Let me know, if you had something else in mind.

lupoDharkael:
Looks good to me!

PR #474, opened 2019-02-22
Add translation of system buttons in dialogs

greno4ka:


greno4ka:
Result of this PR is correctly translated buttons like OK, Cancel in your locale language.

lupoDharkael:
Awesome I didn't know about this. Thanks!

PR #477, opened 2019-02-22
Fixup working with startupLaunch value

greno4ka:
This commit fixes problem, when you want to import or export
information about startup.

So now program can write this value to settings, and then, when it will read it, it first of all analyses if settings value corresponds to file existence, and corrects the situation according to settings' value.

### DIFF
--- a/src/utils/confighandler.cpp
+++ b/src/utils/confighandler.cpp
@@ -242,6 +242,21 @@ void ConfigHandler::setKeepOpenAppLauncher(const bool keepOpen) {
 
 bool ConfigHandler::startupLaunchValue() {
     bool res = false;
+
+    if (m_settings.contains(QStringLiteral("startupLaunch"))) {
+        res = m_settings.value(QStringLiteral("startupLaunch")).toBool();
+    }
+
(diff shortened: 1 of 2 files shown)


greno4ka:
So, @lupoDharkael, can you regard this feature?

greno4ka:
I just noticed, that on clean system, `.config/autostart` can be absent. So the second commit is about verifying the existence of this dir and it's creation, if nessesary. This can be related with many issues about autostart problems.

greno4ka:
Travis has error, not tied with my contribution.

curl: (35) gnutls_handshake() failed: Error in the pull function.
The command "curl --upload-file dist/flameshot_0.6.0_fedora27_x86_64.rpm https://transfer.sh/flameshot_0.6.0_fedora27_x86_64.rpm" failed 3 times.

lupoDharkael:
I'm reviewing this tomorrow (sorry for the wait but lately it's hard for me to take time).

greno4ka:
No problem, @lupoDharkael.

lupoDharkael:
Seems right to me. Thanks!

PR #482, opened 2019-02-28
img/material/*/text.png: Remove executable bits

hosiet:
This commit removes the unnecessary executable bits from the two images.


lupoDharkael:
Thanks!

Issue #485, opened 2019-03-04
[Feature request] highlighter pen improvements


adasomg:
1) Allow for organic drawing with it rather than only straight lines (this would more naturally be toggleable while, say, holding shift) 

2) Right now the highlight is basically drawn onto the screenshot as a color with reduced opacity. This makes the text less readable.
In editing software like Photoshop you'd draw the highlight on a separate layer and set the blending mode to 'multiply'  so that the (usually black/dark on white) text remains crisp while the background gets highlighted. QPainter seems to support blending modes.

magnus-gross:
For 1.: You can already do freehand drawing with the very first tool (The pencil)

PR #490, opened 2019-03-06
Highlight with composition mode set to Multiply

magnus-gross:
This improves readability of marked text.

Before:
before

After:
after


magnus-gross:
See #485

greno4ka:
Cool!

ZetaoYang:
Good improvement!

lupoDharkael:
Awesome! thanks!

PR #495, opened 2019-03-28
Fix problem with disabled tray icon

greno4ka:
Steps to reproduce bug:
1.) reset settings to default
2.) export configuration
3.) set disabled tray icon
4.) import default configuration (with enabled tray icon)

Result: tray icon still disabled

Expected result: enabled try icon

Problem was in the fact, that showTrayIconChanged was executed
only when user has clicked on that checkbox. Now it run even
while importing config file.


greno4ka:
typo in commit: s/try/tray/

greno4ka:
Travis again failed because of repo errors. It's not my fault.

greno4ka:
Wuzzup, @lupoDharkael? It's very trivial patch!

lupoDharkael:
Thanks for the contribution!

PR #516, opened 2019-05-01
Add a config option to auto-save a screenshot to a folder after copying it to the clipboard

joecharamut:
Uses the filename generated from the pattern in the filename editor to save the screenshot after pressing <Enter> or the copy button. 

User is able to change the save path, and enable or disable it in the General config tab

The save path is default to the Pictures folder


greno4ka:
> void saveAfterCopyChanged(bool checked);
Oh, that's really bad variable name. Maybe try "isChecked" or something like that?

joecharamut:
I just named my variable consistent with the other ones in the file. They all use `checked` as the parameter name as well.

https://github.com/lupoDharkael/flameshot/blob/387adc3e356264ca1ff7cfbff6040402cf3bc893/src/config/geneneralconf.h#L37-L41

src/config/geneneralconf.cpp
```diff
@@ -49,6 +53,14 @@ void GeneneralConf::updateComponents() {
     m_sysNotifications->setChecked(config.desktopNotificationValue());
     m_autostart->setChecked(config.startupLaunchValue());
     m_closeAfterCapture->setChecked(config.closeAfterScreenshotValue());
+    m_saveAfterCopy->setChecked(config.saveAfterCopyValue());
+
+    if (!config.saveAfterCopyPathValue().isEmpty()) {
+        m_savePath->setText(config.saveAfterCopyPathValue());
+    } else {
+        ConfigHandler().setSaveAfterCopyPath(
```
DanielChabrowski: It's a little weird that default save-after-copy path is set only after opening configuration window.

joecharamut: Well since you need to open the configuration window to enable the feature, I thought it was a fairly reasonable place to load that field. There probably is a better place though.

ElijahLynn:
To clarify, this will provide an option to save the bitmap to the clipboard, not just the path, correct?

lethalbrains:
+1 for merging this pull request.

ElijahLynn:
Would be really helpful to see some screenshots of the new UI and behavior.

arjunv:
As much as I'd love to see this feature pulled, it's leaving flameshot stuck for around 10 seconds before saving the file (and copying to clipboard).

arjunv:
> To clarify, this will provide an option to save the bitmap to the clipboard, not just the path, correct?

Correct. It adds these config options which does that!
image

joecharamut:
Thanks for pointing out the double notification, I keep them turned off normally. This new commit should fix it by disabling the first "Capture saved to clipboard" notification.

ElijahLynn:
> > To clarify, this will provide an option to save the bitmap to the clipboard, not just the path, correct?
> 
> Correct. It adds these config options which does that!
> image

Thank you for this screenshot, it gave me some motivation to want to test this PR out soon!

ElijahLynn:
@joecharamut If you have time to get this updated, I should have some time soon to get this tested. I am missing this feature from time to time, as I have to open my clipboard manager (CopyQ) then drag the copy to a browser tab which gives me a `/tmp/file.jpg` that I can then copy and use in the application that requires a file. Would be much easier if I could just skip that whole step. 

And I see [Arch compile instructions](https://github.com/lupoDharkael/flameshot#arch) now so maybe I can just checkout your branch and test it as is. I suppose that is the next step.

joecharamut:
I think this should update it.

arjunv:
> And I see [Arch compile instructions](https://github.com/lupoDharkael/flameshot#arch) now so maybe I can just checkout your branch and test it as is. I suppose that is the next step.

If you use an aur helper, install [`flameshot-git`](https://aur.archlinux.org/packages/flameshot-git/) and edit the PKGBUILD file to use the commit directly.

image

Additionally add the following entry in `pacman.conf` so that the package won't be automatically updated and reverting to master branch.
```
IgnorePkg   = flameshot-git
```

ElijahLynn:
> If you use an aur helper, install [`flameshot-git`](https://aur.archlinux.org/packages/flameshot-git/) and edit the PKGBUILD file to use the commit directly.

Thanks for this, I got it installed this way! I added further steps for anyone else on Arch that wants to test too:
1. `yay --sync flameshot-git` # this will remove `flameshot` so no need to do that separately
1. Edited `/home/elijah/.cache/yay/flameshot-git/PKGBUILD` and modified with commit variable change that @arjunv [suggested above](https://github.com/lupoDharkael/flameshot/pull/516#issuecomment-626119096)
1. `cd /home/elijah/.cache/yay/flameshot-git`
1. `makepkg --install` 
1. Edited `/etc/pacman.conf` and put the `IgnorePkg   = flameshot-git` in the `[options]` section. Then reran `yay --sync flameshot-git` and it said there we no updates.

ElijahLynn:
Testing on https://github.com/lupoDharkael/flameshot/pull/516/commits/dfde97e88e616995ea4b16950ec55db43ce93904:

I turned on all the options, as that is what I actually want. 
* Close after capture
* Copy URL after upload # Although I think I am misreading this, I want to copy the local path after upload (in addition to the image) so they are both in my clipboard manager (CopyQ)
* Save image after copy

This resulted in a blank image being saved to the clipboard, and subsequent screenshots didn't save either. The close after capture didn't seem to work either. I do wonder if this PR has extra options in it or if those are already just in latest master. 

At any rate, I unchecked "Close after capture" and "Copy URL after upload" and just left "Save image after copy" and that part is working perfectly!

ElijahLynn:
K, so just "close after capture" checked seems to cause the issue with blank screenshots. Not sure if this is a bug in master yet or because of a change in this PR.

joecharamut:
Strange, that isn't happening on my computer with master or the PR.

arjunv:
> K, so just "close after capture" checked seems to cause the issue with blank screenshots. Not sure if this is a bug in master yet or because of a change in this PR.

Am able to reproduce the same. 

After enabling `Close after capture` option, the only working functionality is saving the screenshot to a file - either through normal `Save Image` function or using `Copy image to clipboard` function with `Save image after copy` option enabled. The results were the same when using Keyboard shortcuts or the GUI buttons.

Functionality Matrix when `Close after Capture` option in enabled while running on [dfde97e](https://github.com/lupoDharkael/flameshot/commit/dfde97e88e616995ea4b16950ec55db43ce93904)
| Action                               | Image Saved | Image Copied to Clipboard | Image Uploaded | Image URL Copied to Clipboard |
| ------------------------------------ | ----------- | ------------------------- | -------------- | ----------------------------- |
| Save Image                           |      ✓      |                           |                |                               |
| Copy Image to Clipboard              |      ✗      |             ✗             |                |                               |
| Copy Image to Clipboard + Save Image |      ✓      |             ✗             |                |                               |
| Upload Image                         |             |                            |        ✗       |                               |
| Upload Image + Copy URL to Clipboard |             |                            |        ✗       |               ✗               |

joecharamut:
Okay after more testing I can also reproduce it on the branch, will try on master soon

joecharamut:
I am having the same issues on master so I don't think this is my PR's fault

edgarmaloverian:
+1 for this feature and to wake up this thread :)

ElijahLynn:
This PR works, for what its worth. I have been using it since May 13th and it is so VALUABLE! So valuable that I will send you $5 @joecharamut if you send me a link to do that (sign up for GitHub Sponsors or verify your GitHub with Brave Rewards etc).

borgmanJeremy:
Thanks for the detailed work on this PR, I am looking to get it merged into master before our 0.8 release. 

It seems like there is an issue related to close after capture? I think this is a confirmed bug per: #557

joecharamut:
The close after capture issues were unrelated to this PR, but I can't get the problem to happen on the current master

joecharamut:
@ElijahLynn 
I'm glad it's useful. I don't have a good way to accept donations but I appreciate the sentiment

arjunv:
> 5. Edited `/etc/pacman.conf` and put the `IgnorePkg   = flameshot-git` in the `[options]` section. Then reran `yay --sync flameshot-git` and it said there we no updates.

Reminder to remove the same, now that the feature has been committed to master :D

ElijahLynn:
W00t! Merged, thank you!!!!! This is amazing!

esantosbr:
Hi all,

I read all this posts and other ones but I could not find how to change the 'Enter" key funcion.
Into flameshot gui when we select an area and hit enter I'd like to save it to a folder and not only to clipboard. (It'd be better than hiting CTRL+S
How can I change this behaviour?

Issue #523, opened 2019-05-08
Feature: Select all on GUI start


weishiuchang:
First, thank you for this application and all the work that has gone into it. Second, just to clarify, this is not a rehash of #324 or #5 

I'm looking to use flameshot in a screen-annotation use case rather than as a pure screenshot program. By that I mean I would like to give some talks/presentations/demonstrations, and in the middle of the presentation I would like to do something akin to:

1. hit print screen to bring up "flameshot gui --select-all"
2. annotate/draw on/mark up my presentation on the screen for emphasis
3. hit esc and continue with the next slide

I know this is getting away from the core purpose of the Flameshot tool, but I think adding an option to the gui startup to select-all would be easier than select-window (see discussion in #5), and could help engender the Flameshot tool to presenters/teachers.

Of course having the pencil tool be automatically selected would be intuitive as well in this use case since there is no need to resize/move the selection box as the entirety of the screen is already selected on start.

Let me know what you think.

WeiShiuChang:
I have implemented the described use case.

nachovizzo:
I guess so, as long as I'm able to annotate the screen with the GUI I will be happy (#273). I did have a [solution](https://github.com/flameshot-org/flameshot/issues/273#issuecomment-613660060) but it is not working anymore

mmahmoudian:
@nachovizzo in your solution, you are making it too complicated. This should easily work for you:

```sh
flameshot gui; sleep 3s; xdotool key ctrl+a
```

borgmanJeremy:
Xrandr uses WxH,x,y (without spaces, mobile keyboard insists on adding those spaces)

PR #536, opened 2019-05-18
Implement gui --interactive option (#523)

WeiShiuChang:
* document --interactive
 * pass interactive mode through the controller
 * implement selectAll()
 * update graphicCapture with interactive boolean

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ Example commands:
     flameshot gui -d 2000
     ```
 
+- Open GUI with entire screen selected and pencil tool picked
+
+    ```shell
+    flameshot gui -i
+    ```
+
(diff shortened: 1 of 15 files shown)


WeiShiuChang:
Need alternate default button selection.

ElijahLynn:
Can you explain more about what this does? I feel like this is coming from knowing about a previous system.

WeiShiuChang:
It's really just so that a presenter can hit printscreen and immediately start drawing/highlighting on the screen. It's very useful for powerpoint presentations and/or demonstrations. See #523

nachovizzo:
I'd love to have this feature, is there any way we can work on it?

veracioux:
@nachovizzo I've opened a discussion: #1974. Please leave your votes/suggestions there.

PR #538, opened 2019-05-19
add config file location

Trollwut:
Add the dotfile location into the readme.


borgmanJeremy:
The config file has been changed to be in the standard XDG location. Thanks for helping document the old location.

PR #545, opened 2019-05-26
dbus/org.dharkael.Flameshot.xml: Mark id as uint

JohnAZoidberg:
In `src/core/flameshotdbusadapter.h` the id is marked as uint.
All other types in the xml are correct.

Took me a while to figure out that I had to use unsigned int.


borgmanJeremy:
@hosiet I think this still applies with all our changes still but I'd like your opinion too. Will changing this break existing setups?

hosiet:
This looks like a nick catch of bugs. I don't think it will break existing setup.

borgmanJeremy (approved):

Issue #548, opened 2019-05-31
Feature: Clear button


WeiShiuChang:
I'm interested in a "clear all modifications" button and/or hotkey. This would allow me to simply clear all instead of undo multiple times.

WeiShiuChang:
Pull request submitted. (#550)

PR #550, opened 2019-06-01
Add Clear button (#548)

weishiuchang:
* implement clear button
 * add clear button graphics
 * add button to gui
 * attach delete key to clear button


luzpaz:
Do all the individual `.ts` needs to be modified? Doesn't just one file get updated and then the rest of the language.ts files are generated from it?

WeiShiuChang:
Sorry - I'm not that familiar with QT5 i18n commands. I ran the command `lupdate-qt5 -verbose flameshot.pro` and those were the changes it made.

luzpaz:
@WeiShiuChang actually that sounds about right. Never mind me.

AlexP11223:
I think currently translations are usually updated in separate PRs, otherwise you will get lots of conflicts.

borgmanJeremy:
Is there a reason to add a clear button now that undo is implemented? IE repeating undo is the same as clear?

weishiuchang:
I personally found use in a clear all, but that's just my own use case. But I'm not fussed either way.

AlexP11223:
> We need to balance new features against complaints that there are too many icons :)

Maybe this icon could be disabled by default.

PR #578, opened 2019-07-09
Changed enter/return key to save not copy

DesertPunk:
simple change that makes the enter key save the screenshot to filesystem and not to the clipboard


borgmanJeremy:
Hot keys are now configurable.

PR #592, opened 2019-08-20
Enable hungarian translation

kikadf:
Hungarian translation is exist, but not enabled default


borgmanJeremy:
This was resolved during the CMake conversion.

Issue #582, opened 2019-07-18
How to make screenshot with delay


Ygarr:
How to make screenshot with delay to make screenshot of e.g. pop-up notifications or descriptions when hovering or focusing on a link using the mouse. Like "scrot -S -d 5" does.
![2019-07-19-022905_422x82_scrot](https://user-images.githubusercontent.com/14309871/61498724-122f7280-a9cd-11e9-892d-86da67cba51e.png)

AlexP11223:
looks like there is no option for this in GUI but you can do this via terminal, there are examples in README: 

```
flameshot gui -d 2000
```

AlexP11223:
Also the README has info about adding global hotkeys on KDE, including the delayed capture via Ctrl+PrtSc https://github.com/lupoDharkael/flameshot#on-kde-plasma-desktop.

You of course can also add a shortcut ("launcher" on Linux), put it on desktop, etc. or modify the default shortcut.

For example, in GNOME you can add an action entry like this in `/usr/share/applications/flameshot.desktop` (you need sudo to edit this file, so for example `sudo gedit /usr/share/applications/flameshot.desktop`):

```diff
@@ -23,7 +23,7 @@
 Type=Application
 Categories=Graphics;
 StartupNotify=false
-Actions=Configure;Capture;
+Actions=Configure;Capture;DelayedCapture;
 
 [Desktop Action Configure]
 Name=Configure
@@ -42,3 +42,7 @@
 Name[ja]=スクリーンショットを撮る
 Name[sr]=Сними екран
 Exec=flameshot gui --delay 500
+
+[Desktop Action DelayedCapture]
+Name=Take screenshot after delay
+Exec=flameshot gui --delay 5000
```

Alt + F2 -> r -> Enter to restart Gnome and it will appear when right clicking in Gnome app menu, etc. (sadly not in the tray icon menu, it's hard-coded in the source code)

![](https://i.imgur.com/Q1gpYfL.png)

AlexP11223:
... also looks like there is Open Launcher menu item in the current master (not yet released) which opens a window that allows to specify delay.

![](https://i.imgur.com/9RRdkC9.png)

xstable:
When will it be released?

AndydeCleyre:
[...]

PR #593, opened 2019-08-22
make the gui -d example more clear

AlexP11223:
The description was a bit confusing, some people may not understand what is "GUI" in this context, such as possibly #582


lupoDharkael:
Thanks!

PR #609, opened 2019-09-15
Add zsh completion support

kapsh:
Both bash and zsh completions could be supported this way.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -148,8 +148,11 @@ configure_file(${CMAKE_SOURCE_DIR}/docs/desktopEntry/package/flameshot.desktop
 configure_file(${CMAKE_SOURCE_DIR}/docs/appdata/flameshot.metainfo.xml
                ${CMAKE_CURRENT_BINARY_DIR}/share/metainfo/flameshot.metainfo.xml COPYONLY)
 
-configure_file(${CMAKE_SOURCE_DIR}/docs/bash-completion/flameshot
-               ${CMAKE_CURRENT_BINARY_DIR}/share/bash-completions/completions/flameshot COPYONLY)
+configure_file(${CMAKE_SOURCE_DIR}/docs/shell-completion/flameshot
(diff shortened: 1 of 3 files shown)


kapsh:
Need some help with installation. ZSH requires file named as `_flameshot` to load it automatically. But I couldn't find good way to do that in qmake. 
I've tried to add `extra` member with `cp` call as described in https://doc.qt.io/qt-5/qmake-advanced-usage.html#installing-files but that does not work as expected (looks like it overrides install target instead of prepending it).

borgmanJeremy:
Since we switched to cmake it should be trivial to change the name. Do you have time to look at it? I can help.

kapsh:
Thanks for the update, cmake should be easier to implement this. I'll try to do that on this week.

borgmanJeremy:
We are trying to do a feature freeze for the 0.8 release at the end of this weekend, so if you can have it done by then we can get it into 0.8

src/CMakeLists.txt
```diff
@@ -148,8 +148,11 @@ configure_file(${CMAKE_SOURCE_DIR}/docs/desktopEntry/package/flameshot.desktop
 configure_file(${CMAKE_SOURCE_DIR}/docs/appdata/flameshot.metainfo.xml
                ${CMAKE_CURRENT_BINARY_DIR}/share/metainfo/flameshot.metainfo.xml COPYONLY)
 
-configure_file(${CMAKE_SOURCE_DIR}/docs/bash-completion/flameshot
-               ${CMAKE_CURRENT_BINARY_DIR}/share/bash-completions/completions/flameshot COPYONLY)
```
kapsh, line 152: Possible typo here. I don't know much about bash, but it is `/usr/share/bash-completion` without s on my systems (and in comment above).

kapsh:
@borgmanJeremy need a hint about rpm spec. Should I also add a newly installed file into `%files` there?

borgmanJeremy:
I Believe so ...@ZetaoYang is the packaging expert

kapsh:
It is done then.

borgmanJeremy (approved):

Issue #554, opened 2019-06-07
Make "noBorder" possible for Screenshots


xstable:
At the Moment if you take a screenshot (rectangle area of the screen) it has an green border around.
Would be nice, if we could toggle this like needed.

Because sometimes I make a screenshot, pin it... move it to another point at the screen, take another screenshot (with the pinned one inside) and so on.
In this case, the Green Borders are a huge downside.

Se this image for an example... I've sliced 3 lines in an phpmyadmin web-browser window and concat them: 

![image](https://user-images.githubusercontent.com/1778068/59091582-f0dc6080-890f-11e9-8464-a82ba202a65d.png)

AlexP11223:
Yeah, I had the same desire yesterday, so I submitted PR.

lr222gw:
Would be nice to be able to set this through GUI, 
meanwhile this can be achived by modifying the .cfg file; **.config/Dharkael/flameshot.ini**
```                                                                         
[General]
buttons=@Variant(\0\0\0\x7f\0\0\0\vQList<int>\0\0\0\0\x12\0\0\0\0\0\0\0\x1\0\0\0\x2\0\0\0\x4\0\0\0\x5\0\0\0\x6\0\0\0\x12\0\0\0\xf\0\0\0\a\0\0\0\b\0\0\0\t\0\0\0\x10\0\0\0\n\0\0\0\v\0\0\0\f\>
contastUiColor=#008080
contrastOpacity=204
disabledTrayIcon=false
drawColor=#000000
drawThickness=0  
showHelp=true  
uiColor=#006e6e6e
```

**Change uiColor from**; 
#6e6e6e      <-- Some six digit hexvalue 
**To**
#006e6e6e <-- Eight digit hexvalue; the first digits should be 00, this will set the opacity to 0%...

xstable:
Awesome Applikation. 
Hope @lupoDharkael will find some time to merge some of the PR's.
I think it will boost the Applikation a lot. There is still no better Screenshot-App for linux out there.

PR #622, opened 2019-10-01
Allow setting pin border width (or removing it), #554

AlexP11223:


Martin-Eckleben:
Looks fine to me but I'm not yet experienced enough to decide I think. @borgmanJeremy should we merge?

lupoDharkael:
I guess this is not a bad thing to add but we should keep in mind that adding individual settings for specific tools might bloat the config GUI. I know there is demand for a few additional settings for specific tools. What I was thinking to add is that every tool would have a config widget. When you click the name of the tool (not the checkbox) it would show the specific widget with the settings for that tool in a contextual fashion.
I'm ok with this as long as we don't keep adding options in the main configuration GUI.

borgmanJeremy:
Lets roll this in with the discussions on #954

MyriaCore:
Just wanna iterate on some of the design convo's that have been happening in here:

> I know there is demand for a few additional settings for specific tools. What I was thinking to add is that every tool would have a config widget. When you click the name of the tool (not the checkbox) it would show the specific widget with the settings for that tool in a contextual fashion.

If we did have per-tool settings be related to the selection menu, I think instead of clicking the name, we should display a settings icon (like a cog or a wrench) that floats to the right of the menu, in the yellowed-out area of this screenshot for example:

image

The user would click this instead of the checkbox/label, just to make things clear (since the primary goal of this UI element is to enable/disable tools). 

However, this still has a few problems. For example, users trying to find tool-specific settings in this screen would probably not think to look in the **Interface** tab. They'd probably check **General** or **Filename Editor** first. Having it be in the **Button Selection** group only makes sense because that's the only existing setting that deals with tools, but it kinda doesn't make sense otherwise.

I think a far better solution is just to have a new tab related to the tools, or a group in **General** for them. Putting a tool-settings group in general might make sense, since there does appear to be some whitespace there:

image

However, I think a new tab is best, since there *are* a few other tools that could use settings (like upload to control different upload destinations, Text to control font, Marker to control opacity, etc.

borgmanJeremy:
Closing this for now. I would like to see this reworked after we add per tool configs.

PR #627, opened 2019-10-02
Improve German translation

simon04:


mmahmoudian:
There is one commit that I'm not sure about it. It fall into @borgmanJeremy area of expertise.

PR #630, opened 2019-10-10
fix text config width in side panel (was too big, adding hor. scroll)

AlexP11223:


AlexP11223:
Not working correctly. Fixed in #875

Issue #613, opened 2019-09-23
toggle button for sidepanel should be on screen


fikrillah:
hi, i would like to purpose a feature please

in tablet you don't have access to keyboard.
currently to show side panel, you have to press space bar which is not available on tablet. so there's no way to have side panel on tablet.

so, to make side panel accessible to tablet user, there should be toggle button for it on screen

fikrillah:
together with #611 and #612 , we can ditch this ugly screen, so users don't have to read those text
![Screenshot from 2019-09-23 18-27-50](https://user-images.githubusercontent.com/48489341/65422177-f19df180-de2f-11e9-8ca2-f2772f10ee9f.png)

AlexP11223:
I submitted PR adding this feature. #631

PR #631, opened 2019-10-11
add sidepanel button to make UI more intuitive and accessible without keyboard (tablets, etc.)

AlexP11223:
addresses #613 by adding the button in the center of the left border of the primary screen that opens the side panel, and also Close button at the bottom of the side panel.

The button can be removed via config option.


mmahmoudian:
I personally think this is a must, although it obscure some part of the the screen. I personally would prefer if user can drag it up and down in cases that this icon is on their way.

AlexP11223:
> I personally think this is a must, although it obscure some part of the the screen. I personally would prefer if user can drag it up and down in cases that this icon is on their way.

Now it is draggable. The position is not saved between captures, the goal is just to be able to move it when it's on the way besides completely disabling it in the settings.

src/widgets/panel/utilitypanel.cpp
```diff
@@ -72,7 +73,7 @@ UtilityPanel::clearToolWidget()
 void
 UtilityPanel::pushWidget(QWidget* w)
 {
-  m_layout->addWidget(w);
+  m_layout->insertWidget(m_layout->count() - 1, w);
```
borgmanJeremy, line 76: If this is being inserted at the end why not use 
```
addWidget()
```

src/widgets/orientablepushbutton.cpp
```diff
@@ -0,0 +1,87 @@
+// Copyright(c) 2017-2019 Alejandro Sirgo Rica & Contributors
```
borgmanJeremy, line 1: Can you help me understand why this is an orientable push button instead of a normal push button? I don't see where the orientation is every changed so it might be better to make it a simple push button.

borgmanJeremy (changes requested):

src/widgets/orientablepushbutton.cpp
AlexP11223: https://github.com/flameshot-org/flameshot/pull/631/files#diff-735729613df2a40b1fc629636c6e7b6bR619

src/widgets/panel/utilitypanel.cpp
AlexP11223: It needs to insert widgets before the Close button now, not at the end.

src/widgets/draggablewidgetmaker.cpp
```diff
@@ -0,0 +1,81 @@
...
+void
+DraggableWidgetMaker::makeDraggable(QWidget* widget)
+{
+  widget->installEventFilter(this);
+}
+
+bool
+DraggableWidgetMaker::eventFilter(QObject* obj, QEvent* event)
+{
+  QWidget* widget = static_cast<QWidget*>(obj);
```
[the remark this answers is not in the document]
AlexP11223: fixed.

borgmanJeremy (approved):

ZetaoYang:
This feature cannot work well at Windows 10.

mmahmoudian:
@AlexP11223 do you have a quick fix for this?

ZetaoYang:
another bug,(not in this PR) is when windows task bar is on the top, it'll be two, see the above screenshot.

borgmanJeremy:
Does the new button work properly and just have wrong colors?

ZetaoYang:
@borgmanJeremy just wrong colors.

PR #648, opened 2019-11-10
correct position of QScreen geometry on HiDPI multi-headed setup

PeterCxy:
Although QT documentation claims that the property `QScreen.geometry` is
pixel instead of dip, this does not seem to be the case at least in my
setup with two 3840x2160 screens on KDE Xorg with a global pixel ratio
of 2.0. `width()` and `height()` return values in dip (1920x1080), while
`x()` and `y()` return values in px (the second screen is indicated to
be at the point (3840, 0)) as of my latest test on QT 5.13.

This weird behavior causes issues when trying to use Flameshot on such
setups because the grabbed image does not match the actual geometry of
the screens. In my case, the grabbed image is 3 times the width of each
of the screen, leaving the rightmost part black, and causing Flameshot
to show a black image on one of them in the editing state.

I am not sure if this strange behavior of QT is consistent for everyone, so
more test in similar setups might be needed. But if it is a behavior of
QT, then a fix from upstream will probably be implausible due to
possible breakages it might cause, so this patch would still be needed anyway.

If anyone has similar setups of two side-by-side HiDPI monitors, please
report if there were issues when using Flameshot, and if so, please test
if this patch fixes the problems properly. I'm not sure if this would
work properly on multiple-monitor setups, either.

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -90,7 +90,10 @@ QPixmap ScreenGrabber::grabEntireDesktop(bool& ok)
 
     QRect geometry;
     for (QScreen* const screen : QGuiApplication::screens()) {
-        geometry = geometry.united(screen->geometry());
+        QRect scrRect = screen->geometry();
+        scrRect.moveTo(scrRect.x() / screen->devicePixelRatio(),
+                       scrRect.y() / screen->devicePixelRatio());
+        geometry = geometry.united(scrRect);
     }
 
     QPixmap p(QApplication::primaryScreen()->grabWindow(
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -126,6 +126,8 @@ CaptureWidget::CaptureWidget(const uint id,
     if (m_context.fullscreen) {
         for (QScreen* const screen : QGuiApplication::screens()) {
             QRect r = screen->geometry();
+            r.moveTo(r.x() / screen->devicePixelRatio(),
+                     r.y() / screen->devicePixelRatio());
 #ifdef Q_OS_WIN
             r.moveTo(r.topLeft() - topLeft);
 #endif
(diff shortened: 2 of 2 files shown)


andrei-pavel:
Same here, a black or white portion, depending on monitor?!, on the rightmost portion of each connected monitor. Hoping for a fix.

```
$ xrandr
Screen 0: minimum 320 x 200, current 4480 x 1440, maximum 16384 x 16384
eDP-1 connected primary 1920x1080+0+360 (normal left inverted right x axis y axis) 344mm x 194mm
HDMI-1 connected 2560x1440+1920+0 (normal left inverted right x axis y axis) 527mm x 296mm
```

printesoi:
@PeterCxy I'm not using flameshot, but I have the same issue with QScreen.geometry on a dual 4K monitor setup (4K laptop on bottom + 4K external monitor on top) on Qt 5.14.2-2, KDE 5.70.0 / Plasma 5.18.5.

    $ xrandr -q | grep -E '\bconnected'
    DP-0 connected 3840x2160+0+0 (normal left inverted right x axis y axis) 609mm x 349mm
    DP-2 connected primary 3840x2160+0+2160 (normal left inverted right x axis y axis) 344mm x 193mm

But QScreen geometry for the bottom monitor is QRect(0,2160 1920x1080)

PeterCxy:
@borgmanJeremy Sorry if I mentioned the wrong person for this, but I'd really appreciate it if someone can take a look at this patch and hopefully merge it into the master branch. Before flameshot resumed development I just locked on one single commit, but now that it has been revived, it has become really inconvenient that flameshot breaks on my multi-headed HiDPI setup every time it updates.

borgmanJeremy:
Sure, I'll get this reviewed soon. Thanks for the contribution!

borgmanJeremy:
@PeterCxy Are you able to test this on windows with multiheaded hidpi?

PeterCxy:
Unfortunately no, I don't have any Windows installation available. I only
know that before this patch flameshot was completely unusable on Linux w/
Xorg with HiDPI multihead setup and it works perfectly with this applied. I
could have tested on Wayland though but sadly KDE Wayland itself does not
even work properly in such a setup.

borgmanJeremy:
Okay the change looks good overall. I'll get it merged now so we get plenty of testing before the 0.9 release. Can you resubmit the PR? There was some unfortunate timing on this merge because we agreed to change our clang-format profile so now there are merge conflicts.

PeterCxy:
@borgmanJeremy I force-pushed to the PR branch and rebased to the latest master branch.

borgmanJeremy (approved):

borgmanJeremy:
Thanks for the contribution!

PR #677, opened 2019-12-17
add notes for compilation on CentOS 8

shikari7:
I compiled from source on CentOS 8 and wanted to share what I learned.

Be gentle on me please - this is my very first PR to a repo on the public GitHub :)


borgmanJeremy:
Hello,

Thanks for the contribution! Could you please update this document for that latest version of flameshot? We have moved to CMake recently so likely these directions will need updating.

PR #686, opened 2020-02-19
imgur: make uploader to be compliant with w3c recommendations

seamus-45:
application/x-www-form-urlencoded is inefficient for binary uploads, instead
we should use multipart/form-data POST requests. Imgur API support this.


borgmanJeremy:
I'm going to close this for now because the entire uploader system is being reworked in 0.9

PR #698, opened 2020-03-16
Fix keyboard shortcuts link in the readme

Fuhrmann:


mmahmoudian:
@borgmanJeremy I checked the commit and it seems the change is correct, previously the link was not pointing at the correct place. I just don't understand why the check failed!

borgmanJeremy:
It's from the old appveyor CI. If it's good to merge we can override.

PR #658, opened 2019-11-24
Unfinished translations and fixes some

ialexsilva:

PR #701, opened 2020-03-18
More accurate Portuguese translation

d4niloArantes:
Nothing big but it kind bothers me.


miguelbrandao:
As a fellow portuguese speaker, I can say this translation is way better

Streppel:
@lupoDharkael <3

mmahmoudian:
As a headsup, this might conflict with #658 as they both are modifying the same file.

borgmanJeremy:
Fixed merge conflicts on my fork. Will merge in a day or so with other translations.

PR #727, opened 2020-04-21
Fix incorrect selection rectangle geometry bug

milnadar:
This commit fixes bug where selection rectangle coordinates and geometry
is set to 0 if is is positioned at the edge of the screen and then resized
to have its width or height to 0

bug


milnadar:
Could anyone from project owners please check the CI issue? Seems like the build fail is not related to my changes?

borgmanJeremy:
This looks good to me. Do you mind rebasing on master? If you don't have time I can get to it in a day or so.

Issue #583, opened 2019-07-20
If hovering over buttons during a screenshot selection area, flameshot's background overlay will flicker in and out


adrianocr:
**Flameshot version**
0.6.0

**Describe the bug**
When the capture overlay/view is open, if you hover over any of the buttons/icons to see their label and then hover over another button, flameshot develops a constant flicker of the screen about 1-2 secs apart. 

Interestingly enough I tried to use Peek to record a gif of it happening, but if Peek is recording flameshot is somehow magically fixed. If I close Peek though, the flicker issue returns. 

**To Reproduce**
Use flameshot to take screenshot of a rectangular area and then hover over the buttons to see their labels. Once a label has appeared, hover over to another button. You will notice that the translucent black overlay that flameshot applies over the rest of the screen flickers in and out every 1-2 second while you're hovering the mouse around. 

**Expected behavior**
No flicker of the black background overlay.

**System Information**
Fedora 30, latest Gnome DE.

Victolabs:
Also having this issue on POP!_OS 19.04.

AlexP11223:
the same issue in master too, POP!_OS 19.04.

Issue #517, opened 2019-05-03
Dimmed area blinks when hover mouse over GUI button


skywtchr:
**Flameshot version**
Flameshot 0.6.0-9(Debian)
Compiled with Qt 5.11.3

**Describe the bug**
After selecting area for screenshot and placing mouse pointer over one of the GUI buttons.
Dimmed area around selection blinks (which means back to normal and dim again).

**To Reproduce**

- Install app, type _flameshot gui_ in console 

- Select area to get screenshot

- Hover on any GUI button and wait until description shows up

**Expected behavior**
Dimmed area around selection stays dimmed, when button description shows up.

**System Information**
Ubuntu 19.04 - Disco Dingo

greno4ka:
Can't reproduce it with flameshot, built from commit 4261915. System ALT Sisyphus.

PR #731, opened 2020-04-26
Disable Qt::BypassWindowManagerHint, workaround #583 #517

filipkilibarda:
Workaround for previously reported screen blinking issue #583 #517 

## System
- Ubuntu 18.04
- Xorg
- Gnome Shell 3.28.4

## The bug
Basically if you hover your mouse over one of the widgets in the screenshot GUI, wait for the tooltip to pop up, then immediately move your mouse off the widget, the whole screen blinks once.

Unfortunately I can't screen record it because it doesn't reproduce if a screen recorder is running. Only option would be to record my screen with my phone.

## What does this PR improve?
- Stops the blinking
- Allows the user to switch workspaces while in the screenshot GUI. Prior to this PR the screenshot GUI would effectively freeze because the window manager would never be able to bring the screenshot GUI back into focus after switching from/to workspaces.

## Coments
I don't know why `Qt::BypassWindowManagerHint` was specified in the first place, so I don't know what kinds of unintentional side effects removing this could have.

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -97,8 +97,9 @@ CaptureWidget::CaptureWidget(const uint id, const QString &savePath,
         }
         move(topLeft);
 #else
-        setWindowFlags(Qt::BypassWindowManagerHint
-                       | Qt::WindowStaysOnTopHint
+        // Disable Qt::BypassWindowManagerHint. Workaround for #583 #517
+        setWindowFlags(/* Qt::BypassWindowManagerHint */
+                       Qt::WindowStaysOnTopHint
                        | Qt::FramelessWindowHint
                        | Qt::Tool);
 #endif


borgmanJeremy:
I want to look into the side effects of removing ```Qt::BypassWindowManagerHint``` Once we document that I'll merge this in.

jfly:
I'd like to see something like this get merged. I played around with it a bit and it has some really nice properties for me as a Xmonad X11 user:

  - It resolves https://github.com/flameshot-org/flameshot/issues/784
  - It also makes it possible for me to clearly see when flameshot has focus, and give it focus, which I'd argue is a workaround/fix for https://github.com/flameshot-org/flameshot/issues/1072, and also has some nice properties as @filipkilibarda mentioned: "Allows the user to switch workspaces while in the screenshot GUI..."

I did some git spelunking and here's the history of the current code:

- c4d9210c35025e328df57a10364c0da5570b1c15 is the first commit where something like this showed up, but it used `Qt::X11BypassWindowManagerHint`, which is just [an alias for `Qt::BypassWindowManagerHint`](https://github.com/qt/qtbase/blob/b75d60abd2012d78387ec0751e205aef970a024b/src/corelib/global/qnamespace.h#L247)
- 0f30529c77476e90ec49cc9abbf5bb6cdc422f6f removed it (yay!)
- 11b0e2db4b4c0aadae9077622e9e2e72bb5d737f added in `Qt::BypassWindowManagerHint` with somewhat cryptic message: "Capture window showing when mouse events are holded" message. I'm not sure what that means.
- Apparently this flag caused issues on Windows and was removed in b1174a1e0501fff9e1d09bdf14028ca230746fe5

So, since flameshot doesn't currently target OSX, I think this change only affects Linux. @borgmanJeremy if I did some investigation into how this behaves with other window managers (and maybe wayland?) would you be open to merging it up?

jfly:
(Btw, not sure if this is the right place to ask, but @borgmanJeremy is there a slack workspace for flameshot you'd be willing to invite @FatBoyXPC and I to?)

borgmanJeremy:
@jfly Absolutely! I should probably link to it in the README somewhere, its a public slack: https://flameshotworkspace.slack.com

I need to actually test this branch. It was submitted before I really took over so I haven't sat down and tested it. Does it basically just allow alt-tab functionality without adding a border?

jfly:
> Does it basically just allow alt-tab functionality without adding a border?

The behavior for me (again, xmonad on x11) with the `Qt::BypassWindowManagerHint` removed is that I end up with a window floating on top of everything on my screen. If I switch workspaces, I see the workspace and can interact with it as normal, no flameshot in the way (I like that!). I don't really have something like alt-tab, so it's hard for me to answer. The flameshot window still appears on top of most everything (notifications still appear on top of it), but I can use my window manager to converts the window to a regular tiling window and see other stuff.

What's still a bit confusing is that if I switch focus to another window, the flameshot window is still on top of it. But that's just how floating windows work in xmonad (floating stuff floats on top of tiled stuff). What's really *really* nice about this change though is that if I interact with the flameshot window again (by clicking on it) it gets focus again and keyboard shortcuts (such as esc) work, which is why I'd say this could be considered a fix for #1072.

Unrelated, the slack link you sent shows me an option to sign in, but I don't see an option to sign *up*?

FatBoyXPC:
Yeah I think we need a special invite link, rather than the link to the workspace. Alternatively, we could be invited via email address, which for me, is my username here with a gmail address (trying to avoid the spam here).

borgmanJeremy:
Here is the invite link: https://join.slack.com/t/flameshotworkspace/shared_invite/zt-in7gdipi-99oWqKeGKK1DRDOUaMSjvA

borgmanJeremy:
@jfly Have you looked at this in more detail to see what happens on other DE's?

jfly:
I have not! Will try to find some time to do that tonight.

jfly:
@borgmanJeremy sorry for the massive delay here. I just installed xfce and tried this out, and as far as I can tell, things work great! During a screenshot, I can alt-tab to other windows and they end up on top of the ongoing screenshot, but I can click back on the screenshot to continue editing the screenshot (strangely, I cannot alt-tab back to the window). Keyboard shortcuts work as expected, and this feels like an improvement in every way, IMO.

borgmanJeremy:
@jfly Ok great thanks for getting back to me. I'll resolve the conflicts and get this merged shortly so I can include it in the 0.9 release candidate.

PR #758, opened 2020-05-30
Fix freeze when changing color

Mantas-2155X:
**Issue:** Changing color in the right click color picker when using a graphics tablet freezes the UI if you hover over a color circle. If you quickly swipe on it, the color changes without a freeze. Not sure why this is happening when a tablet is used, maybe update rate is different.

**Fix:** Make the color update only when the hover selection is different from the current selected color.

### DIFF
--- a/src/widgets/capture/colorpicker.cpp
+++ b/src/widgets/capture/colorpicker.cpp
@@ -92,7 +92,7 @@ void ColorPicker::paintEvent(QPaintEvent *) {
 
 void ColorPicker::mouseMoveEvent(QMouseEvent *e) {
     for (int i = 0; i < m_colorList.size(); ++i) {
-        if (m_colorAreaList.at(i).contains(e->pos())) {
+        if (m_colorAreaList.at(i).contains(e->pos()) && m_drawColor != m_colorList.at(i)) {
             m_drawColor = m_colorList.at(i);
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Can you help me understand why this change fixes the issue? I don't know why the hover selection changing to the same color causes a crash.

Mantas-2155X:
Honestly, I wouldn't know what exactly causes it. 
All I can assume is that either`emit colorSelected(m_drawColor);` (line 97) or `update();` (line 98)
being ran very often can cause it to freeze. 

The graphics tablet has a lot of move events so it could be the reason because if I swipe over the selection very quickly, the freeze is either very minimal or doesn't exist at all. Hovering on a selection for longer than ~0.2s can cause it to lock up for quite a bit.

My PR makes it only update once when the color actually needs to be changed. This is more efficient in general, not only fixes the bug.

Mantas-2155X:
I went ahead and built the latest commits. Seems like the issue isn't there anymore. Could be anything from me being on the latest kernel or flameshot having some changes. Not sure if you need to merge this anymore, I guess only for efficiency (and maybe if anyone on older systems may experience the bug)

borgmanJeremy:
Ok ill close it for now but if it happens again let me know and we will look into it.

PR #762, opened 2020-06-05
Create a Hot Key to run Flameshot.

mailsjb:
I'm using Flameshot too well. Thank you very much. However, there is one function that is absolutely necessary. It needs a Hot Key to run Flameshot. It is inconvenient to click the Flameshot icon on the taskbar with your mouse every time. And sometimes you need screenshots right from the pop-up screen, but they disappear the moment you go to press the FlameShot icon. Please make one Hot Key setting that can be implemented.


mailsjb:
sorry, It's not right. Go to the issue.

PR #779, opened 2020-06-23
draw rounded rectangle using thickness as radius

EricBlanquer:
draw rounded rectangle using thickness radius  instead of drawing simple rectangle

image

image

### DIFF
--- a/src/tools/rectangle/rectangletool.cpp
+++ b/src/tools/rectangle/rectangletool.cpp
@@ -52,7 +52,7 @@ void RectangleTool::process(QPainter &painter, const QPixmap &pixmap, bool recor
     }
     painter.setPen(QPen(m_color, m_thickness));
     painter.setBrush(QBrush(m_color));
-    painter.drawRect(QRect(m_points.first, m_points.second));
+    painter.drawRoundedRect(QRect(m_points.first, m_points.second), m_thickness, m_thickness);
 }
 
 void RectangleTool::paintMousePreview(QPainter &painter, const CaptureContext &context) {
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
@mmahmoudian I think this change is small enough (very minor behavior change) to merge without an RFC. If you agree I will merge it.

mmahmoudian:
@borgmanJeremy I have nothing against it. But just based on the criteria you defined yesterday on Slack, it is a change to the behavior nonetheless. I don't mind if you merge it without RFC, but then again we need to know how much of behavioral change will be"major" to need RFC.  Perhaps this is a good example to open the discussion on Slack as we have two PRs with behavioral changes as examples and this can assist us to quickly come up with a criteria.

borgmanJeremy:
Makes sense! Let's do it for 0.9.

MyriaCore:
Hey guys! I'd just like to add that if we go with this, I really insist it be user-configurable. 

There are plenty of people who already *have* rounded corners, since it's arguably the compositor's job. For example, picom has a [fork for rounded corners](https://github.com/yshui/picom/pull/229). I use a fork like this to round the corners of flameshot windows whose titles start with a capital letter (as the background's title is just `"flameshot"`). 

Users like me will see duplication of functionality if we aren't given control over this in settings. It can actually mess things quite a bit if the app tries to round its corners more than the compositor, since it'd be basically clobbering a pre-existing user setting. A setting would allow users to not round the corners in-app, and yield control of this feature to the compositor instead.

borgmanJeremy:
This does not round the window corners. It allows users to round the corner of the user drawn rectangles.

PR #780, opened 2020-06-23
add more colors

EricBlanquer:
just add more colors

tmp

### DIFF
--- a/src/utils/confighandler.cpp
+++ b/src/utils/confighandler.cpp
@@ -77,15 +77,21 @@ void ConfigHandler::setButtons(const QVector<CaptureButton::ButtonType> &buttons
 QVector<QColor> ConfigHandler::getUserColors() {
     QVector<QColor> colors;
     const QVector<QColor> &defaultColors = {
-        Qt::darkRed,
+        Qt::white,
         Qt::red,
-        Qt::yellow,
         Qt::green,
-        Qt::darkGreen,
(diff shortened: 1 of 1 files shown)


willydee:
This patch might not be necessary, since user defined colors can be placed in `flameshot.ini`:
```
userColors=#ffffff, #888888, #000000, #800000, #ff0000, #ffff00, #00ff00, #008000, #00ffff, #0000ff, #ff00ff, #800080
```

The first three are extra; the others are the built-in colors.

EricBlanquer:
I was not aware of that possibility to edit this file to add this extra configuration, and what kind of users know that?
so, I think at least : white, black and gray should be added by default

willydee:
> what kind of users know that?
> so, I think at least : white, black and gray should be added by default
True. But considering that the repo seems to be sort of unmaintained for a while, and your pull request might not be merged anytime soon, I just wanted to point out a workaround that was mentioned elsewhere.

EricBlanquer:
oh, OK, thank you for that

Le jeu. 16 juil. 2020 à 09:43, Willy Schott <notifications@github.com> a
écrit :

borgmanJeremy:
Additionally the new side panel lets you select any color.

Robou:
I totally vote for this feature, so that users can easily create their own color palette.
I think that at least the quick access color ring must contain black and white additionally to the already present colors.

PR #781, opened 2020-06-25
Changing behaviour of Enter key to save screenshot 

itSumanth:
Changing the behaviour of Enter key to save screenshot from gui to a specified path folder. It is more convenient to save it to the folder rather than clipboard ass it becomes annoying to look for the image from clipboard.


leogx9r:
This definitely should be a configurable option, rather than a default as I do use the clipboard-copy option by default, which means this would break for me.

StyleT:
Hello from Namecheap R&D team! We're (with @thepurple) actually planning soon to contribute ability to have behaviour of Enter key configurable via UI/config. However if it will be done here - this would be great.

joecharamut:
I think my PR #516 does what you are looking for, adding an option to save to clipboard and a folder when you press enter

borgmanJeremy:
Closing this as we are working with @thepurple on merging this part of the namecheap fork.

PR #819, opened 2020-09-04
Fix CMake Package and DBUS Names

borgmanJeremy:


hosiet (changes requested):
From the code changes I believe everything should be OK.

Just one point: can we cherry-pick ae803b95c4dbcfbea2f22b1f5b589b031c128951 on top of current git HEAD to avoid unnecessary messy git history? There are duplicated commits.

图片

Other than this one I think everything should be ok.

borgmanJeremy:
Yeah good catch, Ill do a squash merge.

PR #830, opened 2020-09-05
Update cmake qt5 translation

ZetaoYang:
Should use `qt5_add_translation` instead of `qt5_create_translation`

- https://doc.qt.io/qt-5/qtlinguist-cmake-qt5-create-translation.html
- https://doc.qt.io/qt-5/qtlinguist-cmake-qt5-add-translation.html
- here is an another example: https://github.com/pbek/QOwnNotes/blob/develop/src/CMakeLists.txt

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -25,9 +25,7 @@ add_subdirectory(utils)
 add_subdirectory(widgets)
 add_subdirectory(tools)
 
-qt5_create_translation(
-  QM_FILES
-  ${CMAKE_SOURCE_DIR}
+set(FLAMESHOT_TS_FILES
   ${CMAKE_SOURCE_DIR}/data/translations/Internationalization_ca.ts
   ${CMAKE_SOURCE_DIR}/data/translations/Internationalization_de_DE.ts
   ${CMAKE_SOURCE_DIR}/data/translations/Internationalization_fr.ts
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
The reason I used create translation was so the. ts files would update when the source lines changed.

Perhaps we can add a cmake option that is off by default that can be used to update the ts files?

ZetaoYang:
When using `qt5_create_translation`, I got an error,
image

borgmanJeremy:
Hrmm okay it works with gcc and msvc compiler. Maybe it's a minggw issue?

borgmanJeremy:
I think moving the create translation behind a default off build option is best anyway. Then the ts files won't be updated on every commit.

ZetaoYang:
I even tryed `qt5_create_translation` under MSVC 2019 , but got another error: `src\CMakeFiles\flameshot.dir\build.make(241) fatal error U1036: syntax error : too many names to left of '='`. When I switched to `qt5_add_translation`, it can be builded successfully.

ZetaoYang:
Maybe it's a bug from mingw64 provided by TDM-GCC distribution.

borgmanJeremy:
Can you see if #836 fixes your issue? I think this provides the best of both worlds so by default the .ts files are not modified but we have a way to generate them before release.

Issue #824, opened 2020-09-05
Could not find SHOpenWithDialog under Windows 10 using mingw64(TGM-GCC)


ZetaoYang:
After flameshot building mechanism refactoring, I try run new code under Windows 10 using JetBrains Clion with mingw64(TDM-GCC 9.2.0). found in the file `src/tools/launcher/openwithprogram.cpp`, an error appears: `error: 'SHOpenWithDialog' was not declared in this scope`, as shown in the screenshot, 
![image](https://user-images.githubusercontent.com/10769951/92306566-a50ce100-efc2-11ea-90ce-a7f788409be6.png)

In this header file `shlobj.h` line 2113 need `NTDDI_VERSION >= 0x06000000`, 
 
![image](https://user-images.githubusercontent.com/10769951/92306671-865b1a00-efc3-11ea-81b8-d87209addddf.png)

that means SHOpenWithDialog requires Windows Vista or higher, see https://docs.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers, mentioned in the document,
> If you define NTDDI_VERSION, you must also define _WIN32_WINNT.

This `windows.h` header file introduces these two headers `_mingw.h` and `sdkddkver.h`. In `_mingw.h` it define _WIN32_WINNT as 0x502 (namely Windows Server 2003), from line 233,   
![image](https://user-images.githubusercontent.com/10769951/92306825-b0610c00-efc4-11ea-8532-ec4aa8eb4f30.png)

In `sdkddkver.h`, it will choose NTDDI Version via _WIN32_WINNT, from line 164,
![image](https://user-images.githubusercontent.com/10769951/92306855-edc59980-efc4-11ea-90a0-d0f97cb5e732.png)

My workaroud is add below code after line `#include <windows.h>`:
```
#ifdef _WIN32_WINNT
#undef _WIN32_WINNT
#define _WIN32_WINNT _WIN32_WINNT_VISTA
#endif
```
![image](https://user-images.githubusercontent.com/10769951/92306872-0e8def00-efc5-11ea-90e1-670fbbca8aec.png)

Or
remove `#include <windows.h>`, and add below code before `#include <Shlobj.h>`,
```
#define NTDDI_VERSION 0x06000000
#define _WIN32_WINNT 0x0600
```

hosiet:
I just tried to build using a slightly different toolchain and the build was successful:

* System: Windows 10 v19041
* Mingw-w64 provided by MSYS2
[...]

PR #837, opened 2020-09-06
Enable WIN32 property on Windows build

hosiet:
When building on Windows platform, this patch will set the `WIN32_EXECUTABLE` property for the build target. This helps to suppress the terminal window when running on Windows.

Details on this cmake property can be found at https://cmake.org/cmake/help/v3.13/prop_tgt/WIN32_EXECUTABLE.html .

This solves some problems discussed in issue #824.


ZetaoYang:
A completely resolved PR https://github.com/flameshot-org/flameshot/pull/849.

PR #851, opened 2020-09-07
Deb & rpm packing via Github Actions

ZetaoYang:
- update `flameshot.spec` for fedora. This spec file do not apply to openSUSE, because some packages name is different between fedora and openSUSE.
- add deb & rpm packing via Github Actions based on docker(different distributions to build themselves packages). Images from https://hub.docker.com/repository/docker/vitzy/flameshot, it was built from https://github.com/flameshot-org/flameshot-docker-images.
- I add some file temporary storage services into `<project_root-folder>/scripts/upload_services`, they are `0x0.st, file.io, transfer.sh, wetransfer.com`, so we can upload/download daily build packages.

### DIFF
--- a/.github/workflows/pack.yml
+++ b/.github/workflows/pack.yml
@@ -1,0 +1,87 @@
+name: Packing
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  PRODUCT: flameshot
+  VERSION: 0.8.0
+  RELEASE: 1
+  ARCH: x86_64
+  # dockerfiles, see https://github.com/flameshot-org/flameshot-docker-images
+  # docker images, see https://hub.docker.com/r/vitzy/flameshot
+  DOCKER_REPO: vitzy/flameshot
+  # upload services: 0x0.st, file.io, transfer.sh, wetransfer.com
+  UPLOAD_SERVICE: file.io
+
+jobs:
+  deb-pack:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        dist: [debian-10, ubuntu-20.04]
+    steps:
+      - name: Checkout Source code
+        uses: actions/checkout@v2
+      - name: Get packpack tool
+        uses: actions/checkout@v2
+        with:
+          repository: flameshot-org/packpack
+          path: tools
+      - name: Pack on ${{ matrix.dist }}
+        if: matrix.dist == 'debian-10'
+        run: |
+          cp -r $GITHUB_WORKSPACE/data/debian $GITHUB_WORKSPACE
+          bash $GITHUB_WORKSPACE/tools/packpack
+        env:
+          OS: debian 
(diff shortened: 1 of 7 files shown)


borgmanJeremy:
@hosiet I think you mentioned having the folder structure as /data/deb and /data/rpm would cause an issue. With the fixes @ZetaoYang did here is it okay to leave the packages in /data?

hosiet:
@borgmanJeremy The script actually explicitly copy the /data/debian dir back to be /debian . In this sense there are no problems.

ZetaoYang:
@borgmanJeremy It doesn't work. It must be placed in project root folder.So I copy them `cp -r $GITHUB_WORKSPACE/data/debian $GITHUB_WORKSPACE` `cp -r $GITHUB_WORKSPACE/data/rpm $GITHUB_WORKSPACE`.

borgmanJeremy:
Ohh ok perfect.

borgmanJeremy (approved):

Creteil:
So, to be clear, if I want to build debian package from the git repository, how I have to do ?

I was able to compile until the cmake 3.13 dependency (I am running 18.04 LTS with cmake 3.10...).

borgmanJeremy:
I recommend you install the latest version of cmake using the shell script provided by kitware: https://cmake.org/download/

They also provide a snap if you prefer that.

Creteil:
Since I like to keep my distribution debian oriented, here how to build for 18.04 :

Update cmake using « [Kitware APT Repository](https://apt.kitware.com/) ».
Then fetch code from git and build package like that :

```
mkdir -p ~/tmp/Building/flameshot
cd ~/tmp/Building/flameshot/
git clone https://github.com/flameshot-org/flameshot.git
cd flameshot/
cp -R data/debian .
echo 11 > debian/compat
DEBEMAIL="COLIN Stephane <bigbob.fun@gmail.com>" dch -b -v $(git --git-dir .git --work-tree . describe --always --tags | cut -c2-) "Update to latest GIT Version..." && sed -i '$!N; /^\(.*\)\n\1$/!P; D' debian/changelog ; head -10 debian/changelog
dpkg-checkbuilddeps && rm debian/*~ ; LANG=C dpkg-buildpackage -rfakeroot
cd ..
```
You must find your debian packages here right now.

Hope it help...

Creteil:
hmmm, even the package build fine, the version reported is « 0.8.0 » :

image

Why you pushed versioning up to 0.8.0 ?

Also tried to change the version in « CMakeLists.txt » with same as the one generated by :

`echo "$(git --git-dir .git --work-tree . describe --always --tags | cut -c2-)"`

doesn't work, compilation end with error :

```
CMake Error at CMakeLists.txt:4 (project):
  VERSION "0.6.0-126-g60ef3f3" format invalid.
```
It's regression in terms of package naming, because with the old format, the package generated have all information inside naming (0.6.0-126-g**60ef3f3**)...

This is just my feeling...

Issue #850, opened 2020-09-07
flameshot gui does not launch app as before


zorn-v:
**Flameshot version**
Flameshot v0.8.0(git20200904,Debian)
Compiled with Qt 5.14.2

**Describe the bug**
Previously if you run `flameshot gui` with not prelaunched flameshot it launch app and then allow to take screenshot.
I set hotkey for "take screenshot" which run this command as I supopse and all works fine. 
But some time ago flameshot was updated in Kubuntu 20.10 and if app not launched nothing happens.

**To Reproduce**
Close flameshot and run `flameshot gui` or `flameshot config` or `flameshot launcher` etc. Nothing happens.

**Expected behavior**
flameshot app should be launched if needed

**System Information**
Kubuntu 20.10 (Groovy Gorilla) dev branch

zorn-v:
Just build with https://github.com/flameshot-org/flameshot/pull/852 merged - problem not fixed.

PR #852, opened 2020-09-07
Fix D-Bus name and organization name

hosiet:
Originally the App uses org.Dharkael.flameshot in D-Bus.
Since we have moved the project under organization, the name we are using now should be org.flameshot.flameshot.

This need to be kept unified across the whole project.

This fixes D-Bus invocation problem and should closes: #850 .

* * *

This patch also changes the organization name. This means the program will no longer store config files under `~/.config/Dharkael/flameshot`. It will now use `~/.config/flameshot/flameshot/`. Please let me know if such switch is favorable. If it is too abrupt, we can reverse this change for now.

### DIFF
--- a/data/debian/copyright
+++ b/data/debian/copyright
@@ -1,6 +1,6 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: flameshot
-Source: https://github.com/lupoDharkael/flameshot/
+Source: https://github.com/flameshot-org/flameshot/
 
 Files: *
 Copyright: 2016-2019 lupoDharkael <izhe@hotmail.es>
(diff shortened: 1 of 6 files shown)


borgmanJeremy:
@hosiet There are merge conflicts after the changes made in #851

hosiet:
@borgmanJeremy I have fixed the merge conflict.

@ZetaoYang I saw that you removed the installation of AppData XML when updating the .spec file. Was that intentional or a typo? I believe this file should also be installed in the packaging.

ZetaoYang:
I removed appdata xml because of cannot find this file when rpm-build make install. If GitHub actions can pass on your PR, please add that file back.

ZetaoYang:
BTW, I found our src/cmakelists.txt also doesn't install that file.

hosiet:
Then I think it is the bug in CMakeLists.txt. I will open another bug report for that.

borgmanJeremy:
Agreed.

PR #855, opened 2020-09-07
prototype of undo fix

borgmanJeremy:
@hosiet 

This is a first draft of fixing the issue where the counter numbers do not decrement after undoing. I saw that the nameID() function was never used anywhere in the code base, so in the Undo function I call a decrement function if the undo tool is being applied to the counter bubble tool.

If this is generally reasonable, I want to go through the code base and assign a NameID to all the child classes. Additionally I want to change it to a enum from a QString to make comparison more efficient. 

Before I do all this work I wanted to make sure this implementation seemed reasonable. I think anything else would require a larger refactor. 

pic

### DIFF
--- a/src/tools/abstractactiontool.h
+++ b/src/tools/abstractactiontool.h
@@ -36,6 +36,9 @@ public:
   void paintMousePreview(QPainter& painter,
                          const CaptureContext& context) override;
 
+protected:
+  virtual ToolType nameID() const = 0;
+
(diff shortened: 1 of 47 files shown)


hosiet:
The changes look overall good to me but I'm not sure if nameID() was initially intended to be used like this. Maybe @lupoDharkael may provide some more idea. Personally I am okay with it.

PR #862, opened 2020-09-08
Fixed Plasma 5.20 issue

### LINKED ISSUE #766: Fix upcoming compatiblity issue with Plasma 5.20
Kwin screenshot effect with https://phabricator.kde.org/D29407 will require to add `X-KDE-DBUS-Restricted-Interfaces=org.kde.kwin.Screenshot
` to its .desktop file to stay compatible with Plasma 5.20
In turn it will allow to take screenshot without confirmation click (i.e flameshot will be able to take immediate screenshot like spectacle https://phabricator.kde.org/D29408)
In https://github.com/lupoDharkael/flameshot/blob/91ba28c3ded7ca3314f02418761c982069cf8b33/docs/desktopEntry/package/flameshot.desktop#L30

Also FYI the screenshot effect can now take non-downscaled fullscreen screenshots in Wayland since Plasma 5.19 https://phabricator.kde.org/D29122 using `screenshotFullscreen(QDBusUnixFileDescriptor fd, bool captureCursor = false, bool shouldReturnNativeSize = false);`

borgmanJeremy:
Closes #766

### DIFF
--- a/docs/desktopEntry/package/flameshot.desktop
+++ b/docs/desktopEntry/package/flameshot.desktop
@@ -28,6 +28,11 @@ Type=Application
 Categories=Graphics;
 StartupNotify=false
 Actions=Configure;Capture;Launcher;
+X-DBUS-StartupType=Unique
+X-DBUS-ServiceName=org.flameshot.Flameshot
+X-KDE-DBUS-Restricted-Interfaces=org_kde_kwin_effect-screenshot
+
+
 
 [Desktop Action Configure]
 Name=Configure


hosiet:
I haven't been using KDE for a while so I cannot comment whether this is correct or wrong. However, adding them should be at least harmless.

I'm not sure if we are to implement similar approach in parsing Plasma's version as in https://phabricator.kde.org/D29408 or we have ignored this part for now.

borgmanJeremy:
I'm not sure that will be needed for Flameshot since we always capture in immediate mode. I'll keep an eye out for the KDE 5.20 beta and hopefully get flameshot tested in it before it officially releases.

meven:
There is another expectation, I failed to mention before, the Exec field should contain the absolute path the executable file.
I am thinking about lifting this restriction though.

Yes, the patch is mostly fine.
The "X-DBUS-*" was not necessary, btw.

The fact was that before flameshot would always end up with a click confirmation (that's handled on kwin side) on Wayland.

Issue #867, opened 2020-09-08
Another bug with counter circle undo/redo


hosiet:
**Flameshot version**
git HEAD

**Describe the bug**
I think there are still bugs with counter circle undo/redo.

**Reproduce the bug**
* Invoke counter circle 1, 2, 3, 4, 5
* Undo
* Undo
* Redo
* Redo
* Invoke counter circle again. It is supposed to be 6 but actually gives 4.

borgmanJeremy:
Good catch. I need to have redo emit the increments.

borgmanJeremy:
Thanks for all the testing on this new feature. Hopefully this is the last bug :)

The fix is not as elegant as I want it to be, but the UndoStack as provided by Qt uses strings instead of any other type of ID.

PR #868, opened 2020-09-09
Fix count when the redo tool is called with circle counter

borgmanJeremy:
Checked if the tool about to be called in the "redo" method is the circle counter. If it is, need to increment the count to make up for subtraction during undo.


borgmanJeremy:
Closes #867

PR #878, opened 2020-09-10
Merge Conflicts for #516

borgmanJeremy:
This merges PR #516  and resolves merge conflicts.

### DIFF
--- a/src/config/geneneralconf.cpp
+++ b/src/config/geneneralconf.cpp
@@ -18,16 +18,18 @@
 #include "geneneralconf.h"
 #include "src/core/controller.h"
 #include "src/utils/confighandler.h"
+#include "src/utils/filenamehandler.h"
 #include <QCheckBox>
 #include <QFile>
 #include <QFileDialog>
 #include <QGroupBox>
 #include <QHBoxLayout>
+#include <QLineEdit>
 #include <QMessageBox>
 #include <QPushButton>
+#include <QStandardPaths>
 #include <QTextCodec>
 #include <QVBoxLayout>
-
 GeneneralConf::GeneneralConf(QWidget* parent)
   : QWidget(parent)
 {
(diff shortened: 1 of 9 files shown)


borgmanJeremy:
@hosiet This fixes the merge conflicts of #516  I can't remember if we wanted to merge this before 0.8 or not. If we do, please give it one last look. I did some local testing but it could use a second set of eyes.

hosiet:
I didn't find major problems. There is only one minor part that might be improved:

In `src/utils/screenshotsaver.cpp`, it's possible to send out "imaged saved to clipboard" notification before copying the image into clipboard. Logically the notification should only be sent out after the clipboard action has finished. I believe this issue is not related to this Pull Request (since the original behavior is the same) but this can be improved.

joecharamut:
This fix causes a double notification as saveToFilesystem() also sends out its own.

borgmanJeremy:
Do we not want the double notification? It is notifying the user of 2 different things, saving to disk and copying to clipboard.

joecharamut:
In the other PR someone reported it was freezing flameshot for a long time before saving the image, so maybe we add a special message for this?

joecharamut:
[This comment](https://github.com/flameshot-org/flameshot/pull/516#issuecomment-582666125)

borgmanJeremy:
Hrm thats wierd. But yeah that makes sense too. Lets change it so if it does both the user only gets 1 notification that both actions happened. It doesn't help with what @hosiet pointed out (ie the notification may still happen before the action), but at least all code paths get 1 and only 1 notification.

joecharamut:
Sounds good

borgmanJeremy:
Okay I think this makes it so the message is accurate and only emitted once. I elected to use a "string prefix" rather than a flag to the save function to make it stateless and more flexible. Overall the notifications could use some refactoring to make it its own clean subsystem but no rush to do that.

joecharamut:
Looks like it works on my end

PR #890, opened 2020-09-11
[DRAFT] testing windows ci

borgmanJeremy:


borgmanJeremy:
@ZetaoYang Can you take a look at this for preliminary windows CI? It still needs to upload the binary but this lets us do everything on github actions without appveyor.

borgmanJeremy:
I think we can use CPack to package it, but don't have a windows machine to test on .

ZetaoYang:
There are many options to build installers for Windows platforms, such as Inno Setup, Qt installation framework.

ZetaoYang:
@borgmanJeremy  The reason for the packaging failure is that the network request timed out when pulling the docker image.

Issue #705, opened 2020-03-24
Windows 10?


ghost:
Hello, I was wondering if the program will be released for OS Windows 10 ....

Mackzwellz:
there are available builds, check here for links and instructions https://github.com/lupoDharkael/flameshot/issues/29

mmahmoudian:
It is clear that we do not have good documentation about windows builds. Flameshot works on windows, we just need to add that to readme and assets.

mmahmoudian:
I asked @szszoke to kindly help with this issue and he kindly accepted my request. Unfortunately, I cannot assign this issue to him and as the result I have to use this comment as a form of assignment. I highly appreciate the help 👍🏼

mmahmoudian:
@ghost We now have an automated process ( #898) to make the following Windows versions thanks to @ZetaoYang :
- portable
- windows installer

PR #898, opened 2020-09-13
Windows packaging & merge two workflow yaml files

ZetaoYang:
- add windows packaging: provide two ways to install windows package, one is `portable zip` and the other is `.msi installer`.
- merge the previous `cmake-windows.yml` into `build_cmake.yml`
- for all matrix jobs, add `fail-fast: false`

some userful links:
- https://github.com/minetest/minetest/blob/master/.github/workflows/build.yml
- https://github.com/skypjack/gh-greets-qt/blob/master/.github/workflows/build.yml
- https://cristianadam.eu/20191222/using-github-actions-with-c-plus-plus-and-cmake/ 

--- 
  There two installer types: `.exe` and `.msi` ,
  For `.msi` installers:
  - [Windows Installer XML Toolset(WiX)](https://wixtoolset.org/releases/)

  For .`exe` based installers
  - [Nullsoft Scriptable Install System (NSIS)](https://nsis.sourceforge.io/Download)
  Wix is more modern than NSIS. I prefer to use the `.msi` installer.


.github/workflows/Windows-pack.yml
```diff
@@ -0,0 +1,131 @@
...
+      - 'LICENSE'
+  pull_request:
+    branches: [ master ]
+    paths-ignore:
+      - 'README.md'
+      - 'LICENSE'
+
+env:
+  PRODUCT: flameshot
+  VERSION: 0.8.0
```
mmahmoudian, line 17: Shouldn't this be dynamic (e.g read from "${{ github.workspace }}/CMakeLists.txt")?

borgmanJeremy (approved):

mmahmoudian:
#705 was related. Tagging it here for historical reasons.

PR #902, opened 2020-09-13
Fix side panel button on Windows

AlexP11223:
Fixes [the issue on Windows](https://github.com/flameshot-org/flameshot/pull/631#issuecomment-691658871) by reusing `CaptureButton` CSS instead of using palette.

For now I simply extracted all styling stuff into a separate base class, and created new `CaptureToolButton` class for tool buttons.
Probably should be refactored further.

Now the side panel button also has the same hover style.

### DIFF
--- a/src/config/buttonlistview.cpp
+++ b/src/config/buttonlistview.cpp
@@ -36,9 +36,9 @@ void
 ButtonListView::initButtonList()
 {
   ToolFactory factory;
-  auto listTypes = CaptureButton::getIterableButtonTypes();
+  auto listTypes = CaptureToolButton::getIterableButtonTypes();
 
-  for (const CaptureButton::ButtonType t : listTypes) {
+  for (const CaptureToolButton::ButtonType t : listTypes) {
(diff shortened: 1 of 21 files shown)


borgmanJeremy:
Why did you refactor the name to CaptureToolButton? Since this is coming in right before we lock down can you just make the changes needed to fix the side panel on windows?

AlexP11223:
> Why did you refactor the name to CaptureToolButton? Since this is coming in right before we lock down can you just make the changes needed to fix the side panel on windows?

It's just to extract a separate base class, without it we would need to duplicate [all this code](https://github.com/flameshot-org/flameshot/blob/44956094103b6a0e20fc4e37559d3004d6524e09/src/widgets/capture/capturebutton.cpp).

borgmanJeremy:
Ohhhh, ok thanks for explaining! Will review and merge soon. I think this will be the last merge before we release :)

borgmanJeremy (approved):

PR #904, opened 2020-09-13
data/debian: Prepare for 0.8.0 release

hosiet:
Files under data/debian/ directory are updated to be prepared for flameshot 0.8.0 release.

### DIFF
--- a/data/debian/changelog
+++ b/data/debian/changelog
@@ -1,5 +1,5 @@
-flameshot (0.8.0~dev0) unstable; urgency=medium
+flameshot (0.8.0) unstable; urgency=medium
 
   * Placeholder deb package.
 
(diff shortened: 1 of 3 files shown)


ZetaoYang:
@hosiet I recommend `hicolor-icon-theme` as dependency, it's the default fallback theme used by implementations of the Freedesktop.org Icon Theme specification.

hosiet:
@ZetaoYang  I assumed that anyone with a proper desktop environment installed should have hicolor-icon-theme installed, especially since GTK2 and GTK3 in deb-based distributions depend on it already. But since flameshot is qt-based, adding such dependency should be harmless. Now updated.

borgmanJeremy:
@hosiet Is it safe to merge this now or should it wait until 0.8 is ready to drop?

hosiet:
It should be safe; I don't expect issues with early merge.

PR #910, opened 2020-09-14
fix: flatpak enhancement

ZetaoYang:
* add network support
* add `--device=dri` for wayland
* add notification access
* add system tray icon

---
Test on Manjaro Linux 2020.1 Gnome Wayland, but there is another issue, "**Unable to capture screen**"

Screenshot1

Screenshot2

Screenshot3


borgmanJeremy:
@ZetaoYang Thanks, this seems to resolve the issue on KDE and xorg so I will merge it but keep the bug open to track the Gnome issue. I'll get a gnome / wayland VM setup later to do some testing on it.

PR #927, opened 2020-09-16
Fix base snap performance

borgmanJeremy:


borgmanJeremy:
This fixes the snap so it launches. A couple features need to still be resolved:

- Saving files, currently the snap doesnt not have permission to write to the disk outside the snap squashfs
- I don't think "open with external app" will ever work inside snap due to the confinement

ZetaoYang:
To your first point, I think we can explicitly grant access to `removable-media`, then snap package is allowed to access `/media`. If we want to give a snap package full access to our local system (e.g. /mnt) then we can use `--devmode` when installing the application to [disable confinement](https://snapcraft.io/docs/snap-confinement) entirely.

`removable-media` allows the user to connect the "removable-media" interface and grant software access to the files located in their removable media. This need ask for it to be autoconnected on the [snapcraft forum](https://forum.snapcraft.io/c/store-requests) when we published flameshot on snapcraft store.

---
Some useful links you can follow,
- https://github.com/nextcloud/nextcloud-snap/issues/175
- https://github.com/ksnip/ksnip/pull/267
- https://github.com/ksnip/ksnip/blob/master/snap/snapcraft.yaml
- https://ubuntu.com/blog/snapcraft-secret-sauce-kde-neon-extension
- https://forum.snapcraft.io/t/firefox-removable-media-interface-auto-connect-request/5929

borgmanJeremy:
Yes I believe I can get the file saving working, I just think the "open with external app" may be difficult to do. I'll keep working on it now that my OS is fixed. Never update arch during a crunch :)

borgmanJeremy:
@ZetaoYang Do you mind testing this latest version? I cleaned it up a bit and it seems to work well for everything except external applications. Specifically I want to know if it works out of the box or if you get a FontConfig error. I am not sure if this issue is my local machine or a packing issue. If you get a fontconfig error you may run:

```
snap --shell flameshot # This enters you into the snap container
fc-cache -f -r
exit 
```

ZetaoYang:
@borgmanJeremy see https://github.com/flameshot-org/flameshot/issues/933

Issue #913, opened 2020-09-14
[0.8 Beta] Open Launcher allow only one instance


kpcee:
**Flameshot version**
0.8 Beta testing, compiled from source

**Describe the bug**
Tray Icon -> Open Launcher, repeat this several times.
Each time a new instance of the window is created.

**Expected behavior**
If an instance already exists, it should be brought to the foreground but no new instance should be created.

**System Information**
Arch Linux, Xfce, Xorg

mmahmoudian:
Confirmed!

![image](https://user-images.githubusercontent.com/390889/93092317-f5153300-f6a7-11ea-977d-e3658c3059e3.png)

mmahmoudian:
I confirm that this is solved in #931

Issue #912, opened 2020-09-14
[0.8 Beta] Open Launcher, take Full Screen does not work


kpcee:
**Flameshot version**
0.8 Beta testing, compiled from source

**Describe the bug**
Tray Icon -> Open Launcher -> Full Screen -> Take new screenshoot
The preview image is updated, otherwise nothing happens.

**Expected behavior**
After the screenshot has been taken the image should be ready for further processing or at least a signal should sound that the image has been saved automatically. Currently it is not clear what happens, the image is not saved or I didn't found it and the editing tools are not shown.

**System Information**
Arch Linux, Xfce, Xorg

mmahmoudian:
Confirmed! after pressing "Take new screenshot" nothing happens. The Launcher window closes and re-opens, but nothing is saved in Clipboard or in the following locations that I thought might be saved to:
- `~/Pictures`
- `~/`
- `~/Desktop`

### Setup Info
- `Flameshot-0.8.0-x86_64.AppImage`
- OS: Ubuntu 18.04
- DE: Budgie
- Two monitors both connected to the same graphics card

borgmanJeremy:
Good catch! I did some git bisecting and it turns out this feature never worked and was never finished.

PR #931, opened 2020-09-18
Save screenshot from launcher 

borgmanJeremy:
@mmahmoudian Can you test this build for launcher functionality? 

I adjusted it so if you take a full-screen screenshot you are prompted for a path to save the file.

### DIFF
--- a/src/core/controller.cpp
+++ b/src/core/controller.cpp
@@ -188,8 +188,10 @@ Controller::openInfoWindow()
 void
 Controller::openLauncherWindow()
 {
-  CaptureLauncher* w = new CaptureLauncher();
-  w->show();
+  if (!m_launcherWindow) {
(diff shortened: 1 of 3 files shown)


borgmanJeremy:
Closes #912

borgmanJeremy:
Closes #913

mmahmoudian:
@borgmanJeremy I get the following when I run from terminal:

```
Run experimental self-contained bundle
Qt: Session management error: None of the authentication protocols specified are supported
Fontconfig warning: "/etc/fonts/conf.avail/53-monospace-lcd-filter.conf", line 10: Having multiple values in <test> isn't supported and may not work as expected
```


The launcher works perfectly fine (does not open multiple launcher windows). Also, the delay works very nice. **But** taking full screenshot did not give me any option other save saving to clipboard.

PR #934, opened 2020-09-18
Fixed Missing dependency, Fixed spelling mistake

zaidrpm:
qttools5-dev is also required on ubuntu 20.04 (tested on pop os 20.04). So added it under dependency (debian) in readme.md
Fixed a spelling mistake in comment in src/widgets/capture/capturewidget.cpp

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,7 +341,7 @@ Info: If I take code from your project and that implies a relicense to GPLv3, yo
 
 ## Contribute
 
-If you want to contribute check the [CONTRIBUTING.md](CONTRIBUTING.md)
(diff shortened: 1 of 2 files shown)


mmahmoudian:
@ZetaoYang The typo is correctly spotted and corrected, but I'm not sure about the dependency. Considering that you have made some of the github actions, perhaps you are the best to know the exact list of dependencies,

ZetaoYang:
@mmahmoudian Yes, `qttools5-dev` is a dependency of flameshot. You can see it here https://github.com/flameshot-org/flameshot/blob/master/data/debian/control .

ZetaoYang (approved):

PR #943, opened 2020-09-20
Fixed preview bug with circle counter and added ccache

### LINKED ISSUE #942: Circle counter tool does not show size indicator (Windows, Linux)
**Flameshot version**
0.8.0

**Describe the bug**
With common drawing tools, there will be an indicator following the cursor to show the size of the drawing tool:

(See the black square below):

图片

However on Windows, such indicator is missing for the circle counter tool:

(No square or anything at the cursor location except cursor (not included in the screenshot))

图片

I believe we should show an approximate size outline following the cursor so that users may know the size of counter circle.

**System version**

Tested to be reproducible on both Windows 10 and Debian Sid/Unstable.

borgmanJeremy:
Closes #942

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,9 @@ target_compile_features(project_options INTERFACE cxx_std_17)
 
 add_library(project_warnings INTERFACE)
 
+# enable cache system
+include(cmake/Cache.cmake)
+
 # standard compiler warnings
 include(cmake/CompilerWarnings.cmake)
 # set_project_warnings(project_warnings)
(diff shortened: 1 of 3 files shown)


hosiet:
Looks like the thickness was doubled when compared with the original implementation (`m_thickness * 2`). Is that the desired behaviour?

Besides, I would suggest splitting the ccache part into a separate PR/commit.

borgmanJeremy:
Sorry I was tired when I committed this :) I split the changes to circle counter into the other PR and am merging the cmake stuff separately. I added some comments in the code to explain the *2 (it is related to radius vs diameter).

Issue #941, opened 2020-09-20
Text in circle counter tool is always black (not working well with a dark circle)


hosiet:
**Flameshot version**
0.8.0

**Describe the bug**
![图片](https://user-images.githubusercontent.com/8443926/93691809-96412680-fab8-11ea-8d08-2d42418c829d.png)

As seen in the screenshot above, when the background color is set to black or near black, the counter number is still in black thus cannot be recognized (see the number "21" at the top).

**Expected behavior**
Maybe we should use the brightness information of current color and dynamically determine the color of counter number text. Choosing between black and white would be enough.

PR #944, opened 2020-09-20
Circle font color

borgmanJeremy:


borgmanJeremy:
Closes #941

borgmanJeremy:
Need to change color without QColorConstants. That is not introduced until Qt 5.15

hosiet (changes requested):
I saw you are using Hue value, this must be a typo. We should be looking at Lightness. Otherwise it would be like this:

图片

hosiet:
Besides, this PR mixed with other changes and maybe we can fix it.

hosiet:
This seems to be working, though the threshold can be further adjusted (see those dark blue circles below). Maybe we can merge it for now and tweak the threshold later. The fix for #942 is also working.

图片

borgmanJeremy:
Sounds good. If you find a value that works better feel free to adjust it in a pr.

Issue #813, opened 2020-09-03
can we have a RGB colour indicator?


wotmshuaisi:
a RGB colour indicator can display colour information on current mouse cursor coordinate.

this is very helpful to developer.

![image](https://user-images.githubusercontent.com/16408269/92108242-f52b4e00-edde-11ea-978e-18826a2dccf5.png)

Martin-Eckleben:
I personally would vote for it but right now we have a lot of Bugfixes to catch up first :)
Please don't be upset that it will probably take a while to get to this and thank you for your participation.

wotmshuaisi:
Thanks a lot, it's not urgent request to me, I'm just suggesting it. a vote would a great idea.

mmahmoudian:
@wotmshuaisi i believe what @veracioux implemented would do what you are looking for. It is not in a released version yet, and (assuming you are a Windows user based on the picture you shared) you should compile it yourself or wait for the new release (> v0.10.1)

borgmanJeremy:
Closing this after the work @veracioux  did.

wotmshuaisi:
> Does the color grabber coupled with the recently added hex code text box in the side panel satisfy this feature request?

It does help in most cases, except when the area you want to grab colour is covered by side panel.

mmahmoudian:
@wotmshuaisi but the sidepanel closes when the color grabber is selected and re-opens when the color is selected by the user or cancel by Esc

PR #952, opened 2020-09-21
#813 eyedropper tool

xahon:
Hello maintainers, i'm new to qt and this is my first pr into that repo. I'm going to maintain that project further
The #813 issue seems very straightforward to implement but i have some problems with the code
Couldn't you elaborate on why I can't make the window above the capture's one?
The idea is simple, show small frameless window (or widget, i don't yet know qt's terminology) with color info and keep it above the capture's window while the eyedropper tool is active

### DIFF
--- a/src/tools/CMakeLists.txt
+++ b/src/tools/CMakeLists.txt
@@ -48,6 +48,11 @@ target_sources(flameshot PRIVATE undo/undotool.h undo/undotool.cpp)
 
 target_sources(
   flameshot
+  PRIVATE eyedropper/eyedroppertool.h
+          eyedropper/eyedroppertool.cpp)
+
+target_sources(
+  flameshot
   PRIVATE abstractactiontool.cpp
           abstractpathtool.cpp
           abstracttwopointtool.cpp
--- a/src/tools/CMakeLists.txt
+++ b/src/tools/CMakeLists.txt
@@ -49,7 +49,9 @@ target_sources(flameshot PRIVATE undo/undotool.h undo/undotool.cpp)
 target_sources(
   flameshot
   PRIVATE eyedropper/eyedroppertool.h
-          eyedropper/eyedroppertool.cpp)
+          eyedropper/eyedroppertool.cpp
+          eyedropper/eyedropperwidget.h
+          eyedropper/eyedropperwidget.cpp)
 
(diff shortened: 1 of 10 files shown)


borgmanJeremy:
Hello,

Thanks for the help! I recommend implementing this in the side bar since we already have an eye dropper tool there. Perhaps right above the color wheel we can display the textual color information?

xahon:
@borgmanJeremy side panel is not convenient way to use the tool. It consumes side space and makes picking pixels on the right of the screen impossible. The tool way is much much better IMO

borgmanJeremy:
One issue we always need to balance is the number of tool buttons to keep the UI simple. Before we add this tool to the main tool menu I would like feedback from:
@hosiet @ZetaoYang @mmahmoudian

mmahmoudian:
About having this in sidebar, I agree with @xahon that the best position would be something that does not obstruct the view. Personally I prefer to have it following the mouse pointer with few pixel offset (and of course if the mouse goes to corners it just relocate to available position).

About having this in Flameshot, at the moment we have this concept in play that user can turn on or off the buttons from the config. We can use this to our advantage and have some of these tools as disabled by default and the user can customize the UI to their preference and their usage.

-------

I would also like to make a suggestion here, that we can have a plugin system for such tools. There are many feature requests that may be out of the context of a screenshot tool from our perspective, but some people might have a use for that feature being bundled with their screenshot tool. Therefore if we have a plugin system, people can also add third-party developed features to their setup. This way we still can stick to the constraints of the project while giving the community the freedom to extend the utilities of Flameshot.

ZetaoYang:
> @borgmanJeremy side panel is not convenient way to use the tool. It consumes side space and makes picking pixels on the left side of the screen impossible. The regular tool is much much better IMO

I also agree with xahon's opinions. I don't really like the design of the side panel.

borgmanJeremy:
Ok thanks for the feedback @xahon lets go forward with your implementation. I have a few other issues to catch up on but I can help support you if you get stuck on getting the widget to display on top.

xahon:
@borgmanJeremy i've managed getting the widget to top. My main issue is that cached widget pointer becomes null in hooks draw{Move, Start, etc..}

PR #956, opened 2020-09-22
Fix deb packaging for ubuntu 18.04

borgmanJeremy:


hosiet:
On Ubuntu 18.04 CI I suggest dynamically removing `cmake (>= 3.13~)` from `debian/control` build-dependency list prior to real packaging since we are installing cmake from snap instead of apt.

borgmanJeremy:
@hosiet Do we need to split data/debian into two folders....One for 18.04 and one for 20.04? (I'll be headed off for the night soon so feel free to pick up where I left off or I can come back to it tomorrow).

hosiet:
I am okay with dir splitting, it's just I will need to maintain 2 debian directories.

Of course my personal preference is to keep the file intact and use `sed` to dynamically replace/delete contents. Some dark magic like file symlinking may also work, depending on CI setup.

borgmanJeremy:
Ahh good idea. We are already doing plenty of sed hackery.

Issue #949, opened 2020-09-21
Ubuntu .deb package for v0.8.0 release cannot be installed on Ubuntu 18.04


hosiet:
**Flameshot version**
0.8.0

**Describe the bug**
* `flameshot_ubuntu_0.8.0-1_amd64.deb` as provided on the GitHub release page requires Qt 5.12 or later.
* Ubuntu 18.04 only provides Qt5 5.9.5.
* The deb package cannot be installed since the version requirement is not met.

borgmanJeremy:
We should probably tweak the CI to build against Qt 5.9.5 then. I am not sure if we are fully compatible with that version.

hosiet:
Or we could split the ubuntu build to be a ubuntu-18.04 one and a ubuntu-20.04 one. At least some optional features requires Qt 5.10+, like #318 .

hosiet:
@borgmanJeremy We really need to reopen this issue as the problem described in #959 was not really fixed.

borgmanJeremy:
Sorry it auto closed since the issues was linked. Thanks for pointing it out.

hosiet:
![图片](https://user-images.githubusercontent.com/8443926/94027924-5f556400-fd89-11ea-9da4-90999a8fdeca.png)

I am not sure why this would ever happen; are we building the 18.04 .deb file using 20.04 environment? Reading the github action file, it does not seem to be the intended behaviour but that is what happened. Sorry but I am not familiar with github action so I can't give advice on the fix now.

hosiet:
Could it be we are missing `env:   OS: ubuntu DIST: bionic` for the `ubuntu-18.04` build?

ZetaoYang:
@hosiet No, I try build deb on github actions ubuntu 18.04 images(I didn't use docker images to build it.), it already has cmake installed(cmake version 3.17.0). see https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md.

hosiet:
@ZetaoYang Then we need some further investigation since current outcome is not what we intended. This has nothing to do with cmake.

hosiet:
Alright, I think I get the reason. Here are the output from the ubuntu-18.04 environment from https://github.com/flameshot-org/flameshot/pull/962/checks?check_run_id=1155964243 :

```
gcc-7:
  Installed: 7.5.0-3ubuntu1~18.04
  Candidate: 7.5.0-3ubuntu1~18.04
  Version table:
[...]

PR #962, opened 2020-09-23
WIP: github workflow: make sure libgcc-s1 not install with ubuntu 18.04

hosiet:
This is a debug commit that try to look into the build issue with ubuntu 18.04.


hosiet:
This PR now find the cause of #949 . See the analysis at https://github.com/flameshot-org/flameshot/issues/949#issuecomment-697612265 .

Closing since this is not the real fix.

PR #963, opened 2020-09-23
fix: create a clean environment to build ubuntu-18.04 package

ZetaoYang:
Fix https://github.com/flameshot-org/flameshot/issues/949

First of all I wouldn't type a separate docker images for a particular environment (which contains a specific cmake version) it's not a flexible and elegant approach. 
Secondly, I'll just use the `vitzy/flameshot:ubuntu-bionic` image here, first removing the default cmake, and then downloading the latest binary from https://cmake.org/download/ to install it. I don't have a good way to add it to the existing matrix jobs(`deb-pack`). I had to create another job, which had the disadvantage of wasting a job count.

image

### DIFF
--- a/.github/workflows/Linux-pack.yml
+++ b/.github/workflows/Linux-pack.yml
@@ -30,12 +30,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        dist: [debian-10, ubuntu-18.04, ubuntu-20.04]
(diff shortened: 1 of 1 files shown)


borgmanJeremy (approved):

hosiet (approved):
Thanks, tested and this one is working.

PR #982, opened 2020-09-25
feat: using luminance to determine text colorof circlecounttool

### LINKED ISSUE #978: Improve contrast in circle counter tool
This is a follow-up and improvement for #941 and #944 .

v0.8.1 provided an initial fix to the font color issue of v0.8.0 (choosing black/white font according to the color of circle tool). The current implementation is to set a threshold on the Lightness in HSL color space (threshold=75). However, this does not work well with some colors like dark blue:

图片

Like the one in the image above, No.2 should use a white number instead of a black number.

* * *

It's a classic problem with many discussions before. What I found:

* https://stackoverflow.com/questions/1855884/determine-font-color-based-on-background-color
* https://css-tricks.com/switch-font-color-for-different-backgrounds-with-css/
* https://ux.stackexchange.com/a/114955
* Use "L" from [CIELab color space](https://en.wikipedia.org/wiki/CIELAB_color_space), which better fits human perception.

chux0519:
Fixes #978

looks better
flameshot

### DIFF
--- a/src/tools/circlecount/circlecounttool.cpp
+++ b/src/tools/circlecount/circlecounttool.cpp
@@ -89,8 +89,11 @@ void CircleCountTool::process(QPainter& painter,
           textRect, Qt::AlignCenter, QString::number(m_count));
     }
 
-    // Lightness value ranges from 0-255, we split at 75 as this looks best
-    if (m_color.lightness() <= 75) {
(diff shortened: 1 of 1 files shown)


borgmanJeremy (approved):

borgmanJeremy:
Awesome thanks for the contribution!

PR #989, opened 2020-09-26
Explicitly hide tray icon before exit on Windows

hosiet:
Currently flameshot tray icon will not disappear after program terminates on Windows. This is a known issue of Windows platform and need to be explicitly handled.

This commit connects the aboutToQuit signal to QSystemTrayIcon::hide slot to ensure that the tray icon is removed before program terminates.

### DIFF
--- a/src/core/controller.cpp
+++ b/src/core/controller.cpp
@@ -234,6 +234,13 @@ void Controller::enableTrayIcon()
         }
     };
     connect(m_trayIcon, &QSystemTrayIcon::activated, this, trayIconActivated);
+
+#ifdef Q_OS_WIN
+    // Ensure proper removal of tray icon when program quits on Windows.
+    connect(
+      qApp, &QCoreApplication::aboutToQuit, m_trayIcon, &QSystemTrayIcon::hide);
+#endif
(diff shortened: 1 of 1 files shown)


hosiet:
I just found that this fix seems to be working on my MINGW build but does not work on the MSVC build from GitHub Actions. Converting the PR to draft for further investigation.

hosiet:
It would be great if anyone can help debugging it and find the root cause.

borgmanJeremy:
It shows me that  MSVC built okay in github actions.

hosiet:
@borgmanJeremy It's that the MSVC build from GitHub Actions does not solve the bug. The build itself succeeded.

Maybe we can convert it into an issue instead.

ZetaoYang:
@hosiet I test this PR code on my local machine, it works compiled with MSVC 2019.

ZetaoYang:
> @borgmanJeremy It's that the MSVC build from GitHub Actions does not solve the bug. The build itself succeeded.
> 
> Maybe we can convert it into an issue instead.

github actions won't pull Draft PR code

Issue #928, opened 2020-09-16
Flameshot hardcodes ~/.config/autostart instead of using $XDG_CONFIG_DIRS


hosiet:
I was reading the source code and found that the source code handling autostart is to hardcode `$HOME/.config/autostart/`. According to the XDG specification (https://specifications.freedesktop.org/autostart-spec/0.5/ar01s02.html), we should be using `$XDG_CONFIG_DIRS/autostart` first and fallback to `~/.config/autostart` only when `$XDG_CONFIG_DIRS` was not set.

This would require a little bit of code refactoring, including searching of both canonical and fallback locations. Maybe we can do it after 0.8.x release.

adzimzf:
Hi, I'll try to work on this.
but since this is my first time contribute to flameshot, it may take a bit longer

borgmanJeremy:
@adzimzf Feel free to drop in our slack group to ask for advice in the dev channel or just pm me directly. flameshotworkspace.slack.com/

adzimzf:
thank you so much for warm welcome

mmahmoudian:
@borgmanJeremy Shall I assign this to @adzimzf to indicate others that they should not spend time on this?

borgmanJeremy:
Yup good idea

adzimzf:
Hi @borgmanJeremy 
How to log in to flameshotworkspace.slack.com, do I have to be invited first?

mmahmoudian:
@adzimzf Here you go: https://join.slack.com/t/flameshotworkspace/shared_invite/zt-hpogvgw3-MU3ODwLXe0CCzQinFRhp0g

adzimzf:
Thank you @mmahmoudian

PR #990, opened 2020-09-26
use $XDG_CONFIG_DIR, if it's not exist fallback to $HOME/.config

adzimzf:
Issue: #928 
use $XDG_CONFIG_DIR for default config dir, if not exist fallback to $HOME/.config

### DIFF
--- a/src/utils/confighandler.cpp
+++ b/src/utils/confighandler.cpp
@@ -297,7 +297,11 @@ bool ConfigHandler::verifyLaunchFile()
     bool res = false;
 
 #if defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
-    QString path = QDir::homePath() + "/.config/autostart/Flameshot.desktop";
+    QString path;
+    path = qgetenv("XDG_CONFIG_DIRS") + "/autostart/Flameshot.desktop";
+    if (!QFile(path).exists()) {
+        path = QDir::homePath() + "/.config/autostart/Flameshot.desktop";
+    }
     res = QFile(path).exists();
 #elif defined(Q_OS_WIN)
     QSettings bootUpSettings(
@@ -312,7 +316,12 @@ bool ConfigHandler::verifyLaunchFile()
 void ConfigHandler::setStartupLaunch(const bool start)
 {
 #if defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
-    QString path = QDir::homePath() + "/.config/autostart/";
+    QString path;
+    path = qgetenv("XDG_CONFIG_DIRS") + "/autostart";
+    if (!QFile(path).exists()) {
+        path = QDir::homePath() + "/.config/autostart/";
+    }
+
     QDir autostartDir(path);
     if (!autostartDir.exists()) {
         autostartDir.mkpath(".");


borgmanJeremy (approved):

src/utils/confighandler.cpp
```diff
@@ -297,7 +297,11 @@ bool ConfigHandler::verifyLaunchFile()
     bool res = false;
 
 #if defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
-    QString path = QDir::homePath() + "/.config/autostart/Flameshot.desktop";
+    QString path;
+    path = qgetenv("XDG_CONFIG_DIRS") + "/autostart/Flameshot.desktop";
```
borgmanJeremy, line 301: @hosiet If I understand the request we should actually be checking for the presence of the directory, not the file itself correct?

hosiet: Yes this was not correctly implemented. It should be we check the presence of environment variable to determine whether to use the path from the variable or do the fallback.

And there are many caveats here. What if the directory from envvar does not exist? What if the autostart dir does not exist? They all need to be handled to make the implementation robust.

adzimzf: thanks for the feedback,
@hosiet  What shall we do if the `autostart` dir doesn't exist? shall we create it?

borgmanJeremy: This is the specification: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

kpcee: Sorry but these changes do not make sense to me. With `QDir::homePath()` we get directly the $HOME path and this is all we need. 

The handling of the answer from qgetenv() is wrong. The answer is expected to be a single directory but the manual says 
> The directories in $XDG_CONFIG_DIRS should be seperated with a colon ':'.

In my understanding we have to treat the string as a list and see if we get several directories as result and check them one by one.

Last but not least, it makes imho no sense to handle the results of `XDG_CONFIG_DIRS` because the variable is set to /etc/xdg by default and the owner is root. In other words flameshot does not have the privileges to write or change files and the function `void ConfigHandler::setStartupLaunch(const bool start)` will always fail.

Just my two cents, this is just my opinion and should not be a criticism.

borgmanJeremy: @kpcee Thanks for chiming in, I agree with the first two points. 

For the 3rd point do you think we should just not follow the XDG Standard?

kpcee: In the source from QDir::homePath() is written that "Under non-Windows operating systems the HOME environment variable is used if it exists, otherwise the path returned by the rootPath()."

If you are concerned that the HOME variable is not set in some distributions or `QDir::homePath()` gives a wrong result or as fallback, I would use  `$XDG_CONFIG_HOME`.

`XDG_CONFIG_DIRS` seems to only list directories from root. This has two disadvantages:
* flameshot has no write permissions and would have to be started as root itself to make changes
* the changes apply system-wide for all users and not only for the currently logged in 

If you want to use the XDG standard instead of `QDir::homePath() `it would have to be done for the current user via `$XDG_CONFIG_HOME/autostart`, it always returns only a directory and no list. This is what I would change in the Pull Request, but I never had problems with `QDir::homePath()`.

adzimzf:
Hi,
since there's still discussion regarding the issue, I'll close the PR.
I'll try to look for another easy issue
thanks

src/utils/confighandler.cpp
chux0519: after reading the reference, seems `XDG_CONFIG_DIRS` is a sort of system-wide config paths, I think `XDG_CONFIG_HOME` is the env that we want to detect.
so I guess
1. check if `XDG_CONFIG_HOME` is set, if it is, use that
2. otherwise, if `XDG_CONFIG_HOME` is not set, we fall back to $HOME/.config, which is the same result of current implementation

PS: not only for autostart file, the config file should follow this rule too ($XDG_CONFIG_HOME/flameshot/flameshot.ini and fall back to ~/.config/flameshot/flameshot.ini) .

Issue #994, opened 2020-09-27
Flameshot will not invert the icon colour if the background colour is close to white


Tc-001:
**Flameshot version**
v0.8.1 (have tested appimage and AUR's flameshot-git)

**Describe the bug**
When setting the main or contrast color to white (or any other light color), the icon color does not invert, making the icons near impossible to see with certain colors

**To Reproduce**
- Open flameshot
- Select main color
- Set it to white

**Expected behavior**
The icon color (shown as a white circle) should change from white to black

**System Information**
Manjaro Linux 20.1 KDE

**Screenshots**
Expected:
![image](https://user-images.githubusercontent.com/55956895/94363001-bfb90e00-00c7-11eb-936d-a30362d4b44a.png)

What I got:
![image](https://user-images.githubusercontent.com/55956895/94363008-ccd5fd00-00c7-11eb-9c25-c7b7ce2dff6b.png)

This is also present in the screenshot interface for every icon except the resolution
![image](https://user-images.githubusercontent.com/55956895/94363099-8fbe3a80-00c8-11eb-982d-aa8309d32cb6.png)

mmahmoudian:
**Confirmed:**

![image](https://user-images.githubusercontent.com/390889/94364099-956b4e80-00cf-11eb-99fe-782babcf7cf8.png)

chux0519:
I think it is a tiny typo.

`ColorUtils::colorIsDark` should return `!isWhite`

mmahmoudian:
@chux0519 I saw your commit in your fork. Please create a PR when you confirmed that small modification has fixed the issue. 👍🏼

chux0519:
sure, check this out: https://github.com/flameshot-org/flameshot/pull/995

PR #995, opened 2020-09-27
fix:ColorUtils::colorIsDark

chux0519:


chux0519:
confirmed it works
image

mmahmoudian:
Nice spotting 👍🏼 , but as I suggested [in my comment in the issue](https://github.com/flameshot-org/flameshot/issues/994#issuecomment-699646590), I think it is more conventional not to negate the final boolean right before returning it. Therefore I suggest flipping the `true` and `false` on line 27 and 29 respectively.

chux0519:
@mmahmoudian sure, I will refactor it to avoid negating.

chux0519:
also, I found I wrote some duplicated code in my last PR: https://github.com/flameshot-org/flameshot/pull/982/files#diff-429e156c23094ea2e2cb049030fad1e9R92
there is already a helper function in src/utils/colorutils.cpp called `getColorLuma`. Can I fix this in this PR by the way ? or a separate one is better?

mmahmoudian:
@chux0519 I think @borgmanJeremy is the correct person to answer this, but as my personal opinion, as far as they are in separate commits it should be fine. Let's hear @borgmanJeremy 's opinion before you spend more time on this.

chux0519:
@mmahmoudian sure, I’m in GMT+8, and it’s a bit late now, I will continue this issue tomorrow, thanks for reviewing

borgmanJeremy:
We can do it all in one PR. Would be nice to update the circle counter to use this code also, I think it is also duplicated there.

chux0519:
@mmahmoudian your first thought is right, the problem is not about the `colorIsDark` function. It should be not saving `m_mainColor` in `CaptureToolButton`, so when do `updateIcon`, it still use the default main color QColor(116, 0, 150) which is wrong.

I also refactored the duplicated part I mentioned above. confirmed it works.


image

src/utils/colorutils.cpp
```diff
@@ -24,11 +24,8 @@ inline qreal getColorLuma(const QColor& c)
 
 bool ColorUtils::colorIsDark(const QColor& c)
 {
-    bool isWhite = false;
-    if (getColorLuma(c) <= 0.60) {
```
borgmanJeremy, line 28: Do you know why the previous value was 0.6 and we are now changing to 0.5? Maybe 0.6 looked better with certain colors?

borgmanJeremy (commented):
Changes look good but curious on why the threshold changed and what that will mean.

src/utils/colorutils.cpp
chux0519: Honestly, I'm not sure.
Due to some articles linked by issue [978](https://github.com/flameshot-org/flameshot/issues/978).Seems the threshold should be 0.5 ~ 0.7. Would it be better if I change this back to 0.6 ?

chux0519: I chose 0.5 because of the answers link below:

https://stackoverflow.com/questions/1855884/determine-font-color-based-on-background-color

seems all answers using 0.5

borgmanJeremy: Okay great thanks for the additional info.

PR #996, opened 2020-09-27
Update Internationalization_cs.ts

pafri:
Czech translation update of my previous work. Please commit it. Best regards.


mmahmoudian:
Considering that all changed are translation of previously untranslated strings, this is an **approved** PR.

mmahmoudian (approved):
Considering that all changed are translation of previously untranslated strings, this is an approved PR.

Issue #1012, opened 2020-09-30
Kvantum support broken with version 0.8.3-1 on Arch Linux/ArcoLinux


erikdubois:
**Flameshot version**
0.8.3-1
https://www.archlinux.org/packages/community/x86_64/flameshot/

**Describe the bug**
We had the possibility to theme the right mouse click menu on the system panel icon to follow the arc dark theme with kvantum manager. This is version 0.8.1

![before](https://user-images.githubusercontent.com/10594806/94676377-8b3d9000-031b-11eb-81dd-dae267a7921e.png)

Today we get version 0.8.3 in and we see that our theme is not applied..

![after](https://user-images.githubusercontent.com/10594806/94676400-942e6180-031b-11eb-8c53-eb42f923284e.png)

Could you investigate what part of the code is responsible for that?

Downgrading the package fixes the color scheme set by kvantum.

borgmanJeremy:
I think this will be fixed in #1013. Are you able to test it?

PR #1013, opened 2020-09-30
Moved style override to after app creation

borgmanJeremy:
This fixes the issue where flameshot does not inherit the system theme. 

From: https://doc.qt.io/qt-5/qapplication.html#setStyle
```
Setting the style before a palette has been set, i.e., before creating QApplication, will cause the application to use QStyle::standardPalette() for the palette.
```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -40,7 +40,6 @@
 
 int main(int argc, char* argv[])
 {
-    QApplication::setStyle(new StyleOverride);
     // required for the button serialization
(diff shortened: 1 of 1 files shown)


hosiet (approved):
This looks good to me. The previous regression is fixed.

erikdubois:
This code change works for ArcoLinux - Arch Linux
Roll it out please.
see #1012

borgmanJeremy:
I will do a release over the weekend. Thanks.

PR #1015, opened 2020-09-30
Open imgur upload in screen center

### LINKED ISSUE #806: Imgur save window never appears in the screen center
Captura de tela_2020-08-26_01-11-20

The 'upload to imgur' window never appears in the center of the screen, so I have to drag it up. Is there a way to make it appear on the center or top of screen? Would be awesome to make it just upload to imgur and copy the link to clipboard without buttons or a window.

Thank you so much.

AlexP11223:
Fixes #806

The same code as in `InfoWindow`.

### DIFF
--- a/src/tools/imgur/imguruploader.cpp
+++ b/src/tools/imgur/imguruploader.cpp
@@ -25,8 +25,10 @@
 #include <QApplication>
 #include <QBuffer>
 #include <QClipboard>
+#include <QCursor>
 #include <QDesktopServices>
 #include <QDrag>
+#include <QGuiApplication>
 #include <QHBoxLayout>
 #include <QJsonDocument>
 #include <QJsonObject>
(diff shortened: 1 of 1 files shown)


src/tools/imgur/imguruploader.cpp
```diff
@@ -48,6 +52,13 @@ ImgurUploader::ImgurUploader(const QPixmap& capture, QWidget* parent)
     setWindowTitle(tr("Upload to Imgur"));
     setWindowIcon(QIcon(":img/app/flameshot.svg"));
 
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 10, 0))
```
borgmanJeremy, line 55: Do you know if there is an alternative API that supports QT 5.9?

AlexP11223: I am not sure which part of this is not supported in 5.9 :D

borgmanJeremy: Why is there a QT version check macro?

AlexP11223: ah, it's `screenAt` https://github.com/flameshot-org/flameshot/pull/318#issue-209315899

PR #1019, opened 2020-10-01
Update desktop entry exec

heis2201:
In Gnome (tested on Fedora 33), the command to run flameshot would be `flameshot gui`


borgmanJeremy:
I believe "Flameshot" is correct. This will open the tray icon. There is another open issue #1021 saying on Fedora33 with the new gnome there is an issue with the tray icon not appearing unless you install legacy icons.

xvitaly:
> There is another open issue #1021 saying on Fedora33 with the new gnome there is an issue with the tray icon

Not only Fedora, but all GNU/Linux distributions with Gnome 3.24+.

borgmanJeremy:
This is an issue with gnome though, we provide an icon that is started with flameshot, gnome just no longer comes with systray support.

PR #588, opened 2019-08-05
Create internationalization_eu.ts

alexgabi:
This pull request refreshes eu (Basque) translation.

PR #1027, opened 2020-10-02
Updated Basque translation

Porrumentzio:
Since the last update I had today on Manjaro, from `0.6` to `0.8`, I noticed that the Basque translation appeared on my system.

But there was a error: it was **not** the translation I made some weeks ago, but another one (with typographic errors).
And in fact, in the code, the translation that appeared was `eu_ES`, while what I made was `eu`, as seen in the [first commit](https://github.com/flameshot-org/flameshot/commit/16c1328006cb611d2b533f85e7572ab476378ebe).

I updated the translation and renamed it. Please, for the next version, add only this one, `eu`, and not `eu_ES`.

Cheers :)


Porrumentzio:
About the language code, seems that there was a previous translation in which @ZetaoYang said that the language code should be `eu_ES`, when should **not**.
This time, please, leave it as `eu`.
https://github.com/flameshot-org/flameshot/pull/588#issuecomment-690126196

ZetaoYang:
@Porrumentzio Please update `.ts` file name in `src/CMakeLists.txt` as well.

PR #1041, opened 2020-10-10
Improve pixelation of thin selections

sersorrel:
Assuming I'm correctly understanding the code: Previously the level of pixelation would decrease if the selected area was too thin to hold 10 pixelated blocks. Now the level of pixelation only decreases when the selected area is too thin to hold even one pixelated block. This makes it easier to hide information in relatively thin selections.

This doesn't really solve the issue of pixelation decreasing when the area is thin, see these screenshots (taken with thickness=80) and note that the horizontal one has very large blocks still, but the vertical one has significantly smaller blocks than it should:

image

image

It does significantly improve the situation, though.


borgmanJeremy:
Thanks for the contribution!

PR #1051, opened 2020-10-12
fixing icon regression

borgmanJeremy:


hosiet:
This looks okay to me, but I have no idea whether this breaks Flathub's requirement. Please double check with that part.

borgmanJeremy:
@hosiet Will do, I'll kick off that build next

hosiet:
One more question: is CMake installing duplicated real image files instead of symlinks? I don't have computers around so please check it. It seems to be the case from CMakeLists.txt (those COPYONLY lines are not installing symlinks at all).

borgmanJeremy:
@hosiet Yes its copying the file twice. It seemed that cmake method did not allow for symlinking. I can manually call to the shell to make the symlink if we want:

https://stackoverflow.com/questions/35765106/symbolic-links-cmake

ZetaoYang:
It is recommended to refer to copyq for this way of handling icon names.

borgmanJeremy:
@ZetaoYang I think that will have the same problem because it will only create a org.flameshot.Flameshot.svg....it also needs a flameshot.svg.

borgmanJeremy:
@ZetaoYang I think that will have the same problem because it will only create a org.flameshot.Flameshot.svg....it also needs a flameshot.svg.

xvitaly:
@borgmanJeremy Do not install duplicates. GNU/Linux maintainers now need to remove them manually.

On Flatpak's side you should use `"rename-icon": "flameshot"`.

borgmanJeremy:
@xvitaly I am symlinking them now. Is that still an issue? If you have a way to only have one icon set work for all the different packages please submit a PR. I am honestly very burned on on working on packaging conflicts and the current solution is working.

xvitaly:
> I am symlinking them now. Is that still an issue?

Yes. 1 icon for 1 application.

> If you have a way to only have one icon set work for all the different packages please submit a PR

You need to revert all Flathub* commits and then use `"rename-icon": "flameshot",` on Flathub's side.

borgmanJeremy:
thanks!

xvitaly:
@borgmanJeremy https://github.com/flameshot-org/flameshot/pull/1057

PR #1066, opened 2020-10-14
Added code signing information

borgmanJeremy:


README.md
```diff
@@ -367,11 +369,17 @@ Note: If you install from source, there is no uninstaller, you will need to manu
 
 Info: If I take code from your project and that implies a relicense to GPLv3, you can reuse my changes with the original previous license of your project applied.
 
-## Contribute
+## Privacy Policy
+This program will not transfer any information to other networked systems unless specifically requested by the user or the person installing or operating it.
```
mmahmoudian, line 373: Should we also state that we do not have telementry?

mmahmoudian (approved):

README.md
borgmanJeremy: The snap release actually does have minimal telemetry

PR #1074, opened 2020-10-15
Updated releasing instructions

borgmanJeremy:


borgmanJeremy:
@ZetaoYang Seems like something broke with the environment changes?

ZetaoYang:
> @ZetaoYang Seems like something broke with the environment changes?

Because in your update_release_Doc branch, the latest version in the git tag (0.8.5)doesn't match the version in CMakeLists.txt(0.8.4).
I suggest rebase master branch first.

https://github.com/borgmanJeremy/flameshot/blob/update_release_Doc/CMakeLists.txt

PR #1090, opened 2020-10-22
Fixed OpenSSL Cmake, and removed close after upload. This did not pla…

borgmanJeremy:
…y nice with notifiactions and I think the delete window should always appear since files cannot be deleted


borgmanJeremy:
Actually just fixed the close after upload bug on 0.8.x maintenance branch, will adjust this PR accordingly.

Issue #812, opened 2020-09-03
Hot key to commit text and image in one shot


amosbird:
When typing in a text area, `<enter>` key doesn't commit but inserts a new line. I have to left-click on the image to first commit the text area, then press `<enter>` to commit the image. It would be more convenient to have a hotkey say `shift-<enter>` to insert newlines while keep `<enter>` key as the commit key.

Martin-Eckleben:
Hi @amosbird,

thank you very much for your proposal and participation.
Right now we have a lot to catch up to. Please don't be upset that it will take a while until we get to discuss it but it will happen eventually :)

Personally I'm undecided.
I hate ```alt-<enter>``` in most spreadsheet editors. It confuses most non computer people and I really often see them not doing any newlines which often hurts the readability.
But there are prominent examples which do it i.e. Whatsapp comes to my mind first.

It needs discussion.

amosbird:
Many IM applications uses shift-enter to insert newlines while sending messages by enter. for instance, telegram, discord, gitter, slack, etc.

PR #1110, opened 2020-10-28
Add a configurable keyboard shortcut to commit text.

jfly:
This new shortcut defaults to `ctrl+enter`, and it commits the text
you're currently typing. It does *not* change the behavior of enter
(pressing enter still inserts a newline), nor does it actually exit
flameshot. This isn't exactly what was requested, but I'm going to
somewhat cavalierly say that this fixes
https://github.com/flameshot-org/flameshot/issues/812.

*Note: I added a new user-visible string here, but I didn't update any
ts files. Is that ok?

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -235,6 +235,26 @@ QPixmap CaptureWidget::pixmap()
     return m_context.selectedScreenshotArea();
 }
 
+// Finish whatever the current tool is doing, if there is a current active
+// tool.
+bool CaptureWidget::commitCurrentTool()
+{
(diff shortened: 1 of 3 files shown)


src/widgets/capture/capturewidget.cpp
```diff
@@ -235,6 +235,26 @@ QPixmap CaptureWidget::pixmap()
     return m_context.selectedScreenshotArea();
 }
 
+// Finish whatever the current tool is doing, if there is a current active
```
borgmanJeremy, line 238: +1 for moving this to a function :+1:

borgmanJeremy (approved):

PR #1116, opened 2020-10-29
Look up the autostart directory using QT's helper methods.

jfly:
This fixes https://github.com/flameshot-org/flameshot/issues/928.


borgmanJeremy:
Wow had no idea QT had a library for this. Great find.

PR #1122, opened 2020-10-30
Symmetric new selection

astrosticks:
Enables the symmetry modifier key (shift) to work not only when dragging on an existing selection handle, but also when creating a new selection (centered about the initial click).

Other changes:
- Refactor different mouse selection modes to only define the new selection area, not implement the entire interaction
- Fix a typo in SelectionWidget enum

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -407,22 +407,23 @@ void CaptureWidget::mouseMoveEvent(QMouseEvent* e)
         QRect inputRect;
         if (m_newSelection) {
             // Drawing a new selection
-            inputRect = symmetryMod 
-                ? QRect(m_dragStartPoint*2 - m_context.mousePos, m_context.mousePos)
-                : QRect(m_dragStartPoint, m_context.mousePos);
+            inputRect = symmetryMod
+                          ? QRect(m_dragStartPoint * 2 - m_context.mousePos,
+                                  m_context.mousePos)
+                          : QRect(m_dragStartPoint, m_context.mousePos);
 
         } else if (m_mouseOverHandle == SelectionWidget::NO_SIDE) {
             // Moving the whole selection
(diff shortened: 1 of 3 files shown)


borgmanJeremy:
Typically I associate the shift modifier with locking the aspect ratio. I like the symmetric idea but maybe we need to move it to a different hotkey so shift can be used for locked aspect ratio?

@mmahmoudian thoughts?

astrosticks:
Hmm, I can't remember where else I've associated shift with symmetry aside from Inkscape. But, flameshot already uses shift as the symmetry modifier when pulling the handles on a selection. Krita seems to agree with you, though. It uses shift for locking aspect ratio and control for symmetry.

mmahmoudian:
The idea is brilliant and I can't believe it never occurred to me! I support the idea.

About the mod key, I'm writing this from my phone, so I have not opened any software and fact check, but:
1. in most softwares I always struggle to find that key esp when I want to draw perfect circle (instead of oval) or perfect square instead of rectangle. I think in inkscape Ctrl is for that and shift is for symmetry, but I always have to spend few seconds try all these combinations amd the work with inkscape or LibreDraw or Krita.
2. there is a logic behind naming of keys on keyboard. Alt should alternate the function, Ctrl is Control and etc. but now a days the naming is not typically followed by softwares and I don't know of any well-established standard.

**Final thoughts and remark**

1. I personally would suggest going with inkscape keyboard shortcut scheme as it is also used by many Windows and OSX users
2. I would also suggest having the following features (probably I should open an issue for it)
    - use Ctrl to draw perfect square selection area
    - use Ctrl to draw perfect square and circle with tools
    - use a kry combination to resize the selection area and respecting the aspect ratio

borgmanJeremy:
@mmahmoudian good suggestions. Let's open another issue for the locked aspect ratio and perfect shapes. 

@astrosticks Before merging this can we make the hotkey configurable through the new hotkey system? Then users can change it if they want to.

Issue #792, opened 2020-07-31
changing brush size in the UI


soundsbeard:
im using graphical tablet almost all the time, and if i need to change size of the arrow, pencil or whatever i need to grab a mouse to use a scroll wheel.
please, add size changing to UI.
and thank you for a great tool!

mmahmoudian:
@soundsbeard There has been improvements regarding this issue. Considering that you have the appropriate hardware, would you please test this:
https://github.com/flameshot-org/flameshot/issues/1003#issuecomment-700240081

soundsbeard:
@mmahmoudian 
thank you for the reply.
i can confirm, that both gestures and tablet scroll circle work for changing size.
but its not an option for me.
im a 3d artist, and as i said previously, i use tablet all the time, and my workflow means using keyboard shortcuts all the time, so all my touch functions for the tablet are disabled bc i keep tapping somewhere by accident. so it would be great to have ui thingy for changing size. but i know that i might ask too much.
anyway, flameshot is the best screenshot tool already, thank you and your team for the hard work.

mmahmoudian:
@Martin-Eckleben Seems your code works. Perhaps we should also add an option in the "interface" tab of the configuration window to add + and - buttons (disabled by default imho). What do you think?

borgmanJeremy:
@mmahmoudian  @Martin-Eckleben  I think this is a good idea. Let's turn off by default but get it added in.

I had not considered the case where gestures are turned off.

PR #1131, opened 2020-11-01
Sizetools

### LINKED ISSUE #1003: Add optional buttons to increase / decrease toolsize for tablets etc.
Forked from #782 as its a different issue I think.
Almost done with it. Created + / - Icons in B&W. Created the tools. Made them default configured tools. etc. etc.
Just needs one more evening with a coffee which should be 30.09.2020 UTC+2 :)

Martin-Eckleben:
Fixes #1003
"Add optional buttons to increase / decrease toolsize for tablets etc."

### DIFF
--- a/src/tools/sizedecrease/sizedecreasetool.cpp
+++ b/src/tools/sizedecrease/sizedecreasetool.cpp
@@ -1,0 +1,59 @@
+// Copyright(c) 2017-2019 Alejandro Sirgo Rica & Contributors
+//
+// This file is part of Flameshot.
+//
+//     Flameshot is free software: you can redistribute it and/or modify
+//     it under the terms of the GNU General Public License as published by
+//     the Free Software Foundation, either version 3 of the License, or
+//     (at your option) any later version.
+//
+//     Flameshot is distributed in the hope that it will be useful,
+//     but WITHOUT ANY WARRANTY; without even the implied warranty of
+//     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//     GNU General Public License for more details.
+//
+//     You should have received a copy of the GNU General Public License
+//     along with Flameshot.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "sizedecreasetool.h"
+#include <QPainter>
+
+SizeDecreaseTool::SizeDecreaseTool(QObject* parent)
+  : AbstractActionTool(parent)
+{}
+
+bool SizeDecreaseTool::closeOnButtonPressed() const
+{
+    return false;
+}
+
+QIcon SizeDecreaseTool::icon(const QColor& background, bool inEditor) const
+{
+    Q_UNUSED(inEditor);
+    return QIcon(iconPath(background) + "minus.svg");
+}
+QString SizeDecreaseTool::name() const
+{
+    return tr("Increase Tool Size");
+}
+
+ToolType SizeDecreaseTool::nameID() const
+{
+    return ToolType::SIZEDECREASE;
+}
+
+QString SizeDecreaseTool::description() const
+{
+    return tr("Decrease the size of the other tools");
+}
+
+CaptureTool* SizeDecreaseTool::copy(QObject* parent)
+{
+    return new SizeDecreaseTool(parent);
+}
+
+void SizeDecreaseTool::pressed(const CaptureContext& context)
+{
+    Q_UNUSED(context);
+    emit requestAction(REQ_DECREASE_TOOL_SIZE);
+}
--- a/src/tools/sizedecrease/sizedecreasetool.cpp
+++ b/src/tools/sizedecrease/sizedecreasetool.cpp
@@ -34,7 +34,7 @@ QIcon SizeDecreaseTool::icon(const QColor& background, bool inEditor) const
 }
 QString SizeDecreaseTool::name() const
 {
-    return tr("Increase Tool Size");
+    return tr("Decrease Tool Size");
 }
 
 ToolType SizeDecreaseTool::nameID() const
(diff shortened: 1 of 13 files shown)


Martin-Eckleben:
Sorry this got pretty messy because I took so long between 'init' and 'done'.

src/tools/sizedecrease/sizedecreasetool.cpp
```diff
@@ -0,0 +1,59 @@
+// Copyright(c) 2017-2019 Alejandro Sirgo Rica & Contributors
```
AlexP11223, line 1: Do we really need to add the license header to every file? =) Also the date seems wrong.

GPL only recommends that.

https://www.gnu.org/licenses/gpl-faq.html#NoticeInSourceFile

> You should put a notice at the start of each source file, stating what license it carries, in order to avoid risk of the code's getting disconnected from its license. If your repository's README says that source file is under the GNU GPL, what happens if someone copies that file to another program? That other context may not show what the file's license is. It may appear to have some other license, or no license at all (which would make the code nonfree). Adding a copyright notice and a license notice at the start of each source file is easy and makes such confusion unlikely.This has nothing to do with the specifics of the GNU GPL. It is true for any free license.

But it doesn't sound like a good reason, because if someone is okay with violating the license they can just remove the header when copying. :thinking: 

And if they are not intending to violate, then they should just check where this file came from, not sure how it could get disconnected so severely that it would be difficult. It’s always possible just to google the class name, etc. if everything else is lost somehow. :)

src/tools/sizedecrease/sizedecreasetool.h
```diff
@@ -0,0 +1,41 @@
...
+//     GNU General Public License for more details.
+//
+//     You should have received a copy of the GNU General Public License
+//     along with Flameshot.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include "src/tools/abstractactiontool.h"
+
+class SizedecreaseTool : public AbstractActionTool
```
AlexP11223, line 22: Maybe should be `SizeIncreaseTool`, `SizeDecreaseTool`

ZetaoYang:
I tried this PR Windows CI build, the new added icon color should be black rather than white. Another problem is that you can't go back and switch between the two checkboxes. I suggest installing the CI daily build package and testing it first.
image

ZetaoYang:
Another bug on this PR, 
image

ZetaoYang:
Plus, I don't really like the idea of setting the tool size in such a way.

src/tools/sizedecrease/sizedecreasetool.cpp
Martin-Eckleben: Hmm you are right - should be discussed if we do that in the future. Will ask everybody in Slack.

src/tools/sizedecrease/sizedecreasetool.h
Martin-Eckleben: Yea makes sense.
Probably subconsciously tried to "save" camelcase words - which is nonsense :D

Martin-Eckleben:
@ZetaoYang "Plus, I don't really like the idea of setting the tool size in such a way."

It's not intended as standard - only as optional possibility for users on Tablets / without a mouse etc.
Was discussed and decided together with @borgmanJeremy and @mmahmoudian .

What do you guys think?
Should we add it or not?

Martin-Eckleben:
@ZetaoYang 
"Another problem is that you can't go back and switch between the two checkboxes."
Ooh this is bitter. How could I not notice that O.o
Will fix for sure.

mmahmoudian:
This feature was requested multiple times and as far as I'm concerned, it is intended for users with touch interface (e.g tablet) as they cannot "scroll" to change the size of the tools. Therefore I think it is a good addition to Flameshot as it will broaden the devices the software can be used on.

borgmanJeremy:
As @mmahmoudian indicated it is intended to help users with touch interfaces.

AlexP11223:
Do the current tool buttons work well for them though? I don't have any touch device that can run Flameshot, but the buttons may be too small and too close to each other for touch usage.

mmahmoudian:
The OP in #792 was kind enough to test a build once. Perhaps he can also provide feedback about the button sizes.

soundsbeard:
> The OP (@soundsbeard) in #792 was kind enough to test a build once. Perhaps he can also provide feedback about the button sizes.

unfortunately, im not a coder and i have no idea how to compile flameshot to test changes.
the last time i've checked for gestures to work, they were already working in a release build.

mmahmoudian:
@soundsbeard You don't need to, we have automated the building step with Github Actions. Follow the following steps to get the file (I thought directly giving you the link would not be as good as showing you how to get it):
image
image

soundsbeard:
> I thought directly giving you the link would not be as good as showing you how to get it

sorry for the late reply
thank you! i love learning new things.
though, i use manjaro plasma, so i've downloaded appimage, bc i haven't found any arch based packages (i don't even know if any exist).

i've turned on resize tool, and there was an issue, that after i've resized configuration window all checkboxes gain offset, so when i tried to click them i've clicked somewhere else. after clicking 'select all' flameshot crashed.
then i finally turned on tools and tested them. there is another issue, that a circle with size value appears on another monitor. but if i try to scroll with mouse wheel it jumps to the main monitor.

i assume that its not the fault of the flameshot, bc my second monitor is very old and plugged in through DP to VGA adapter, so i have similar issues with different apps on a daily basis.
overall, its great to have a possibility to change the size of tools with wacom pen! thank you guys! cant wait to have the update in the AUR repos.

borgmanJeremy:
Will merge via #1426 to resolve merge conflicts. Sorry for how long this took me to merge.

PR #1135, opened 2020-11-02
Enforce utf-8 charset when building flameshot

ZetaoYang:


borgmanJeremy:
Is there a reason this only applies to Windows?

borgmanJeremy (approved):

ZetaoYang:
> Is there a reason this only applies to Windows?

This only for msvc compiler, because there is a full-width character "/" in flameshot code. When I run code in my local machine, it will appear an error.

PR #1152, opened 2020-11-11
Added screenshot with interval

timurchak:


borgmanJeremy:
Have you tried setting a delay through the launcher which can be activated from the tray icon?

timurchak:
> Have you tried setting a delay through the launcher which can be activated from the tray icon?

You are absolutely right =)

borgmanJeremy:
Sorry my question was, is there a reason to add this feature since we support delays in the launcher?

timurchak:
@borgmanJeremy Sorry, apparently I didn't see it.  after looking better, found this [discussion](https://github.com/flameshot-org/flameshot/issues/582) and indeed such a function is already there.  Sorry, the pull can be closed =)

PR #1166, opened 2020-11-22
Updated the bug report template

mmahmoudian:
- Updated the URL of the issue reporting help page
- Enforced 80 character line length (to ensure all text will be visible to user without any horizontal scrolling)
- Added a statement that bug reports lacking sufficient information might not get processed

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -5,15 +5,21 @@ about: Create a report to help us improve
 ---
 
 <!--
-If you don't know how to get some of the following information from your computer, visit:
-https://flameshot.js.org/#/data/issue-reporting
+If you don't know how to get some of the following information from your
+computer, visit:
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Should we add a third template for asking simple questions? I don't want to scare away first time posters with the mandatory info, but we also need to constantly ask for it so it is really mandatory.

mmahmoudian:
tbh, seeing a bug report without proper information looks like a wild-goose chase. It gets even more hurtful when at the end after insisting on the info, we get to know that this is a known issue (e.g in case of wayland).


I agree with your point that making things mandatory make the new users less likely to report stuff. We can for sure add a third category for questions. I'll look into it and add to this PR.

I think the "most proper" way would be to add a button/command line argument that produce most of these info so that users just press a button and paste the output for us. I don't know if this is doable for Windows, but for Linux we can bundle together a sequence of commands and save the output to a file.

borgmanJeremy:
Yeah a script that generates the info is a great idea.

PR #1191, opened 2020-12-17
Update Internationalization_sv_SE.ts

eson57:
Correct and complete Swedish translation.
I also fixed the source string on line 438, to better match GUI and translate. You might want to fix that error in other translations. Thanks!


borgmanJeremy:
Thanks!

PR #1223, opened 2021-01-05
first pass at adding logging framework

borgmanJeremy:
@mmahmoudian A while back you asked to have logging / better debugging info available. This is a first pass at integrating that. Right now I just have the "circle counter" instrumented to help debug the issue with missing numbers. I'll continue to improve the framework.


mmahmoudian:
tbh the code change is so much that I cannot easily understand what it does based on the changed files in this commit. 😅

But overall I think this is a great step into a good direction.

borgmanJeremy:
Most of that is the external library, but we will be able to start annotating the code like this:
image

I'll make it an environmental variable before adding to an official release.

PR #1257, opened 2021-01-25
Show the Help Text, Tool Settings on the current active monitor

Mantas-2155X:
**Issue:** Help text and tool settings button, panel always displayed on the primary monitor.

**Solution:** Make the help text and tool settings button, panel display on the current active monitor. The active monitor is determined by the mouse position when flameshot gui was ran.

**System information:**
Distribution: Kubuntu 20.10
Kernel version: Linux Kernel 5.9.16-tkg-upds
Flameshot version: 265e42c

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -322,9 +322,9 @@ void CaptureWidget::paintEvent(QPaintEvent*)
     painter.setClipRect(rect());
 
     if (m_showInitialMsg) {
-#if ((QT_VERSION >= QT_VERSION_CHECK(5, 10, 0) && defined(Q_OS_LINUX)) ||      \
-     defined(Q_OS_MAC) || defined(Q_OS_MAC64) || defined(Q_OS_MACOS) ||        \
-     defined(Q_OS_MACX))
+#if (defined(Q_OS_LINUX) && QT_VERSION < QT_VERSION_CHECK(5, 10, 0))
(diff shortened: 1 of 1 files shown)


Mantas-2155X:
This should be all the changes for the PR. If there is anything else, please do tell me

src/widgets/capture/capturewidget.cpp
```diff
@@ -322,7 +322,8 @@ void CaptureWidget::paintEvent(QPaintEvent*)
     painter.setClipRect(rect());
 
     if (m_showInitialMsg) {
-#if (defined(Q_OS_MAC) || defined(Q_OS_MAC64) || defined(Q_OS_MACOS) ||        \
```
borgmanJeremy, line 325: Do you happen to know if this will also work on Windows so we can remove the platform check? I don't have a Windows box handy.

borgmanJeremy (approved):

src/widgets/capture/capturewidget.cpp
Mantas-2155X: I don't have Windows but I do have a Windows 10 VM set up. After my class ends I will check it out and see if it works.

Mantas-2155X: I just tested the automatic build from 6523c34 on Windows 10. Seems like it works, but there is one problem - the tool settings window pops up on the primary monitor, the rest seems to open on the current active monitor like intended.

I am not sure on why this is happening and I have no experience on working with Qt applications.

Issue #1243, opened 2021-01-17
JPG for clipboard


dd1303:
Hi,

is it already possible to save the screenshot as JPG to clipboard instead of PNG? If not already available ... would it be possible to plan it for a future release?

This would save a lot of space in my emails. For example a 50%-quality JPG takes 1/20 of the size compared to PNG

Of course I found the save button (floppy disk symbol) which can save as PNG, BMP and JPG. But that way includes too mayn steps later to include it into my emails and documents.

I use flameshot v0.8.5 on Lubuntu 18.04 only since recently and I'm really happy otherwise about the features, ease of use and stability of flameshot.

Best regards,
Dietrich

borgmanJeremy:
This is a good idea. I think we could add an option for this.

sonulohani:
Hi @borgmanJeremy I want to work on this issue. Could you please assign this issue to me. Thanks.

borgmanJeremy:
@sonulohani  Sure, let me know if you want any help or direction. Either can leave comments here or hit me up on our slack: https://join.slack.com/t/flameshotworkspace/shared_invite/zt-kx50keyc-H2GHTjQmmC4X2oRgSeDIVg

sonulohani:
@borgmanJeremy Thanks for the directions. I will sure connect with you if I would have any problem.

dd1303:
Hi Jeremy, Hi Sonu Lohani,

first of all - thank you very much for accepting and implementing my
issue in such a short time :o)

Of course I wanted to use the new feature, but I failed. Probably I'm
doing something wrong.

I'm new to github and I'm not sure how handle my question to you. Just
an email like this? Or is there a better way, like a forum or opening a
new issue or opening the old issue?

The problem:

I still only see PNG on the clipboard. I can see PNG with

a) Commpose a Thunderbird email and look a the email source with Control U
b) Inspect Clipboard with Qcopy




What I did ...

I followed you compilation readme:

pkill flameshot

cd ~didi
mkdir -p git/repo
cd git/repo
git clone https://github.com/flameshot-org/flameshot

LANG=C # only for git infos
$ git status
On branch master
Your branch is up to date with 'origin/master'.
nothing to commit, working tree clean

$ git log
...
commit a944860d12bf0dab925433ef5a039cf969bd1244
Author: Sonu Lohani <sonulohani@gmail.com>
Date:   Sat Jan 30 10:01:14 2021 +0530

     Add config for jpg/png when copy

[...]

PR #1263, opened 2021-01-29
JPG for clipboard

sonulohani:
Issue detail: https://github.com/flameshot-org/flameshot/issues/1243


borgmanJeremy (commented):
I think we should make this a configurable option so users can choose.

sonulohani:
> I think we should make this a configurable option so users can choose.

Okay

borgmanJeremy:
Did minor tweaks and merged on #1273 

Thanks for the contribution!

PR #1272, opened 2021-02-02
Add support for sway/wlroots Wayland compositors

nullobsi:
This PR adds code for using `xdg-desktop-portal-wlr` to take screenshots when XDG_CURRENT_DESKTOP is set to Sway. It currently works with just 1 non-scaled monitor.

To-do:

- [x] fix hidpi/multiple monitors (fullscreen across all displays)
- [x] fix issue where program restart is required to take more than 1 screenshot (see https://github.com/emersion/xdg-desktop-portal-wlr/pull/82)


nullobsi:
This PR now works fine on my machine with 2 monitors of varying DPIs.

borgmanJeremy (approved):
Is the intent of the proxy class to make it usable for other wayland implementations?

nullobsi:
Without the proxy class, there's no way to subscribe the Response signal to a lambda function. I figured adding a proxy class to use the function pointer was easier and less complicated than some other way (perhaps creating a slot on the ScreenGrabber and using some more event loop stuff...)

Either way, the Screenshot DBus interface works for any DE that implements it. But there's already support for Gnome and KDE.

borgmanJeremy:
Ok awesome. Thanks for the contribution!

Sunderland93:
Why not using xdg-desktop-portal as common interface for it on GNOME and KDE?

nullobsi:
there's no reason; there are just already working implementations for GNOME and KDE. gnome also gives you a popup each time you try to screenshot with the desktop portal (does this happen with the current GNOME implementation?)

borgmanJeremy:
Yes it happens in the current gnome version. I recently tried it.

LiterallyJohnny:
I'm on Sway/Wayland, and it's still not working for me. Is this on the stable build?

Sway version: `1.5.1`

Flameshot version:

```
Flameshot v0.8.5
Compiled with Qt 5.15.1
```

borgmanJeremy:
You need to build the master branch from source.

LiterallyJohnny:
So it's not on the Arch repos yet? I'll go see about checking if we can get it update. Thanks for your work! Flameshot is a very good program.

borgmanJeremy:
On arch you can probably just update the pkgconfig from the aur package flameshot-git

LiterallyJohnny:
Just curious, don't you need to create a new release for that to work? I'm not sure how to do this, the README.md says that it has the latest version, but the package in the official repo hasn't been updated since October.

borgmanJeremy:
Yeah our latest stable release is 0.8.5 and this is what is in the official arch repo's.

However, if you install flameshot-git from the aur, the PKGBUILD looks like it is cloning our master branch (unstable, not official release) https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=flameshot-git

I am not sure when we will have an official release, it will likely not be until 0.9 is ready and that has quite a bit of work remaining.

nullobsi:
Also please note that xdg-desktop-portal-wlr currently has an issue with trying to take multiple screenshots. See emersion/xdg-desktop-portal-wlr#82

vvrein:
Still can't make it work... Here how it looks like:
https://tort.icu/bin/raw/flameshot.mp4
```
Arch Linux
Linux tokk 5.10.11-arch1-1

$ sway --version
sway version 1.5-f267965b (Feb  4 2021, branch 'master')
```

nullobsi:
Seems like a DBus issue. All I can think of is to make sure you have the desktop portal installed.

vvrein:
Yay! It works after properly starting xdg-desktop-portal!

dyercode:
Is there a recommended way to get XDG_CURRENT_DESKTOP set? Sway doesn't and it seems like they'd be unwilling tohttps://github.com/swaywm/sway/issues/4025

nullobsi:
It should be imported into the systemd/dbus activation environment so that when xdg-desktop-portal is started it chooses the right portal. See the wiki page for xdg-desktop-portal-wlr

dmayle:
+subscribe

PR #1307, opened 2021-02-13
fix(appimage): add executable permission for ld-linux-x86-64.so

ZetaoYang:
Fix 
image


mmahmoudian:
This worked for me on Manjaro KDE 👍🏼

borgmanJeremy (approved):

PR #1308, opened 2021-02-13
0.9 release staging

borgmanJeremy:


packaging/debian/changelog
```diff
@@ -1,5 +1,5 @@
-flameshot (0.8.5-1) unstable; urgency=medium
```
hosiet, line 1: I guess this is a typo.

mmahmoudian (commented):
I marked Some typos.

hosiet:
Besides, I'm not sure why the previous v0.8 release note was discarded. If the old release note is discarded, it seems that it can be found nowhere unless someone goes into git history to pull it out.

Since this is changelog-like information, the best way might be adding v0.9 changelog before old v0.8 changelog text.

mmahmoudian:
I agree with @hosiet

docs/ReleaseNotes.md
```diff
@@ -1,63 +1,40 @@
...
+- New global shortcut menu. All actions hotkeys are fully customizable. 
+  
+- Ability to take "symmetric" selections by holding down the Shift key while resizing the selection. 
+
+- The rectangle tool will now round the corners of the rectangles based on the current thickness
+
+- All imgur uploads are now tracked in the "Upload History" menu. This makes it much easier to delete of images off imgur
+or find the upload link later.
+
+- Added "check for new release" feature. This allows users on MacOs / Windows / and AppImages to easily check for updated versions.
```
ZetaoYang, line 30: MacOs -> MacOS

ZetaoYang (commented):
I agree with @mmahmoudian. The release notes for previous versions should be retained.

ZetaoYang:
Besides, `docs/dev/package.md` and `/docs/dev/release.md`. These two documents are outdated and I recommend removing them.

borgmanJeremy:
Thanks everyone, Ill get this feedback cleaned up and then get the release out. Do you think its worth releasing this and RC1 or should I just release it as 0.9 and any issues that come up can be 0.9.x?

mmahmoudian:
Having it as Release Candidate is a good idea, but I don't know Joe many people will be out beta tester. Of course we ourselves can try it out and make sure there is no major issues, as the end user would go for alternative software if they face major bugs. Also this would help us in not getting flooded with similar bug reports.

packaging/debian/changelog
borgmanJeremy: Yeah, it should read 0.9.0-1, Fixed. Thanks

Issue #1287, opened 2021-02-08
Double pathname when opening saved screenshot from notification. 


hcombee:
When I saved a screenshot with Flameshot 0.8.5 on Kubuntu 20.04 and click on the plasmashell notification to open the file I get an error saying Gwenview cannot open the file /home/hans/pictures/screenshots/pictures/screenshots. It repeats the relative location at the end of the configured path. Not sure if this is an error of Flameshot or KDE.

mmahmoudian:
Would you please post here your Flameshot config file:

```sh
cat ~/.config/flameshot/flameshot.ini
```

it should look like this:

> ```
> [General]
> closeAfterScreenshot=false
> disabledTrayIcon=false
> drawColor=#ff0000
> drawThickness=5
> saveAfterCopy=false
> saveAfterCopyPath=/home/mehrad/Pictures
> savePath=/home/mehrad/Pictures
> uiColor=#fa8c00
> ```

hcombee:
Hi mmahmoudian, thanks for the quick reply, my config looks like this:

> [General]
buttons=@Variant(\0\0\0\x7f\0\0\0\vQList<int>\0\0\0\0\x13\0\0\0\0\0\0\0\x1\0\0\0\x2\0\0\0\x3\0\0\0\x4\0\0\0\x5\0\0\0\x6\0\0\0\x12\0\0\0\xf\0\0\0\x13\0\0\0\a\0\0\0\b\0\0\0\t\0\0\0\x10\0\0\0\n\0\0\0\v\0\0\0\f\0\0\0\xe\0\0\0\x11)
disabledTrayIcon=false
drawColor=#ffff00
drawThickness=1
saveAfterCopyPath=/home/hans/Pictures/Screenshots
savePath=Pictures/Screenshots

mmahmoudian:
Can you change your config file and put full path for the `savePath` and restart Flameshot and see if the issue persists.

hcombee:
I changed the path and I even did a reboot but this didn't solve the issue. Weird is that the notification shows a correct path but Gwenview tries to open the "double path". By the way, I installed flameshot using the deb package.

![image](https://user-images.githubusercontent.com/39181810/107574201-bc459f80-6bee-11eb-9352-0f5a9195b409.png)

![image](https://user-images.githubusercontent.com/39181810/107574496-12b2de00-6bef-11eb-98d1-be1b63db0e32.png)

mmahmoudian:
[...]

PR #1318, opened 2021-02-14
Prompt where to save the capture when invoked by KDE shortcut

dandv:
How can I change the KDE shortcut file (or the `Arguments` in the shortcuts) to have Flameshot prompt me where to save the screenshot and with what name, as it does when invoked by clicking it? Is setting the first argument to `''` the correct way to do it?

image

If so, I would like to suggest this be the setting in the KDE shortcuts config file, not the hardcoded `Pictures/Screenshots`, for three reasons:

1. If the user has changed their directory in the Flameshot configuration, then saving to `Pictures/Screenshots` is counter-intuitive
2. The directory might not even exist
3. For consistency with pressing the 💾 icon when invoking Flameshot via clicking its icon, Flameshot should also prompt.

### DIFF
--- a/docs/shortcuts-config/flameshot-shortcuts-kde
+++ b/docs/shortcuts-config/flameshot-shortcuts-kde
@@ -23,7 +23,7 @@ Type=SIMPLE_ACTION_DATA
 ActionsCount=1
 
 [Data_1_1Actions0]
-Arguments='Pictures/Screenshots' 0 0
+Arguments='' 0 0
 Call=graphicCapture
 RemoteApp=org.flameshot.Flameshot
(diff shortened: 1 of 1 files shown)


mmahmoudian:
Your third point is related to https://github.com/flameshot-org/flameshot/issues/1287#issuecomment-777427679

mmahmoudian:
I believe this PR is obsolete considering the new command-based KDE shortcuts. As far as I quickly checked on my phone, we don't use the absolute path anymore in those commands. The file path is:

docs/shortcuts-config/flameshot-shortcuts-kde.khotkeys

PR #1324, opened 2021-02-15
Handle exception when locales are not properly configured

borgmanJeremy:
@mmahmoudian  per your suggestion we now catch the exception generated by the c++ standard library when locales are not properly configured. 

Some documentation for future reference: 
https://en.cppreference.com/w/cpp/locale/locale/locale
-- See specifically where std::runtime_error is thrown when the system locale is improperly configured.

### DIFF
--- a/src/utils/filenamehandler.cpp
+++ b/src/utils/filenamehandler.cpp
@@ -20,12 +20,23 @@
 #include <QDir>
 #include <QStandardPaths>
 #include <ctime>
+#include <exception>
 #include <locale>
+
+#define SPDLOG_ACTIVE_LEVEL SPDLOG_LEVEL_DEBUG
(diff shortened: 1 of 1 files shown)


hosiet:
The code itself looks ok to me, though I believe we may need to find some environments where system locale is broken to actually test that the fix is working.

borgmanJeremy:
I tested it locally by breaking my locales :-)

https://wiki.archlinux.org/index.php/locale
- Overriding system locale per user session

mmahmoudian:
I wonder if the fallback overwrite user's locale or it just use this for flameshot (my confusion is because of the `std::locale::global()`)

borgmanJeremy:
It's "global" for the current process but will not affect anything outside of flameshot.

PR #1331, opened 2021-02-16
Moved SSL check to be windows specific

borgmanJeremy:


hosiet:
This look good to me.

hosiet (approved):

PR #1349, opened 2021-02-18
fix: compiled binary is stuck on win

ZetaoYang:
These files are soft links, which are useless on a Windows NTFS file system. When I compile in Jetbrains clion on local Windows, the compiled binary is stuck and cannot be opened. So I suggest replacing them with normal image files of the same name.

image
image


borgmanJeremy:
@ZetaoYang Seems like this broke all the builds?

ZetaoYang:
@borgmanJeremy I closed this PR and I will raise an issue.

Issue #1351, opened 2021-02-18
Disabling Autostart not possible after resetting settings


kpcee:
Autostart is generally enabled, but disabling it can cause problems if you reset the settings.

To follow up:
* Autostart is enabled
* Click on reset settings
* Deactivate Autostart, now click on another checkbox, Autostart activates itself again, repeat this ... Autostart cannot be deactivated and the file "Flameshot.desktop" is not deleted either. 
* To disable Autostart you have to click Autostart 3 times until the checkbox behaves normally again. 

Compiled Git version with latest changes (commit 66c9f67626b7d339894ca7cea91806178d84ed91) 
Arch Linux

borgmanJeremy:
Thanks for reporting this I can confirm it.

I love when you test flameshot, you find great bugs :)

kpcee:
Wow thanks :)

PR #1357, opened 2021-02-19
fixed glitch with autostart checkboux

borgmanJeremy:


borgmanJeremy:
@thepurple I think you will want to port this to your fork, it was introduced when the autostart default changed. 

closes #1351

PR #1384, opened 2021-02-23
Adjust desktop entry so that launcher uses the gui

tartansandal:
This is needed so the application can be launched normally from Gnome3


ZetaoYang:
I think this change is unreasonable.

borgmanJeremy:
I think so too, perhaps there is a misunderstanding. The current desktop entry works as intended by launching the daemon process.

tartansandal:
Oh, I missed the client server model.

Hmm this makes for a problematic end-user experience for Gnome3. Let me explain. 

I spotted this app by chance browsing the software center. After installing, I get:

from-software

If I click on the launch button, nothing obvious happens, because that simply runs the Exec action from the desktop entry, and that just silently launches the server (perhaps it could send a desktop notification).  A similar issue occurs when trying to launch the app from Activities, which is what Gnome3 users use instead of a tray:

from-activities

Many users would think, "Pfft! Just another broken app!" and uninstall it.

Having the desktop entry just launch the server is counter to expectations. I can launch the default screen `gnome-screenshot` program from Activities because of the Exec lines in the desktop entry. Here's the abbreviated content of the `gnome-screenshot` desktop file:

```
[Desktop Entry]
Name=Screenshot
Comment=Save images of your screen or individual windows
Terminal=false
Type=Application
StartupNotify=true
Actions=screen-shot;window-shot;
DBusActivatable=true
Exec=gnome-screenshot --interactive

[Desktop Action screen-shot]
Name=Take a Screenshot of the Whole Screen
Exec=gnome-screenshot

[Desktop Action window-shot]
Name=Take a Screenshot of the Current Window
Exec=gnome-screenshot -w
```
Maybe something like that could be envisioned?

This is still problematic since `flameshot` needs a server. Perhaps we need a wrapper script to launch the server if its not already, then launch the gui? Of even make the gui version spawn a detached server if there is not one running already.

Really love the app, by the way, and really want it to be accessible to more Fedora users :smile:

borgmanJeremy:
I think the root cause of the issue is Gnome removed the tray icon. Flameshot centers around launching tasks from the tray icon (or the Ptr Screen button on the keyboard). 

Have you tried installing a tray icon extension as explained in the README? https://github.com/flameshot-org/flameshot#considerations

tartansandal:
> I think the root cause of the issue is Gnome removed the tray icon. Flameshot centers around launching tasks from the tray icon (or the Ptr Screen button on the keyboard).

Yeah this was a big and deep design decision by Gnome which is unlikely to change and is kinda out of scope if we want to support Gnome.  The extensions to try to reintegrate the tray are flaky and currently don't work very well.  Maybe this is something for whoever maintains the Fedora package. Maybe making a separate package `gnome-flameshot` that supports launching in the Gnome fashion.

tartansandal:
Actually nix this. It looks like Fedora are maintaining there own Desktop file for this and I should be pestering them about it instead :smile:

PR #1396, opened 2021-02-25
removed loop that occured when screenshot failed to save.

### LINKED ISSUE #1364: Serious bug if path not present
I have found an extremely unsightly bug that, in the worst case, can cause Flameshot to run amok and possibly cripple the system. The bug consists of 2 parts:

Part 1:
First, you have to set a path that doesn't exist or that Flameshot can't save to. Since the path ~/Pictures is set at startup, this is easy to do on systems with languages other than English. 

Now select Open Launcher and choose "Full Screen" and "Take new screenshot". Next a messagebox pops up saying that it is not possible to save. If you click on Ok, the message appears again → endless loop! 

Part 2:
If the user now tries to quit Flameshot via Exit from tray menu, the fun really starts and Flameshot runs amok. The message box opens and closes automatically in an endless loop and only closing it via the task manager can prevent worse.

Git version (commit 16fd3b2915829ccb550a023d7b829b7a46beebde)
Arch Linux

borgmanJeremy:
This fixes the issue where if ~/Pictures does not exist, an infinite loop occurs. I cannot figure out why this loop was ever present. 

Closes #1364

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # common
 
 .idea
-
+.cache
 # C++ objects and libs
 
 *.slo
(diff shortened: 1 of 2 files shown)


hosiet:
I think the code was originally meant to repeat the save procedure (if it fails) until the file save is successful. With this pull request, the change seems to be ok to me. I haven't tested it yet though.

PR #1398, opened 2021-02-26
NC/fixes 25-02-2021

thepurple:
- fix - File name pattern editor 'Clear' and 'Reset' behavior (configuration window)
- fix - Reset button doesn't reset configurations to the initial values
- fix - MacOS - Image is not copied to the clipboard when the 'Use JPG format for clipboard' checkbox is set
- fix - 'Launch at startup feature' doesn't work after uncheck/check in the Configuration window

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -386,6 +386,10 @@ void GeneralConf::initUseJpgForClipboard()
       tr("Use JPG format for clipboard (PNG default)"));
     m_layout->addWidget(m_useJpgForClipboard);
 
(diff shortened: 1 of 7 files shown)


borgmanJeremy:
Are you sure the jpeg stuff does not work on MacOS? It's possible the programs you are pasting into simply dont support decoding the jpeg header. Can you try pasting it into something where you can inspect the contents like thunderbird?

thepurple:
I haven't much time to investigate the issue and find a root cause, but it doesn't work at least on two MacBooks (QAs and mine) with BigSur.
So I just disabled it for now.

PR #1428, opened 2021-03-04
Small fixes after merge review

borgmanJeremy:


hosiet:
I took another look and the problem seems to be that there is `data/img/material/black/delete.svg` in v0.9.0 release but no `data/img/material/white/delete.svg`.

In fact we indeed did not use `data/img/material/white/delete.svg`: looking through all source files, the only usage is:

https://github.com/flameshot-org/flameshot/blob/5690afbd4be857e20e0ff6f100c890a7250cccfe/src/widgets/historywidget.cpp#L153

This explicitly used the black variant.

Still the problem is that we don't know where the updated icons come from and under which license were they released.

Issue #1356, opened 2021-02-18
Doesn't work on sway


fabifont:
**Flameshot version**
Flameshot v0.8.5
Compiled with Qt 5.15.1

**Describe the bug**
Instant crash. This is the log:
```
qt.qpa.plugin: Could not find the Qt platform plugin "wayland" in ""
qt.qpa.xcb: could not connect to display :0
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, xcb.

Aborted (core dumped)
```

**To Reproduce**
Type `flameshot` in your terminal.

**Expected behavior**
Launch `flameshot`.

**System Information**
Linux arch 5.10.16-arch1-1

borgmanJeremy:
What windows manager are you using

fabifont:
> What windows manager are you using

Sway

borgmanJeremy:
Do you have qt5-wayland installed? 

We use xcb on the backend so I suspect you are missing something to enable that.

fabifont:
> Do you have qt5-wayland installed?
> 
> We use xcb on the backend so I suspect you are missing something to enable that.

Now yes, flameshot doesn't crash anymore but clicking on the icon it does nothing. The log says: `qt.qpa.wayland: Wayland does not support QWindow::requestActivate()`

borgmanJeremy:
Checkout this thread: https://github.com/flameshot-org/flameshot/pull/1272

Specifically:

>Yay! It works after properly starting xdg-desktop-portal!

tinywrkb:
It works for me with 0.9.0 but only once and after that it fails.  

I'm using the Flatpak from here https://github.com/flathub/org.flameshot.Flameshot but building it myself as Flathub is not updated.  

Flameshot doesn't do a good job handling `XDG_CURRENT_DESKTOP` it should split on `:` and try the first sub-string instead of the current randomly chosen order by the developer, [this is wrong](https://github.com/flameshot-org/flameshot/blob/5b0bb501dc513f16e93a6d72e06a222a2da6c404/src/utils/desktopinfo.cpp#L29).  
I don't remember where it's documented but the split in on colon.  
[...]

PR #1431, opened 2021-03-05
fix XDG desktop check

nullobsi:
This should fix the issue mentioned by @tinywrkb in #1356.

Properly parse the XDG_CURRENT_DESKTOP variable to handle colon seperated lists. I've checked around and while there is no specification, [xdg-desktop-portal](https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/portal-impl.c#L189) splits the variable by colons.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -38,6 +38,7 @@ int waitAfterConnecting(int delay, QCoreApplication& app)
     return app.exec();
 }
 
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
Can you fix the clang-format check? If not i can fix it in a different PR.

PR #1432, opened 2021-03-05
improve Sway support

nullobsi:
this PR improves sway support by:
- using a unique token (works with vanilla xdg-desktop-portal-wlr)
- subscribing to signal before calling screenshot (to prevent race conditions)


borgmanJeremy:
Looks like the version of Qt packed on ubuntu 18.04 does not support those Qt functions.

nullobsi:
alright, i've made it use another function which should work for qt < 5.11

PR #1456, opened 2021-03-12
Spanish translation

GuxMartin:


borgmanJeremy:
I think you may have corrupted a tag which is why the continous integration is failing. You could try adjusting the translation using weblate: https://hosted.weblate.org/projects/flameshot/flameshot/

PR #1472, opened 2021-03-14
fixed invalid path when tilde is used as a param for key shortcut

### LINKED ISSUE #1465: Using "-p ~" gives "Invalid path, it must be a real path in the system"
**Flameshot version**

0.8.5

**Describe the bug**

I've set a keyboard shortcut for `flameshot gui -p ~`

**To Reproduce**

1. set a keyboard shortcut for `flameshot gui -p ~` (as mentioned in docs https://flameshot.org/command-line-options/ )
2. press shortcut
3. flameshot says "Invalid path, it must be a real path in the system"

If changing ~ to `/home/user`, it works.

**Expected behavior**

Have flameshot start and save to /home/user

**System Information**

Fedora 34 beta, gnome 4, bash

TomaszPluta:
Fixes #1465 "Using "-p ~" gives "Invalid path, it must be a real path in the system"

Why problem occured?  The "~" as path argument is replaced (by home dir) performed by shell - not by language, not by Qt. 
My propose to fix? There is need to add checking if tilda is used as a path (and replace it with home path if needed).
Tested on: Ubuntu 20.10, 5.8.0-44-generic

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -206,7 +206,11 @@ int main(int argc, char* argv[])
     const QString pathErr =
       QObject::tr("Invalid path, it must be a real path in the system");
     auto pathChecker = [pathErr](const QString& pathValue) -> bool {
-        bool res = QDir(pathValue).exists();
+    	QString pathToCheck = pathValue;
+    	if ((pathValue == "~") || (pathValue.startsWith("~/"))) {
+            pathToCheck.replace (0, 1, QDir::homePath());
+    	}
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
I don't really want flameshot to be required to do shell expansion. My concern is that the tilde is one small part of shell expansion. Is there a way to force the shell to expand the argument before passing to flameshot?

TomaszPluta:
Ok, so I don't  have other ideas

PR #1479, opened 2021-03-15
README: add MacPorts installation instructions

herbygillot:


herbygillot:
Thank you @borgmanJeremy

borgmanJeremy:
Thanks for submitting the update! Sorry for the delay, it took me a while to build the ports on my mac.

herbygillot:
> Thanks for submitting the update! Sorry for the delay, it took me a while to build the ports on my mac.

Are you running macOS Catalina (10.15)?  Right now builds on Catalina are taking longer than they should because each package and dependency is being built from source (Cmake, Qt5's dependencies, OpenSSL, etc.).

Source building is happening because the MacPorts buildbot for 10.15 has recently been offline.

It's back now, and slowly binary packages are being made available for 10.15 as the buildbot works through the backlog.
Once binary packages are fully available, installation should only take a few minutes.**

** = also depending on the licenses of some particular packages (some licenses prohibit binary distribution, and so must still be built from source)

PR #1513, opened 2021-03-27
Change .desktop file to support Plasma Wayland

### LINKED ISSUE #1380: Does not work when using Plasma+Wayland
**Versions**
Flameshot: 5b0bb501dc513f16e93a6d72e06a222a2da6c404
Kwin/Plasma: v5.21
Protocol: Wayland

**Describe the bug**
Flameshot does not take a screenshot. I tried tracking this down with the hopes of creating a patch, but I'm SOL due to my limited knowledge in the Linux desktop space. These are my findings: 

* KWin requires you to use the D-Bus. 
* KWin enforces security by ensuring you have the `X-KDE-DBUS-Restricted-Interfaces` key with the value `org.kde.kwin.Screenshot`.
* KWin uses something called KApplicationTrader to find the desktop file of the process and check if the aforementioned key exists. It compare the `Exec` key in the desktop files and the executable location obtained from procfs to do so.
* Flameshot does not specify the full path to the binary in it's desktop file, unlike Spectacle. 
* Flameshot sets the `X-KDE-DBUS-Restricted-Interfaces` key to `org_kde_kwin_effect-screenshot` instead of `org.kde.kwin.Screenshot`.

I tried setting the `Exec` key full path to the binary as well as changing the `X-KDE-DBUS-Restricted-Interfaces` key to the intended value. However, Flameshot still does not work. I hope someone can fix this issue or help track this down further.

Thanks and sorry if a similar issue already exists.

abhijeetviswa:
Close #1380

### DIFF
--- a/data/desktopEntry/package/org.flameshot.Flameshot.desktop
+++ b/data/desktopEntry/package/org.flameshot.Flameshot.desktop
@@ -37,7 +37,7 @@ Keywords[nl_NL]=flameshot;schermfoto;screenshot;schermafdruk;vastleggen;schermop
 Keywords[es]=flameshot;screenshot;capture;shutter;captura;
 Keywords[de]=flameshot;screenshot;Bildschirmfoto;Aufnahme;
 Keywords[pt_BR]=flameshot;screenshot;captura de tela;captura;shutter;
-Exec=flameshot
+Exec=/usr/bin/flameshot
 Icon=org.flameshot.Flameshot
 Terminal=false
 Type=Application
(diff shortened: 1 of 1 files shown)


abhijeetviswa:
Well, seems like a path cannot be provided to the Exec key when building an appimage.

borgmanJeremy:
@abhijeetviswa you can edit the github action file to use sed or something to replace the string. As long as the app image still works on KDE wayland, I think this is an okay fix.

cossonleo:
It can not be drag when pin the screenshot.

borgmanJeremy:
Replaced by #1603

Issue #1399, opened 2021-02-26
Issues with Save as


kpcee:
If Flameshot is set to ask for save location and name each time, the following minor bugs exist:
* It cannot be saved as BMP, after selecting BMP it is saved as PNG. (By the way, instead of BMP I would offer the modern WebP as an option, it is afaik the most modern format what Qt currently supports out of the box).

* Cancel save as dialog: Flamesot shows the desktop notification that the file is saved.

* Cancel save as dialog: Flameshot will change the save path in the preferences to ".png".

Latest Git (commit 5705e45e32387b3c57514d8150b0a22888d53ddf )
Arch Linux, Xfce

MrNilDev:
Try giving us some more information or we can not help you with that 
**Flameshot version**
**To Reproduce**
**Expected behavior**
and please keep them separately.

kpcee:
I have written which version I use, even the commit was given. It's all about the save as dialog when it's enabled and is it really necessary to write an "Expected behavior" when it's logical what is meant or what is going wrong?
  
If saving as BMP is not possible, even though there is the option, the "Expected behavior" should be that Flameshot saves as BMP. Likewise, there should be no desktop notification that saving was done when you cancel the save operation, and also the save path in the settings should not be changed to ".png" when you cancel the save operation.

hosiet:
I can confirm that:

* Saving as BMP file is working well on Microsoft Windows (with v0.9.0 release).
* Saving as BMP file is indeed not working on Linux platform **if the user does not explicitly provide the .bmp suffix in the filename**:

1.  If the user choose bmp in the saving dialog and use `my_filename` as file name, the saved file would be of PNG format.
2. If the user choose bmp in the saveing dialog and use `my_filename.bmp` as file name, the saved file would be of BMP format.


This is considered a bug.

TomaszPluta:
Would it be ok to if after choosing a bmp in file save dialog ".bmp" sufix will be added automatically?

mmahmoudian:
@TomaszPluta are you planning to make a PR? If so, it would be nice to have the feature also configurable (auto-extension can be enabled/disabled in config)

TomaszPluta:
I have tested this issue, and I can confirm it is easy to reproduce. Fixing in that way shouldn't take a lot of time, so I could try.
Provide it as config option is good idea and only one thing need clarification - such option should be default enabled or disabled?

mmahmoudian:
[...]

PR #1521, opened 2021-03-30
Auto add file extension when save - issue #1399

TomaszPluta:
Added auto add the file extension in "Save As" file dialog based on selected filter. The method " QFileDialog::getSaveFileName" in linux envrionment  doesn't give this possibility, so here is added  "ShowSaveFileDialog" class. 

This class performs what was needed in issue #1399:
_- If a graphic format would be selected, the file name is completely ignored and the extension of the selected graphic format would be appended to the entered file name and will be saved in this format.
- If the filename is empty, the saving process would be aborted.
- If the file exists, it would be best to ask whether it should be overwritten, if not call the Save As dialog again.
- It would only save the selected file extension entry in the Save As dialog and restore this setting the next time._

Please review and give remarks

Based on: https://stackoverflow.com/questions/9822177/is-there-a-way-to-automatically-add-extensions-to-a-file-using-qfiledialog-on-li

### DIFF
--- a/src/utils/screenshotsaver.cpp
+++ b/src/utils/screenshotsaver.cpp
@@ -115,23 +115,27 @@ QString ShowSaveFileDialog(QWidget *parent,
     dialog.setWindowModality(Qt::WindowModal);
   }
 
+
   QRegExp filter_regex(QLatin1String("(?:^\\*\\.(?!.*\\()|\\(\\*\\.)(\\w+)"));
   QStringList filters = filter.split(QLatin1String(";;"));
 
   dialog.setAcceptMode(QFileDialog::AcceptSave);
 
-
+  dialog.selectNameFilter(ConfigHandler().getSaveAsFileExtension());
   if (dialog.exec() == QDialog::Accepted) {
+
+	ConfigHandler().setSaveAsFileExtension(dialog.selectedNameFilter());
 
     QString file_name = dialog.selectedFiles().first();
     QFileInfo info(file_name);
 
-   if ((dialog.selectedNameFilter() == "By extension (default: *.png)")){//if no sufix change to default, otherwise leave it as it is
+
+   if ((dialog.selectedNameFilter() == "By extension (default: *.png)")){//change to default if no suffix given, otherwise leave it as it is
 	   if (info.suffix().isEmpty()){
 		   file_name = info.completeBaseName() + QLatin1String(".") + "png";;
 	   }
    }
-   else	if ( !dialog.selectedNameFilter().isEmpty()) {//if selected sufix from menu is not an empty entry
+   else	if ( !dialog.selectedNameFilter().isEmpty()) {//if selected suffix from menu is not an empty entry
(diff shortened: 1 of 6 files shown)


borgmanJeremy:
>Solution is based on some discussion found in internet (it is know issue in QT), so code may look a little bit complicated, but i don't see a simpler way. If you have an idea to simplify a code - please add remars in review.

Can you link to the discussions?

ZetaoYang:
Tested this PR from CI build, Win64,
image
When cancel "Save" window,
image

borgmanJeremy:
@TomaszPluta Can you add a reference to the stack overflow link to the comments? It's required per the stack overflow license: https://stackoverflow.com/help/licensing

src/utils/screenshotsaver.cpp
```diff
@@ -99,6 +99,55 @@ bool ScreenshotSaver::saveToFilesystem(const QPixmap& capture,
...
+{
+#if defined(Q_WS_WIN) || defined(Q_WS_MAC)
+    return QFileDialog::getSaveFileName(parent, title, directory, filter);
+#else
+    QFileDialog dialog(parent, title, directory, filter);
+    if (parent) {
+        dialog.setWindowModality(Qt::WindowModal);
+    }
+
+    QRegExp filter_regex(QLatin1String("(?:^\\*\\.(?!.*\\()|\\(\\*\\.)(\\w+)"));
```
borgmanJeremy, line 115: Could you define this regex as a variable name, move to a function, or add a comment to explain what it does? I'm not a regex wizard :)

src/utils/screenshotsaver.cpp
```diff
@@ -99,6 +99,55 @@ bool ScreenshotSaver::saveToFilesystem(const QPixmap& capture,
...
+    dialog.setAcceptMode(QFileDialog::AcceptSave);
+
+    dialog.selectNameFilter(ConfigHandler().getSaveAsFileExtension());
+    if (dialog.exec() == QDialog::Accepted) {
+
+        ConfigHandler().setSaveAsFileExtension(dialog.selectedNameFilter());
+        QString file_name = dialog.selectedFiles().first();
+        QFileInfo info(file_name);
+
+        if ((dialog.selectedNameFilter() == "By extension [default: *.png]")) {
```
borgmanJeremy, line 127: "By extension [default: *.png]" should be defined as a variable so when it is re-used on line 174 there is no chance of it only changing in once place causing a future bug.

borgmanJeremy (changes requested):

src/utils/screenshotsaver.cpp
TomaszPluta: I changed the regex to some QString operations (extracting extension from chosen filter). IMHO is easier to understand it now.

src/utils/screenshotsaver.cpp
TomaszPluta: Yeah, good point. I've moved it to const variables

borgmanJeremy (approved):

borgmanJeremy:
Awesome, thanks for making the changes! Could you run clang-format and then I'll get it merged?

PR #1539, opened 2021-04-07
Namecheap - Basic objects editing (move, remove, change color, change thickness)

thepurple:
Unfortunately one more big PR, but I was trying to deliver a minimal working feature.

Current object editing contains:
- move objects
- remove object
- edit color
- edit thickness

There are still some issues and I will send a number of improvements soon (edit color for CircleCount, performance and possible some fixes for unknown issues yet). But I think it worth to be merged now. I don't want to get a bigger code base for merging, even this one was a headache.


thepurple:
@borgmanJeremy Unfortunately no documented design, just in my head.
If it would be critical I'll try to find some time to do it, but not this week.
Here is in shot:
- It work with layers, each object is a layer, which are stored in CaptureWidget::m_captureToolObjects.
- Searching is implemented in CaptureToolObjects. For searching it draws one by one on hidden transparent layers and checking for a non-transparent pixel in two steps (first with radius 3, second with radius 5). For text objects it uses a bigger radius.
- Completely new approach with undo/redo. Now each step saves current objects state (CaptureToolObjects with the list of layers), no more backup pixmaps. I've stated to implement caching (you can see some code about it) but I didn't finish it and it is not used now. Will finish it later, on the next iteration.
- Completely new approach with CircleCounter tool, now it's re-calculated on drawing and no more skipped items.

thepurple:
Change log:
- Base objects editing:
  - Move objects
  - Remove any object
  - Edit thickness for any selected object
  - Edit color for any selected object
- Completely new undo/redo stack
- CircleCount - make circles more contrast and mouse preview opaque
- fix - Shapes are displayed as semi-empty at the very beginning of drawing if the line is thick
- fix - Rename "Reset" to "Restore" in file settings for restoring saved file name rules in the configuration widget
- fix - Increase and Decrease Size tool icon colors are not changed (dark/light) when changing main tool background color
- fix - Latest Uploads max size value doesn't restore after reset/import settings
- fix - Rectangle preview and final objects are not the same
- fix - Make rectangle tool with rounded borders (not cut), Make selection tool without rounded edges
- fix - Circle Counter Bubble numeration can skip some items if CPU is high loaded at that moment
- fix - No bubble border for Circle Count Tool

src/tools/circle/circletool.cpp
```diff
@@ -36,36 +36,19 @@ QString CircleTool::description() const
 
 CaptureTool* CircleTool::copy(QObject* parent)
 {
-    return new CircleTool(parent);
+    auto* tool = new CircleTool(parent);
```
borgmanJeremy, line 39: Won't this cause a slow memory leak? I don't see where either the original or new copy is destructed.

src/tools/abstracttwopointtool.cpp
```diff
@@ -36,6 +36,19 @@ AbstractTwoPointTool::AbstractTwoPointTool(QObject* parent)
   , m_padding(0)
 {}
 
+void AbstractTwoPointTool::copyParams(const AbstractTwoPointTool* from,
```
borgmanJeremy, line 39: Might be cleaner to replace with assignment operator? Just a suggestion, not important.

src/tools/copy/copytool.cpp
```diff
@@ -16,7 +16,7 @@ bool CopyTool::closeOnButtonPressed() const
 
 QIcon CopyTool::icon(const QColor& background, bool inEditor) const
 {
-    Q_UNUSED(inEditor);
+    Q_UNUSED(inEditor)
```
borgmanJeremy, line 19: I assume you removed all the macro semicolons because an IDE suggested it? Just curious which IDE suggested it and why?

src/tools/pixelate/pixelatetool.cpp
```diff
@@ -35,27 +35,22 @@ QString PixelateTool::description() const
...
+    QPoint p1 = points().second;
     QRect selection = QRect(p0, p1).normalized();
     auto pixelRatio = pixmap.devicePixelRatio();
     QRect selectionScaled =
       QRect(p0 * pixelRatio, p1 * pixelRatio).normalized();
 
-    // If thickness is less than 1, use old blur process
-    if (m_thickness <= 1) {
+    // If thickness is 0, use old blur process
+    if (thickness() <= 0) {
```
borgmanJeremy, line 53: I think this is a bug because thickness() never returns a value of 0, therefore the old blur cannot be used.

src/widgets/capture/capturetoolobjects.cpp
```diff
@@ -0,0 +1,131 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// SPDX-FileCopyrightText: 2017-2019 Alejandro Sirgo Rica & Contributors
```
borgmanJeremy, line 2: You can list yourself as the copyright author since this is authored by you

src/widgets/capture/capturetoolobjects.cpp
```diff
@@ -0,0 +1,131 @@
...
+int CaptureToolObjects::find(const QPoint& pos, const QSize& captureSize)
+{
+    if (m_captureToolObjects.empty()) {
+        return -1;
+    }
+    QPixmap pixmap(captureSize);
+    pixmap.fill(Qt::transparent);
+    QPainter painter(&pixmap);
+    // first attempt to find at exact position
+    int radius = 3;
```
borgmanJeremy, line 52: I see radius of 3 and 5 are used throughout this file. Can we define this as a constant in the header?

borgmanJeremy (changes requested):
Overall great work! Really appreciate you putting in the work to get this done. I really like the interaction between the utility side bar and the ability to move. IE hit a shape name in the history, and it draws the selection box around it.

I am sure the community will quickly ask  to add "handles" so the shapes can be resized as well as moved :)

One question I have: Is the purpose of the caching to allow faster selection of objects when the need to be moved?

src/tools/copy/copytool.cpp
thepurple: I did it by recommendation of IDE CLion, decrease number of warnings, it is redundant.

src/tools/pixelate/pixelatetool.cpp
thepurple: Thickness can be zero, but I'll check it. Possibly I made a new error with thickness somewhere.

src/tools/abstracttwopointtool.cpp
thepurple: I don't exactly remember why did I do it in this way. I had some difficulties with "operator=" and this was was easier. I used it in another similar case, but not here. If it's critical, I'll refactor it.

src/tools/circle/circletool.cpp
thepurple: I'll check it tomorrow or the day after tomorrow. I bit tired after the long road today.

src/widgets/capture/capturetoolobjects.cpp
thepurple: Sure good point, I just forget to refactor this part of code after the debugging.

thepurple:
>Overall great work! Really appreciate you putting in the work to get this done. I really like the interaction between the utility side bar and the ability to move. IE hit a shape name in the history, and it draws the selection box around it.
Thanks

> I am sure the community will quickly ask to add "handles" so the shapes can be resized as well as moved :)
No doubts, I hope that I will get time for implementing this "further" requests :)

> One question I have: Is the purpose of the caching to allow faster selection of objects when the need to be moved?
Yes, because if you'll draw 30 or more objects and will try to select the first created object you may see some delay (on some slow computer).
Also still need to do some optimization for redraw image on objects moving.

src/tools/circle/circletool.cpp
thepurple: In general nothing changed here, 
previous code:

```
CaptureTool* CircleTool::copy(QObject* parent)
{
    return new CirceTool(parent);
}
```

Now we have:
```
CaptureTool* CircleTool::copy(QObject* parent)
{
    auto* tool = new CircleTool(parent);
    copyParams(this, tool);
    return tool;
}
```

The code is still returning the same pointer.
I even did a check to be sure. I have placed a huge 1Gb array into the CircleTool and each new object takes it and I can see it, after closing CaptureWidget memory utilisation by Flameshot backs to 24.3Mb.

src/tools/abstracttwopointtool.cpp
thepurple: I'll take a look at it and will do refactor later, I have noted it to myself.

src/widgets/capture/capturetoolobjects.cpp
thepurple: Fixed

mmahmoudian:
@thepurple @borgmanJeremy  While testing I accidentally managed to create segfault:

Peek 2021-04-20 14-26


basically there are three bugs in the screencapture:

1. when running the Flameshot with `gui` argument, it works, but when it is initiated without `gui` and then tray icon is clicked, pressing the move button causes few seconds of freeze and segfault
2. when running the `gui` argument and clicking on move button, it is not possible to deactivate the move mode with clicking on it again. In the video I click again on the move button but it does not get deactivated. The re-clicking for deactivation works for other tools that I tested (didn't test everything though)
3. when clicking on the move button, the button background color does not change unlike other buttons.

thepurple:
I can reproduce it too, thanks.
Will fix it in day or two.

mmahmoudian:
@thepurple  I also realized that when invoked with the `gui` argument, by default the move mode (or what ever name it has) is active which prevents me from selecting the object and modifying it:

Peek 2021-04-20 14-36

thepurple:
@mmahmoudian please check this one https://github.com/flameshot-org/flameshot/pull/1569
I've fixed segfault, it was an old legacy which wasn't critical before. Two other points didn't check yet.
This PR also has two other small fixes:
- fix - do not set black color for the active tool on Custom Color
- fix - Add antialiasing for lines, arrows, etc.

PR #1556, opened 2021-04-13
Added FAQ

borgmanJeremy:
@mmahmoudian Feel free to edit this if you have more ideas for the FAQ or add more later. I feel like we spend a bunch of time answering the same questions so we might as well make the answers easier to find.


mmahmoudian:
I suggest moving this FAQ to the website as there is already a dedicated page for it: https://flameshot.org/faq/

(I can do the move if you like)

borgmanJeremy:
good point, I moved the PR there. Once you approve it there I'll update the links in this PR.

gur111 (approved):

PR #1565, opened 2021-04-18
Windowed capture mode

karliss:
Properly handling multiple screens with varying scaling factors is challenging. I saw few attempts fixing it, but I am not sure if a sufficiently robust solution which works in all configurations on all the desktop operating systems will be reached anytime soon. That doesn't mean people shouldn't try.

I am suggesting an alternative simpler approach which could be user selected to make flameshot somewhat somewhat usable in setups where it couldn't be used otherwise. Instead of trying to show captured image on all the screens in fullscreen mode display it in a window with ability to move around and maybe zoom.

Having this mode might even help resolving some of the issues related to mixed display setups incrementally without solving all of them at the same time.

The code itself isn't ready yet, I wanted to discuss the approach first before spending more work on the code.

What's currently there:
* Added option for choosing one of 3 modes: all screen fullscreen, current screen fullscreen, maximized window. The first one is not available on macOS.
* CaptureWidget opens in selected mode
* In window mode view can be dragged around using middle click
* tools and preview are drawn taking current view offset into account

What's missing, needs to be done, known issues:

* Testing on all the platforms and various display configs.
* Cleanup the code to comply with this projects coding style and other contribution requirements
* in current screen fullscreen mode capture image from corresponding screen only (partially works)
* Cursor sometimes disapears in color picker
* Ctrl->C shortcut doesn't work. Copying to clipboard using tool button works.

Some other problems to consider related to multiple screens and their scaling, possibly for separate PRs:

**Single screen fullscreen mode** Similar to "all screen fullscreen" and "windowed" modes have third mode which captures image from single screen and displays it fullscreen on the same window. It preserves the inplace screen annotation functionality while avoiding challenging cases. From what I understand something similar Is already done for macOS. Having it as separate mode which can be selected on tested on any OS instead macOS specific fullscreen handling, would help some of Linux users. I assume that in many cases when you are selecting part of single screen anyway so doesn't need to be across mutliple screens.

**Scaling factor of resulting image.** It seems that currently flameshot currently creates single image consisting of all the screens scaled so that there is no information loss for highest DPI screen. This is fine when you are creating screenshot across multple screens, but if the selected rectangle is inside lower DPI screen it makes it unnecessarily upscaled.  That could be handled by advanced user option choosing betweeen following modes or choosing automatically based on selection rectangle.
* scaling aware mode -> current behavior, matches how user sees things in real world (or at least desktop config), capturing parts of lower DPI screens result in unwanted upscaling
* no scaling -> glue together images at the resolution of each display,  no scaling artifacts but captures across multiple screens  might not fit together. If QT doesn't provide information about such layout it might be nontrivial to decide how to layout all the screens.
* automatic mode based on selection rectangle -> If selection rectangle is across multiple screens behave like in "scaling aware mode", if rectangle is on single screen use take image from that screen thus avoiding upscaling issue. This approach will  require storing separate QPixmap for screencapture of each individual display in addition to current combined one since downscaling image which has been upscaled by fraction coefficient would cause some degradation.

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -5,6 +5,7 @@
 #include "src/core/controller.h"
 #include "src/utils/confighandler.h"
 #include <QCheckBox>
+#include <QComboBox>
 #include <QFile>
 #include <QFileDialog>
 #include <QGroupBox>
@@ -41,6 +42,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initCopyAndCloseAfterUpload();
     initCopyPathAfterSave();
     initUseJpgForClipboard();
+    initWindowMode();
     initSaveAfterCopy();
     inituploadHistoryMax();
     initUndoLimit();
@@ -68,6 +70,13 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
     m_checkForUpdates->setChecked(config.checkForUpdates());
     m_showStartupLaunchMessage->setChecked(config.showStartupLaunchMessage());
     m_screenshotPathFixedCheck->setChecked(config.savePathFixed());
+    auto configWindowMode = config.windowMode();
+    for (int i = 0; i < m_windowMode->count(); i++) {
+        if (m_windowMode->itemData(i) == configWindowMode) {
+            m_windowMode->setCurrentIndex(i);
+            break;
+        }
+    }
     m_uploadHistoryMax->setValue(config.uploadHistoryMax());
     m_undoLimit->setValue(config.undoLimit());
 
@@ -461,6 +470,44 @@ void GeneralConf::initUseJpgForClipboard()
             &QCheckBox::clicked,
             this,
             &GeneralConf::useJpgForClipboardChanged);
+}
+
+void GeneralConf::initWindowMode()
+{
+    m_windowMode = new QComboBox(this);
+    m_layout->addWidget(m_windowMode);
+    struct
+    {
+        QString label;
+        CaptureConfig::CaptureWindowMode mode;
+    } modes[] = {
+#if !defined(Q_OS_MACOS)
+        { tr("All screens"), CaptureConfig::CaptureWindowMode::FullScreenAll },
+#endif
+        { tr("Current screen"),
+          CaptureConfig::CaptureWindowMode::FullScreenCurrent },
+        { tr("Maximized window"),
+          CaptureConfig::CaptureWindowMode::MaximizeWindow },
+        { tr("Debug window"),
+          CaptureConfig::CaptureWindowMode::DebugNonFullScreen },
+    };
+    for (auto& item : modes) {
+        m_windowMode->addItem(item.label, item.mode);
+    }
+    connect(m_windowMode,
+            QOverload<int>::of(&QComboBox::currentIndexChanged),
+            this,
+            &GeneralConf::windowModeChanged);
+}
+
+void GeneralConf::windowModeChanged()
+{
+    auto selectedData = m_windowMode->currentData();
+    if (selectedData.canConvert<CaptureConfig::CaptureWindowMode>()) {
+        auto selectedMode =
+          selectedData.value<CaptureConfig::CaptureWindowMode>();
+        ConfigHandler().setWindowMode(selectedMode);
+    }
 }
 
 void GeneralConf::saveAfterCopyChanged(bool checked)
--- a/src/config/generalconf.h
+++ b/src/config/generalconf.h
@@ -12,6 +12,7 @@ class QPushButton;
 class QLabel;
 class QLineEdit;
 class QSpinBox;
+class QComboBox;
 
 class GeneralConf : public QWidget
 {
@@ -39,6 +40,7 @@ private slots:
     void resetConfiguration();
     void togglePathFixed();
     void useJpgForClipboardChanged(bool checked);
+    void windowModeChanged();
 
 private:
     const QString chooseFolder(const QString currentPath = "");
@@ -59,6 +61,7 @@ private:
     void initSaveAfterCopy();
     void initCopyPathAfterSave();
     void initUseJpgForClipboard();
+    void initWindowMode();
 
     void _updateComponents(bool allowEmptySavePath);
 
@@ -84,6 +87,7 @@ private:
     QCheckBox* m_screenshotPathFixedCheck;
     QCheckBox* m_historyConfirmationToDelete;
     QCheckBox* m_useJpgForClipboard;
+    QComboBox* m_windowMode;
     QSpinBox* m_uploadHistoryMax;
     QSpinBox* m_undoLimit;
 };
--- a/src/core/controller.cpp
+++ b/src/core/controller.cpp
@@ -285,7 +285,7 @@ void Controller::startVisualCapture(const uint id,
             return;
         }
 
-        m_captureWindow = new CaptureWidget(id, forcedSavePath);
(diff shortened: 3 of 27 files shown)


borgmanJeremy:
I'm open to adding this because I honestly don't have a good line of sight to fixing mixed DPI any other way. I am having issues using this on a tiling window manager though. I think you did a good job of capturing "missing items" that would need to be added before merging.

Thanks!

karliss:
Found a little bit more time to work on this.  Wanted to write down some of the current design choices. One more reminder that as indicated by draft state, this is still work in progress and a lot of stuff is in broken state, possibly more than what's listed in the initial PR description.

Coordinate systems:
* screen geometry, global mouse position, ... - mostly relevant during capture process and setting up multiscreen fullscreen window
* capture coordinates - pixel coordinates in the captured image
* view coordinates - positions relative to capture widget, mouse events from Qt report these, relevant for UI interactions

Tools and `CaptureTool` **interface methods operate within capture coordinates** and don't know anything about what parts the view is currently showing. Might need some exceptions for special tools, but should be mostly fine.

UI structure
- CaptureWidget: `QScrollArea` 
  - viewport - part of QScrollArea structure
    - QScrollArea::widget() - parent of all the widgets inside scroll area
     - selection widget and tool buttons
  - toolSettings button 
  - toolSettings panel

Things like tool settings button and panel are added directly to CaptureWidget so that they are positioned relative to window and don't move when scrolling. Selection widget and tool buttons are placed inside the scrollview so that they stick to the image.

borgmanJeremy:
Awesome work and good progress! 

You might want to make sure you are based on the master branch because it might have moved a bunch since this started.

karliss:
There are few things remaining to fix, but afterwards comes the annoying part of testing in all the configs.

Test matrix:

Platform and mode
* Windows all screen fullscreen mode
* macOS current screen fullscreen
* Linux all screen fullscreen
* Linux window mode
* Other combinations nice to have, but less important

Screen configs (starting with most important ones)
* Single screen 1x scaling
* Multiple screens 1x scaling
* Single high DPI screen
* Multiple screens with identical >1 scaling
* Multiple screens mixed DPI

bbbco:
So I've been waiting patiently for Flameshot to implement this feature, and when I came across this pull request, I was excited!

I went ahead and cloned the flameshot repo, built it, and tested my build to ensure it worked properly. Then I fetched your branch @karliss and rebased against origin/master to pull in the latest updates and created a fresh build.

Running on Linux Mint 20.1 with a single monitor, when I attempt to get a new screenshot, the initial display is a blank grey screen. I can use the Flameshot tools here and eventually create a screenshot (that is not obvious to me because of the lack of display). See below:

image

Please let me know what other info you need from me to help you work through this issue.

Thanks for your initial work to attempt to implement this much needed feature!

karliss:
@bbbco The PR is still work in progress, please be patient. I had seen the the gray screen problem before on Windows and resolving it was one of the next things on my list, but it's interesting to see that it also happens on Linux. I have an approximate idea what should be changed.

bbbco:
Good to know!

Let me know if I can help in any way with the testing.

@bbbco

On July 24, 2021, Pedro Alonso ***@***.***> wrote:
> @bbbco <https://github.com/bbbco> The PR is still work in progress,
> please be patient. I had seen the the gray screen problem before on
> Windows and resolving it was one of the next things on my list, but
> it's interesting to see that it also happens on Linux. I have an
> approximate idea what should be changed.
>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/flameshot-org/flameshot/pull/1565#issuecomment-
886158628>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAAHPWC5AAKGRGWJ7KCZQH3TZO2WLANCNFSM43ECS3PQ>.

karliss:
Repeated the gray screen problem on Linux by forcing specific theme engine with `QT_QPA_PLATFORMTHEME=kvantum` .

karliss:
I haven't given up on this. Will try to finish soon.

karliss:
Ok, I have done the remaining things and caught up with changes in master. Time to try doing the all config testing again.

veracioux:
I have a few suggestions. Disclaimer: I'm just a contributor so don't take these on any kind of authority. And they should be discussed first (maybe open a discussion?)

#### Config option

I don't see that you have added an option to change the default window size. How about renaming the `windowMode` config option to `windowSize` and have its value specified as one of:
- WIDTHxHEIGHT
- FullScreenAll
- FullScreenCurrent
- MaximizeWindow
- DebugNonFullScreen

Whichever option name we choose, the `flameshot.example.ini` file should document them. And please add a `ValueHandler` for the option in `src/utils/valuehandler.{h,cpp}` for config checking to work. As a consequence, the `CONFIG_GETTER_SETTER` macro should probably be used in `ConfigHandler` instead of the manual implementation that is currently in place.

#### MaximizeWindow does not maximize the window
Instead the window is regular size. Maybe has something to do with the fact that I'm using a tiling WM (i3wm).

#### `NotifierBox`

The notifier box that is shown when tool thickness is changed is not displayed correctly in windowed mode if the window is not maximized.

#### Help message
The help message is not shown properly in full screen mode when multiple displays are used. It is shown at the center of all screens which means that part of the message will be on one screen and the other half on another.

If any of the suggestions are too burdensome, I can implement them in another PR. But I would like you to implement the `ValueHandler` and give me feedback about the developer-friendliness of the API (which I implemented just recently).

karliss:
One more thing I would like clarifying with regular devs is the debug mode. For now I added it as additional mode. That was done before introduction of build system flag. Any thoughts on how it should work?

veracioux:
@karliss

>> Instead the window is regular size. Maybe has something to do with the fact that I'm using a tiling WM

> Sounds like things are functioning as intended. Isn't that the point of tiling WM that they override most window resizing functionality except maybe for fixed size windows. I would prefer keeping the window mode resizable instead of forcing non resizable window. Maybe it should be renamed to just "windowed mode".

My bad. I should have been more clear. I meant for `WIDTHxHEIGHT` to be the default size, not to fix the window size and make it unresizable.

The flameshot window in my tiling WM appears in float mode (not restricted to the tiling layout). In this mode the window has free size. And float mode makes more sense since flameshot is not a program you would usually want to stay around for long. But the default size of the window is too small for me, so I'd like to be able to change it. Maybe the best course would be to have `MaximizeWindow` which opens in windowed mode but makes the window fullscreen by default and add a `Windowed` mode which is specified in the config as `WIDTHxHEIGHT`.

In a regular DE, does `MaximizeWindow` actually take up the whole screen?

> Which fullscreen mode, Fullscreen all screens or fullscreen current screen?

FullScreenAll. I have two monitors: left one is 1680x1050, right one is 1920x1080. They are placed right beside each other, with no gap. The message appears in the center of where both screens meet. Instead I believe it should appear in the center of the primary screen. In window mode, the message should appear in the center of the window.

> > WIDTHxHEIGHT

I agree that we can leave this for a future PR.

> One more thing I would like clarifying with regular devs is the debug mode.

I think the build system flag will become obsolete after this PR, if we add a `--config` option to specify the exact config file which I may have already suggested in an issue. This way developers could test flameshot using a different config file. So I vote to leave the build flag as it is until a future PR.

karliss:
> In a regular DE, does MaximizeWindow actually take up the whole screen?

Yes, at least on Gnome it takes up whole screen except the taskbar area whichis what I expect it to do. The code is calling `ShowMaximized()`.

borgmanJeremy:
@veracioux that plan sounds good to me.

PR #1569, opened 2021-04-20
fix - Crash on Move Tool with an active Text editor

thepurple:


borgmanJeremy (approved):

mmahmoudian:
This solves both https://github.com/flameshot-org/flameshot/pull/1539#issuecomment-823200606 and https://github.com/flameshot-org/flameshot/pull/1539#issuecomment-823203897 :+1:

PR #1582, opened 2021-04-25
RFC: Add an opacity slider to the Tool Settings

PrSunflower:
## RFC for issue #249 .

Please assist me in the RFC, I am not a developer and there are parts I cannot write due to lack of knowledge + understanding.

Thanks a lot for the templates @mmahmoudian  and @borgmanJeremy 😀.


borgmanJeremy:
We didnt end up really using RFC's but im going to merge this to approve it because I also would like opacity added.

PR #1598, opened 2021-04-29
CMake: add option to use external singleapplication

pavel-the-best:
- Look for QtSingleApplication in system. The version is hardcoded,
  because QtSingleApplication does not provide symlink to the current
  version
- Add ifdef that changes the included filename and SingleApplication
  class name depending on the option value

Signed-off-by: Pavel Kalugin <pavel@pavelthebest.me>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,7 +86,17 @@ enable_sanitizers(project_options)
 set(QAPPLICATION_CLASS
         QApplication
         CACHE STRING "Inheritance class for SingleApplication")
-add_subdirectory(external/singleapplication)
+
+if(USE_EXTERNAL_SINGLEAPPLICATION)
+  # look for external QtSingleApplication
+  # package dev-qt/qtsingleapplication provides no symlink to current version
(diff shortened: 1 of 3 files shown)


hosiet:
Warning to anyone reviewing this PR: This PR implicitly assumes that the external SingleApplication implementation is the QtSolutions' one ( https://github.com/qtproject/qt-solutions/tree/master/qtsingleapplication ), not Itay Grudev's one ( https://github.com/itay-grudev/SingleApplication ). I'm not sure if it is intended.

pavel-the-best:
It is intended. Most distributions provide qtsolutions' implementation if they do provide any.

borgmanJeremy:
Is it api compatible? I don't like the idea of shipping totally different versions.

pavel-the-best:
This PR does not actually ship anything.
It just provides the ability to use already installed in the system library.
It seems, that the implementations are API-compatible (at least, flameshot works with no noticed flaws).

borgmanJeremy (approved):

borgmanJeremy:
I'm okay merging this but I cannot provide support if using the external dependency causes bugs. It looks like the main difference is the external one is QT4.

Issue #1568, opened 2021-04-20
Can't paste into non-Qt apps on Sway


alecmev:
**Flameshot version**

55e61b8

**Describe the bug**

100% the same bug as #1035, but the fix - [`QT_QPA_PLATFORM=xcb`](https://github.com/flameshot-org/flameshot/pull/1282/files#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2R63) - isn't enabled for Sway. The pasting works when I run `QT_QPA_PLATFORM=xcb flameshot`, but then I'm unable to select a region outside of the status bar (Waybar in my case).

[This is supposed to be fixed in Qt 5.15.2](https://bugreports.qt.io/browse/QTBUG-88293), which I have, but it doesn't appear to be so. I have [the exact same problem](https://github.com/telegramdesktop/tdesktop/issues/9836) with Telegram Desktop, which is also Qt-based, but unlike Flameshot I'm able to run it in XWayland without side effects.

~~Enabling "Use JPG format for clipboard" doesn't help either, so this isn't just a matter of getting rid of `application/x-qt-image` ([which is absent in the "JPEG mode"](https://github.com/flameshot-org/flameshot/blob/55e61b874320c864c3abbd47c61d33babfc5fe3d/src/utils/screenshotsaver.cpp#L59-L60)).~~ ["Use JPG format for clipboard" helps, but not with browsers.](https://github.com/flameshot-org/flameshot/issues/1568#issuecomment-830627639)

This sounds like a Qt bug, but reporting here for visibility, and to also figure out if there's something that can be done until it's fixed, e.g. utilize `wl-copy` instead of `QClipboard` when running on Sway.

**System Information**

Up-to-date Arch as of writing this, Sway 1.6, Qt 5.15.2.

borgmanJeremy:
Does anything here help? https://github.com/flameshot-org/flameshot/pull/1272

alecmev:
It doesn't, since I don't have any issues with capture. Are you unable to reproduce this?

nullobsi:
i can confirm this issue, as it's been annoying me a lot. if i copy into other applications, it appears to have the image/png mimetype, but is actually in BMP format. i feel like this would have to be a Qt issue, since I don't think there's an easy way to fix it from flameshot... unless I'm mistaken

mmahmoudian:
@borgmanJeremy @nullobsi  should this be marked as "Upstream" bug?

[...]

PR #1282, opened 2021-02-06
added wayland XCB hack for clipboard

borgmanJeremy:
Fixes #1245

PR #1601, opened 2021-05-02
Use image/png on Wayland to improve copy support

nullobsi:
uses image/png only when wayland is detected on linux. works around a Qt bug where all images formats declared by Qt return an image in BMP format

### DIFF
--- a/src/utils/screenshotsaver.cpp
+++ b/src/utils/screenshotsaver.cpp
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: 2017-2019 Alejandro Sirgo Rica & Contributors
 
 #include "screenshotsaver.h"
+#include "utils/desktopinfo.h"
 #include "src/core/controller.h"
 #include "src/utils/confighandler.h"
 #include "src/utils/filenamehandler.h"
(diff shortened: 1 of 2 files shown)


nullobsi:
closes #1568

borgmanJeremy (approved):

src/utils/screenshotsaver.cpp
```diff
@@ -25,6 +26,31 @@ ScreenshotSaver::ScreenshotSaver(const unsigned id)
...
+    QImageWriter imageWriter{ &buffer, "PNG" };
+    imageWriter.write(capture.toImage());
+
+    QPixmap pngPixmap;
+    bool isLoaded = pngPixmap.loadFromData(
+      reinterpret_cast<uchar*>(array.data()), array.size(), "PNG");
+    if (isLoaded) {
+        // Need to send message before copying to clipboard
+        SystemNotification().sendMessage(
+          QObject::tr("Capture saved to clipboard"));
```
alecmev, line 42: This sends a second notification in both branches of `saveToClipboard`.

alecmev:
@borgmanJeremy Should [this hack](https://github.com/flameshot-org/flameshot/pull/1282) be removed now, perhaps?

borgmanJeremy:
I *think* that fixes a different wayland issue but would need to test it.

nullobsi:
They both refer to the same Qt issue, but testing would definitely have to be done to make sure.

I've made another commit that fixes the double notification, so I open another PR as soon as I can figure out how to get github to work with me

PR #1604, opened 2021-05-03
improve screenshot saver copy

nullobsi:
screenshot saver had some duplicate code so I cleaned it up a bit.

- when "save after image is copied" is enabled, respect the JPEG clipboard option
- fix double notification when wayland workaround is in place
- unify jpeg and png workaround code paths

### DIFF
--- a/src/utils/screenshotsaver.cpp
+++ b/src/utils/screenshotsaver.cpp
@@ -26,7 +26,8 @@ ScreenshotSaver::ScreenshotSaver(const unsigned id)
   : m_id(id)
 {}
 
-void ScreenshotSaver::saveToClipboardMime(const QPixmap& capture, const QString& imageType)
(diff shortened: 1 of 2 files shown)


borgmanJeremy (approved):

borgmanJeremy:
This introduced a weird latency when I first take the screenshot. Will debug further.

nullobsi:
i always get a pretty big latency when using the desktop portal method- it's been there since I added it, I will also try some debugging...

nullobsi:
i don't think its very fixable, the biggest lag in the chain is the return from xdpw (~2 seconds or so) and the 2nd biggest seems to be whatever happens after the reply happens (file deletion, qpixmap creation?)

but if you are getting a bigger latency, then i am not sure

PR #1630, opened 2021-05-17
Removed hardcoded binary path

xvitaly:
Removed hardcoded binary path.

Some distributions or package formats does not uses `/usr/bin` at all.


borgmanJeremy:
Hardcoding the binary path was actually required to make it work with KDE Wayland. 

https://github.com/flameshot-org/flameshot/pull/1603/files

xvitaly:
Then you should use cmake_configure_file() instead of hardcoding paths. It will stop working in Flatpak now.

borgmanJeremy:
Yeah we can put either a configuration in cmake or simply sed the file in the CI. Feel free to submit a new PR that does this.

xvitaly:
> Yeah we can put either a configuration in cmake or simply sed the file in the CI. Feel free to submit a new PR that does this.

Are you sure that hardcoded path matters? I think that `X-KDE-DBUS-Restricted-Interfaces=org.kde.kwin.Screenshot` is enough.

borgmanJeremy:
I can test it again, but I think we had some back and forth on it and it was determined both was needed.

Issue #1640, opened 2021-05-21
Add support for the H265 file extension to save space 


iMonZ:
Please add support for the HEIF/HEIC file extension to save space in screenshots
(This file extension is supported on the file Hoster IMGBB too)

mmahmoudian:
Interesting suggestion. While searching I also stumbled upon this:

https://github.com/nokiatech/heif

kpcee:
The next standard and jpg successor will be AVIF or JPEG XL and will certainly be natively supported by Qt. Hopefully something will happen this year. By the way, to bridge the time until then we can use WebP, which is already natively supported by Qt 5.x and can also be displayed by most browsers. The compression is also much better than JPG.

Here is some background info: https://cloudinary.com/blog/how_jpeg_xl_compares_to_other_image_codecs

P.S.
H.265 is the abbreviation for the video codec and not for images.

dchakro:
@kpcee, Nice article about JPEG XL.

> By the way, to bridge the time until then we can use WebP, which is already natively supported by Qt 5.x and can also be displayed by most browsers. 

Btw, WebP is only [partially supported on [MacOS](https://apple.stackexchange.com/a/392466) & [Safari](https://caniuse.com/webp), vs HEIF which is very well supported on Mac.

kpcee:
However, the same page indicates that no browser currently supports [HEIF images](https://caniuse.com/heif). Which makes sense, since HEIF is patented and the browser companies have to purchase an expensive license to use it.

Whereas many browsers support WebP. Also Qt natively supports WebP, whereas support for HEIF would have to be built in separately only to be replaced later by AVIF or JPEG XL. I strongly expect that Qt 6.x will later support at least one of these formats natively.

iMonZ:
The point is about saving pictures as HEIF.

borgmanJeremy:
As of #1680  we allow saving in any supported format.

PR #1674, opened 2021-06-20
Add WebP support for saving captures

yamanq:
WebP can be selected to save an image in the "Save As" menu. Tested on Linux with KDE. This partially addresses #1640


yamanq:
Closed in favor of #1680

Issue #1741, opened 2021-07-22
disable capture full desktop behavior when user do capturing.


zw963:
**Is your feature request related to a problem?**

Yes, please check #1654 

This issue happen frequently when use a laptop, and enable `tap-and-drag-lock` feature, this will cause
when use with flameshot, often can not release the mouse successful after draw a rectangle.

in this case, when user press `Ctrl+C`, will capture the full desktop, that was not expected.

So, the expected behavior is:

when user started to create a rectangle region use mouse/touchpad, but before release mouse/touchpad to create the region successful, we should disable the default capture full desktop behavior temporarily, instead of just no operation, after region created, `Ctrl+C` wil capture the rectangle area as old behavior.

Thank you.

chuahou:
The same behaviour occurs (but with saving instead of copying to clipboard) when pressing `Ctrl+S` in the same situation. Should the behaviour of both shortcuts be changed when the mouse is not yet released?

mmahmoudian:
I believe literally every action should be disabled if the user is still "dragging". Anything apart from hitting Esc to cancel. This would also be inline with Inkscape and other software

chuahou:
I misread the intention of the original issue, thinking that Ctrl+C/S should select the region selected so far instead of that the shortcuts should be disabled, and had already worked on a PR.

I do believe that disabling all shortcuts makes more sense, but I opened the PR anyway just to keep that option open.

zw963:
For the issue i meet, I thought both solution resolve it anyway, thank you.

Issue #1714, opened 2021-07-19
The colour picker seems to detect colours from a wrong place on the screen


micnekr:
**Flameshot version**
Flameshot v0.10.0
Compiled with Qt 5.15.2

**Describe the bug**
When using the "grab color" option in "tool settings", the tool seems to detect a color from a different portion of the screen, for example here: 
  
![2021-07-19_11-55](https://user-images.githubusercontent.com/44928743/126150268-1c8afd31-03b7-4a6b-80bb-c239c6436399.png)

If I move the color picker to the location of the pink cross, it seems to detect the color at the location of the green cross. If I move the color picker around, it shows colors adjacent to the green cross.

**To Reproduce**
Run `flameshot gui` and select "tool settings", press the "grap color" button and then move the cursor somewhere on the screen.

**Expected behavior**
The color selected should be the same as the color under the cursor

**System Information**
Arch Linux, KDE plasma.

mmahmoudian:
Please fill the bug report with extra attention to details so that we can better understand the situation and move towards fixing it. You didn't mention if you are on Wayland or Xorg. If you don't know how to find that information, please check out the following:

https://flameshot.org/guide/issue-reporting/

borgmanJeremy:
My gut says this is a fractional scaling issue since this works for me

micnekr:
> Please fill the bug report with extra attention to details so that we can better understand the situation and move towards fixing it. You didn't mention if you are on Wayland or Xorg. If you don't know how to find that information, please check out the following:
> 
> https://flameshot.org/guide/issue-reporting/

Yes, sorry, I did not see that this information is needed. I am using X. Also, it does seem to be an issue with scaling because changing "global scale" in plasma to 100% made everything work as expected.

thepurple:
This is a fractional scaling with scales like 125%, 150%, 175%. It should be fine on 100% and 200%.
Possibly it will be solved if we'll start to capture just current screen like on MacOS (also commented here: https://github.com/flameshot-org/flameshot/issues/1610#issuecomment-883524218).

PR #1749, opened 2021-07-24
Copy/save selected region even before mouse release

chuahou:
This fixes the counter-intuitive behaviour where copying/saving before releasing the mouse button (via shortcuts) copies/saves the entire screen instead of the region selected so far.

It does so by updating the selected region's geometry (in the manner handling of mouse release events does it) before performing the copy/save.

Since there is now repeated functionality, this updating is extracted to `CaptureWidget::updateGeometry()`.

---

Motivating issue: #1741

(Misread the intention of the issue...)

---

It's probably better (as @mmahmoudian mentioned in #1741, and as proposed by the issue in the first place) to disable these shortcuts during the selection phase instead, but since I'm done with this approach I'm opening this PR as an option.

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -721,23 +721,9 @@ void CaptureWidget::mouseReleaseEvent(QMouseEvent* e)
         // Show the buttons after the resize of the selection or the creation
         // of a new one.
 
-        // Don't go outside
-        QRect newGeometry = m_selection->geometry().intersected(rect());
-        // normalize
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
Since we are going to eventually fix this with the solution in #1714 I am inclined to close this so we dont need to adjust it later. Thoughts?

chuahou:
Sorry, do you mean #1741? I'm not too sure how #1714 relates.

In any case, I think it'd be safe to close this since the preferred solution mentioned in #1741 is to disable shortcuts while selecting instead.

PR #1750, opened 2021-07-24
Add more clipboard format options

### LINKED ISSUE #1738: Not Copying into Clipboard
Flameshot version: 0.10.0-1
OS: Arch with Gnome
Installed from AUR

Just captured a section like I always did, Ctrl+C to copy the image and paste to discord, nothing happens. Same result even if I manually click the copy button on the GUI.

Had been working before, could it be the newest version's problem?

Copied an image from discord directly and pasted it alright. So my clipboard is working as it should.

thetaprimeprime:
This is mainly to fix #1738 .

Because some programs only detect particular image formats in clipboard.
For example, xfreerdp only accepts BMP, but Firefox prefers PNG.

If we specify only one format, some programs would complain when you try to paste. So I added an `Auto` option to store as many as possible formats into clipboard.

image

There are four possible values:
* Auto
  Store as many as possible formats into clipboard.
* PNG
  Store only PNG into clipboard.
* JPEG
  Store only JPEG into clipboard.
* BMP
  Store only BMP into clipboard.

Because there are many different systems, no one of the above could make clipboard working all platforms. Users should try each of them to see if it works as expected.


borgmanJeremy:
Is there a reason this is wayland only?

thetaprimeprime:
> Is there a reason this is wayland only

Because the old option `useJpgForClipboard` only affects Wayland. Since I extend this function, I think it is still wayland only.

src/config/generalconf.cpp
```diff
@@ -464,25 +441,40 @@ void GeneralConf::undoLimit(int limit)
...
-      new QCheckBox(tr("Use JPG format for clipboard (PNG default)"), this);
+    m_clipboardFormat = new QComboBox(this);
     ConfigHandler config;
-    bool checked = config.useJpgForClipboard();
-    m_useJpgForClipboard->setChecked(checked);
-    m_useJpgForClipboard->setToolTip(
-      tr("Use JPG format for clipboard (PNG default)"));
-    m_layout->addWidget(m_useJpgForClipboard);
+
+    m_clipboardFormat->setToolTip(tr("Clipboard foramt (only for Wayland)"));
```
borgmanJeremy, line 449: Can you fix typo in "format"

src/config/generalconf.cpp
```diff
@@ -464,25 +441,40 @@ void GeneralConf::undoLimit(int limit)
...
+    m_clipboardFormat->addItem(tr("PNG"), "png");
+    m_clipboardFormat->addItem(tr("JPEG"), "jpeg");
+    m_clipboardFormat->addItem(tr("BMP"), "bmp");
+
+    QString format = ConfigHandler().clipboardFormat();
+    m_clipboardFormat->setCurrentIndex(m_clipboardFormat->findData(format));
+
+    m_layout->addWidget(m_clipboardFormat);
+
+    QGroupBox* box = new QGroupBox(tr("Clipboard foramt (only for Wayland)"));
```
borgmanJeremy, line 460: typo in "format"

src/utils/screenshotsaver.cpp
```diff
@@ -64,21 +64,23 @@ void ScreenshotSaver::saveToClipboard(const QPixmap& capture)
...
-    } else {
-        // Need to send message before copying to clipboard
+
+    // Need to send message before copying to clipboard
 #if defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
-        if (DesktopInfo().waylandDectected()) {
-            saveToClipboardMime(capture, "png");
-        } else {
+    QString format = ConfigHandler().clipboardFormat();
+    if (DesktopInfo().waylandDectected()) {
```
borgmanJeremy, line 71: Let's allow this on X or Wayland. I think it should work on both.

borgmanJeremy (changes requested):

Issue #1732, opened 2021-07-21
Add a copy button to the About window


mmahmoudian:
I suggest adding a button in the bottom of the About window to copy the version info into the clipboard. This can assist users to easily copy the information for bug reports.

![image](https://user-images.githubusercontent.com/390889/126475081-cc7a91e1-408f-4154-ad6f-efdb30f583c6.png)

borgmanJeremy:
Check out #1756

borgmanJeremy:
Im going to close this since the first version is merged in. If you have more suggestions for metrics to collect let me know and I'll add it.

PR #1756, opened 2021-07-25
Added ability to copy version info

borgmanJeremy:


borgmanJeremy:
Closes #1732

hosiet:
Code looks OK, but I'm curious about the implementation choice. Why not make the version string selectable by users so that users can copy it by themselves? That would avoid messing around with clipboard (again).

borgmanJeremy:
I could do that too, @mmahmoudian suggested this implementation. 

I'm fine with either way.

mmahmoudian:
I actually had a longer plan for this button and thought I start with this. My longer-term plan was to add more info to this button so that users simply press this button and get the info in a unified format we need for debugging in their clipboard. Version is generally the major info we need, but OS, Xorg or Wayland, DE, ...

I know some are not easy to determine (like DE), but we can have some of the info already there.

borgmanJeremy:
We can get many of those things through Qt directly: https://doc.qt.io/qt-5/qsysinfo.html

hosiet:
Including full info might be a good idea. Actually making text selectable (thus can be copied) can coexist with button. They may all be implemented to make the retrieval of sysinfo easier.

mmahmoudian:
Then @borgmanJeremy how about make the text selectable and for now have the button put some minimal info (eg. Version & OS) in the clipboard? Would that be easy and fast?

borgmanJeremy:
Yup I can do that tonight.

borgmanJeremy:
@mmahmoudian give the latest build a try and let me know if you think of more to be added before merging.

mmahmoudian:
image

The OS is not copied though

PR #1775, opened 2021-07-27
Automatically set the correct path to the Flameshot executable

### LINKED ISSUE #1766: flameshot path hard coded in .desktop file
Hello,

The change in https://github.com/flameshot-org/flameshot/commit/57dac0d73de3b3de7838a639467a4fe3bbdfe087 breaks the desktop file on some systems/distributions that install flameshot to a custom location like `/usr/local`. Packagers can fix this, but it would be nice if upstream could honor `PREFIX` here and install it with the correct path.

**System Information**
OpenBSD-current

xvitaly:
Closes #1766.


xvitaly:
GNU/Linux maintainers can now forward `DUSE_LAUNCHER_ABSOLUTE_PATH:BOOL=OFF` to avoid hardcoding of absolute paths. Now used for AppImages instead of patching it manually.

borgmanJeremy (approved):

Issue #466, opened 2019-02-11
Is there a way to change the title of imgur uploaded images?


SulaimanAlabbar:
https://i.imgur.com/caNEerC.png
The title is "flameshot_screenshot".

It's possible to change the name of an image(which sets the description of an image on imgur) but not the title.

AlfredoRamos:
Not currently, it's title is hard-coded:

https://github.com/lupoDharkael/flameshot/blob/8887b4e7ceb9f3bf7c2e8262ffe0518dcb68522a/src/tools/imgur/imguruploader.cpp#L104

Title and description are optional, and in my opinion it would be better to exclude them from the API request.

SulaimanAlabbar:
I wanted to blank both out. Some places would embed the imgur image differently depending on whether or not it has a title.

AnInternetTroll:
Hey, any update on this?

mmahmoudian:
@AnInternetTroll as far as I know and as far as I can see in the code, it has not been implemented. There has been also very low demand for this (you are the second person in 2 years).

But regarding this feature request, it is very clear fir me that what do you and @sulaiman1073 specifically want. Would it be enough to have a string in the config? Or do you want it to be something that Flameshot asks from user every time before uploading (this would be very annoying for some people and there must be then a config added to toggle this behaviour).

It would be nice if you can in details explain what is in your mind so that devs or the community can see its importance and also how it should be implemented.

AnInternetTroll:
A string in the config is more than enough. In my use case (which is discord) I just want to remove the title and description since it embeds with those, and I like to be able to send just a photo, no other metadata. Asking for a title every time I take a screenshot seems overkill and unnecessary

AnInternetTroll:
Here's an example of the embed I'm talking about in the context of discord https://i.imgur.com/5y8Y2dB.png

PrudePrograms:
I'd like to echo this - the default title of the uploads really doesn't serve any useful purpose for my use case at least, and as previously noted, makes Discord embeds a bit more inefficient (by making them non-direct), which is the main place they end up being posted in my case. An option in the config to change it (for those who do find the feature useful) and/or disable/blank it entirely would be great. An additional demonstration: [https://i.imgur.com/YiJHMQn.png](https://i.imgur.com/YiJHMQn.png). Otherwise, Flameshot feels great to use and is pretty much at parity with ShareX for my purposes (which I have historically used when I'm on Windows).

joobus:
[...]

PR #1778, opened 2021-07-27
Remove imgur image title

Chryseus:
Removes the image title to make embedding images in discord much nicer.
Without changes to set a custom title this is a reasonable solution to issue #466


borgmanJeremy:
fixing clang formatting on #1791

PR #1783, opened 2021-07-29
flameshot to dharkael

IsmailHazem:
On my Kubuntu machine. Global shortcuts didn't work. I went to the (Global Shortcuts) window and triggered the (Take Screenshot) action: Global Shortcut for Flameshot
I saw in the D-Bus Browser that the call was to (org.dharkael.Flameshot) instead of org.flameshot.Flameshot).
d-bus browser


borgmanJeremy:
I suspect you are on an old version

org.flameshot.Flameshot is correct

IsmailHazem:
You're right. Sorry

borgmanJeremy:
No problem. Thanks for the offer to contribute! I always appreciate that.

Issue #496, opened 2019-03-28
Setting to always save as JPG


erikdubois:
Is it already possible to set flameshot to always save as JPG rather than PNG.

JPG is much better for websites than PNG regarding the size of the file.

Neyromancer:
Hi i am new here but if it is needed i can try to do this. It seems it is possible to implement by rewriting a bit code in screenshotsaver.cpp

magnus-gross:
JPG is literally the worst lossy image format, you need to let it die already. If you do need a lossy format, use something like BPG (or any other modern lossy format) instead.

erikdubois:
I do not care about that. I need an image with that is low in KB for the website.
I want a fast loading webpage.

magnus-gross:
That's literally, what I was talking about.

A BPG file has about half the size compared to a JPG file of equal quality.

Jacalz:
I think this along with the other supported formats should have an option in the configuration tool where you can set the default format, not just for JPEG.

erikdubois:
Any prospect on the development to have the possibility to save as jpg.

AlexP11223:
>JPG is much better for websites than PNG regarding the size of the file.

it depends on the content, sometimes png has smaller size.

![](http://lbrandy.com/assets/jpg_vs_png2.png)

arcolinuxz:
Has anything been done about this question?
Can we / will we be able to save as jgp?
If it is never gonna be on the agenda, please tell us and close the issue.

AlexP11223:
@arcolinuxz it's a relatively small open source project, the best way to make something happen is to implement it yourself and send PR.

jam-2000:
> Hi i am new here but if it is needed i can try to do this. It seems it is possible to implement by rewriting a bit code in screenshotsaver.cpp

It will be awesome!!!
A lot people need JPG very much!

erikdubois:
Follow up post - is there any development going on on this subject?

th317erd:
It wouldn't be hard to setup a watcher script that would watch the folder flameshot saves to and automatically generate a JPG for every PNG in the folder...

blaueente:
Don't forget the use case with ->clipboard->chat; there, the quality loss is not important at all, but nobody wants to wait for 5MB to be transmitted.

navigante:
Yes, lossy compression format would be strongly appreciated ...

kirjeldus:
[...]

PR #1790, opened 2021-07-31
Allow saving to any supported image format

borgmanJeremy:


vsahishnu:
Hi Flameshot team. 

I was wondering, any update on this [commit](https://github.com/flameshot-org/flameshot/pull/1790/commits/a640a58b8e1d2c6253f24a14c36e36ebaa110c20)? 

To be particular, I see that, when we copy images directly without saving it (in MacOS), the image will be pasted in the .tiff format. Any idea if this can be changed to something like what @borgmanJeremy suggested in this [comment](https://github.com/flameshot-org/flameshot/issues/496#issuecomment-698483762)?

Issue #1543, opened 2021-04-08
Uncheckable checkbox in configuration


Mibou:
**Flameshot version**
v0.9.0 for windows installed from MSI

**Describe the bug**
Clicking on "Use JPG format for clipboard (PNG default) in the configuration leaves the checkbox blank.

**To Reproduce**
Go to configuration
Click the checkbox
![image](https://user-images.githubusercontent.com/1668996/114036279-39ecfb00-9880-11eb-979f-def2e36486bb.png)

**Expected behavior**
The checkbox should become checked but isn't (only the onclick shadow appears in)

![image](https://user-images.githubusercontent.com/1668996/114036569-7a4c7900-9880-11eb-87a7-dcb9d0a474de.png)

**System Information**
Windows 10

mmahmoudian:
It works fine on Linux (Manjaro KDE), perhaps it is Windows-specific.

remggo:
I have this issue as well, also on Windows 10.

HarisGusic:
I have added a PR that fixes this issue.

Anyway, the problem was the directive `#if not defined(Q_OS_MACOS)` that has an unexpected behavior when compiled with MSVC. The correct syntax is `#if !defined(Q_OS_MACOS)`. The good news is that this also works in gcc.

There are more places in the code that contain this incompatibility and I will create a separate issue for that.

I have also discovered another bug and fixed it in the PR. Namely, if the daemon was started with the system tray disabled, and it never gets enabled subsequently, interacting with the 'Automatic updates' checkbox crashes the program. What happens is that the action `Controller::m_appUpdates` never gets constructed. So when the 'Automatic updates' checkbox was interacted with it tried to access this variable which is null. Hence crash.

I have also cleaned up some redundant parts in the code.

I've tested all my changes on Windows 10, Arch Linux and an Ubuntu VM.

gaduaxe:
I got same issue here: cannot Check on "JPG using" and also output file cannot be "JPG"

Plz fix

veracioux:
@gaduaxe Are you using the version of flameshot from the latest commit on master? Or at least from [this commit](https://github.com/flameshot-org/flameshot/commit/b2ecb70e386c51315d25fd1ab5b45a7bfc9e04a9). Also, you might have to close the flameshot daemon if it was running previously.

gaduaxe:
[...]

PR #1793, opened 2021-08-01
Fix checkboxes in General Config (#1543)

veracioux:
- Cleaned up code in generalconfig.cpp
- Fixed useJpgForClipboard checkbox that was unselectable on Windows
- Fixed crash  when 'Automatic update' is checked with system tray disabled
- Corrected tooltip text under 'Show welcome message on launch' checkbox

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -225,6 +225,7 @@ void GeneralConf::initShowSidePanelButton()
             this,
             &GeneralConf::showSidePanelButtonChanged);
 }
+
 void GeneralConf::initShowDesktopNotification()
 {
     m_sysNotifications = new QCheckBox(tr("Show desktop notifications"), this);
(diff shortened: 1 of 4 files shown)


src/config/generalconf.cpp
```diff
@@ -51,7 +51,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     updateComponents();
 }
 
-void GeneralConf::updateComponents()
+void GeneralConf::_updateComponents(bool allowEmptySavePath)
```
borgmanJeremy, line 54: Small detail but maybe this should have a more descriptive name.

borgmanJeremy (approved):

src/config/generalconf.cpp
veracioux: This is the best I could come up with. My reasoning was that it should have the same name as `updateComponents` because it does the same thing except for a small tweak. And the underscore indicates that it is a helper function. Your suggestions are welcome.

gaduaxe:
> * Cleaned up code in generalconfig.cpp
> * Fixed useJpgForClipboard checkbox that was unselectable on Windows
> * Fixed crash  when 'Automatic update' is checked with system tray disabled
> * Corrected tooltip text under 'Show welcome message on launch' checkbox

Hi

I still got issue on vers 0.10.1 (latest) when cannot Check on "JPG" - Windows 10 Pro x64
image

Plz fix

PR #1801, opened 2021-08-04
Refactor code related to CaptureWidget

veracioux:
Some notes:
- I think we should gradually move towards decoupling `CaptureTool` from
  `CaptureToolButton`. The code will be more readable that way and it will add
  more room for features that allow tools to be activated by means other than
  by buttons (or shortcuts).
- The `pixmap()` method contained a block of code that modified a local object
  without returning it, hence being useless. This could either be redundant
  code, or a bug. From what I could gather this was meant to apply the changes
  to the output image if the text tool was active when you triggered
  save/copy/etc. But I have verified that the image gets saved regardless, so I
  don't think it is a bug.
- I smuggled in a commit that refactors the config part a bit. I was supposed
  to add it in a previous pull request, but I accidentally put it in the
  wrong branch so here it is :)
- In the latest commit I added a small comment among all things, please verify that it is
  correct
- Just throwing this idea in: I think signals and slots should to be renamed so
  that you can identify which is which from the name. I haven't done this here.
- This is just a thought of mine, haven't really thought it through: Maybe it
  would be more elegant to do away with a fair bit of sigslots. Maybe store a
  reference or pointer to `CaptureWidget` in the tools/buttons. And let them
  access various functions directly. One major beneficiary of this would be
  the `requestAction` mechanism, which doesn't seem very elegant to me.
- It seems to me that there are a lot of instances where something is
  implemented in an abstract class that only a few derived classes use. This
  point is related to the previous one.

Perhaps these observations have been made already. But I can't risk any idea going
to waste!

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -1026,7 +1026,12 @@ void CaptureWidget::setState(CaptureToolButton* b)
         }
     }
     if (m_activeButton != b) {
-        processTool(b->tool());
+        auto backup = m_activeTool;
+        // The tool is active during the pressed().
+        // This must be done in order to handle tool requests correctly.
+        m_activeTool = b->tool();
+        m_activeTool->pressed(m_context);
+        m_activeTool = backup;
     }
 
     // Only close activated from button
@@ -1065,15 +1070,6 @@ void CaptureWidget::loadDrawThickness()
         m_context.thickness = m_config.drawThicknessValue();
     }
     m_sidePanel->thicknessChanged(m_context.thickness);
-}
-
-void CaptureWidget::processTool(CaptureTool* t)
-{
-    auto backup = m_activeTool;
-    // The tool is active during the pressed().
-    m_activeTool = t;
-    t->pressed(m_context);
-    m_activeTool = backup;
 }
 
 void CaptureWidget::handleToolSignal(CaptureTool::Request r)
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -129,9 +129,11 @@ CaptureWidget::CaptureWidget(uint id,
         move(currentScreen->geometry().x(), currentScreen->geometry().y());
         resize(currentScreen->size());
 #else
-        // Comment For CaptureWidget Debugging under Linux
+
+#if (defined(QT_NO_DEBUG) && defined(Q_OS_LINUX))
         setWindowFlags(Qt::BypassWindowManagerHint | Qt::WindowStaysOnTopHint |
                        Qt::FramelessWindowHint | Qt::Tool);
+#endif
         resize(pixmap().size());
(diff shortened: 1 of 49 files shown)


src/widgets/capture/capturewidget.cpp
```diff
@@ -129,15 +131,17 @@ CaptureWidget::CaptureWidget(uint id,
         move(currentScreen->geometry().x(), currentScreen->geometry().y());
         resize(currentScreen->size());
 #else
-        // Comment For CaptureWidget Debugging under Linux
+
+#if (defined(QT_NO_DEBUG) && defined(Q_OS_LINUX))
```
borgmanJeremy, line 135: Why is behind QT_NO_DEBUG?

borgmanJeremy (changes requested):

src/widgets/capture/capturewidget.cpp
veracioux: Not sure I understand.

borgmanJeremy:
> Just throwing this idea in: I think signals and slots should to be renamed so
that you can identify which is which from the name. I haven't done this here.

Yeah this is fine with me. I switched between Clion and just vim. Neither of these have a good way of identifying what is a signal and what is a slot. 


>This is just a thought of mine, haven't really thought it through: Maybe it
would be more elegant to do away with a fair bit of sigslots. Maybe store a
reference or pointer to CaptureWidget in the tools/buttons. And let them
access various functions directly. One major beneficiary of this would be
the requestAction mechanism, which doesn't seem very elegant to me.

I think in some cases a function pointer (or maybe even a lambda) would work better. In other cases I think the signal and slot mechanism is useful, but I trust your judgement on this if you start refactoring. 


>It seems to me that there are a lot of instances where something is
implemented in an abstract class that only a few derived classes use. This
point is related to the previous one.

Yes :) Flameshot started as a much more simple application and over time has had pieces bolted on and never had a good refactor. This is more and more apparent as we add more features.

src/widgets/capture/capturewidget.cpp
veracioux: Nevermind, I do. It is a stupid error. I will fix it.

veracioux: @borgmanJeremy Done.

borgmanJeremy: Something weird happened.... I see you pushed 2 commits:

3df20d3: Fixed issue
d2a83c8: Reintroduced issue

image

veracioux: No, this was intentional. Now the only time it doesn't get executed is in debug builds on linux. I just translated the comment into a preprocessor directive. You don't want this?

borgmanJeremy: Ohhhh I see the change now. I'll do some testing and then merge it. I want to test against a bunch of DEs

Issue #1785, opened 2021-07-29
Add --format option to the CLI


veracioux:
It would be nice to have a `-f`/`--format` option that goes along with the `--path` and `--raw` option. The option should take a single argument which can be any of the supported formats (png, jpg, bmp). This is somewhat related to #1784

I can implement this feature.

mmahmoudian:
Your suggestion makes general sense and brings clarity to CLI, but we should be vigilant about the conflicts. For instance I remember that someone implemented something once to determine the file format based on the given extension. So what happens when the given file is png but the format (with `--format`) is jpg?

HarisGusic:
@mmahmoudian I'm not completely sure how to go about this. The approach in CLI programs generally is to not care at all about the file name. Sometimes you want to save to an extensionless file.  But this approach is more prone to error. Therefore I'd say on balance, the best approach is to do as other programs do.

I can't think of many relevant programs off the top of my head, but I tried this in inkscape and it just replaces the extension and prints a warning. If the file has no extension it appends it, without any warning.

If you have more examples, please share.

mmahmoudian:
@HarisGusic As a rule of thumb, I go with Inkscape whenever possible. It is a very well designed software with years and years of history and therefore it should be a very good template for such things.

Issue #1784, opened 2021-07-29
A few improvements to the --path option


HarisGusic:
I Would like to propose a few changes to the behavior of the --path option. 

1.  Relative paths don't work right in my opinion (don't know if they are even meant to be supported). If I run `flameshot screen --path .` I expect the screenshot to be placed in `.` relative to my current  working directory. Instead it gets placed relative to the working directory of the daemon, which is usually `~`.

2. Related to 1, when screenshots are saved to a relative path, the notification
   should contain the **absolute** path.

3. Only existing directories are currently allowed to be passed to `--path`. I think we should
   allow paths to also contain the destination file. I will illustrate this with
   a few examples:
   - If path/to/dir exists and is a directory:
    The behavior should remain the same, i.e. the screenshot should be saved to
    `path/to/dir/%F_%H-%M.png` or whatever the user set as the file template.

   - If `path/to/dir/filename` doesn't exist and `path/to/dir` is an existing directory:
    The file should be saved to `path/to/dir/filename`.

   - In all the other cases: `path/to/dir` is not an existing directory or if `path/to/dir/filename`is an existing file; an Invalid path error should be reported. We could add a mkdir or overwrite the file, but I think that would be too intrusive. Alternatively, if the file exists, we could just use `FileNameHandler::fixPath` on the file name. I also propose that the file name not be altered at all (no additional extensions and such), which would make for more consistent behavior in scripts.

I have already identified the relevant places in the code and I am ready to start
working on this upon your approval.

mmahmoudian:
I think the 3rd point also addresses #1224

I general I agree with the approach that of the file name is defined, it should be saved to that file. This would make mage people that use `flameshot gui --raw` happy esp. if they are not on Linux.

veracioux:
We might want to keep this in mind as well: #751

veracioux:
I have implemented the changes in a draft PR [here](https://github.com/flameshot-org/flameshot/pull/1815). In the implementation I have amended my proposal about the alteration of file names. The way it is implemented is as I discussed with @mmahmoudian in #1785 (following Inkscape's example).

Issues #217 and #1224 should be covered by this. Issue #751 is also partly covered but needs some more discussion. Please correct me if I overlooked some detail. For the other issues, we will have to do some figuring out. I have a few more things to implement (CLI messages, file save dialogs etc.), but I hesitate to proceed without some review and discussion.

[...]

PR #1815, opened 2021-08-09
Improve screenshot path handling

veracioux:
This is a work in progress. Tracks issue #1784

**Some notes:**
- The `format` argument to `properScreenshotPath` was added to predict for issue #1785.
- The reason I decided to rework `FileNameHandler` is because I wanted something easier to work with. And the previous implementation was confusing.


src/utils/filenamehandler.cpp
```diff
@@ -52,70 +52,71 @@ QString FileNameHandler::parseFilename(const QString& name)
...
-{
-    ConfigHandler config;
-    directory = config.savePath();
-    if (directory.isEmpty() || !QDir(directory).exists() ||
-        !QFileInfo(directory).isWritable()) {
-        directory =
-          QStandardPaths::writableLocation(QStandardPaths::PicturesLocation);
+    QFileInfo info(path);
+    if (info.isDir()) {
+        return QDir(QDir(path).absolutePath() + "/" + parsedPattern() + "." +
```
mmahmoudian, line 79: Is using the "/" in the file path valid for Windows?

mmahmoudian:
image

mmahmoudian:
I tried with the AppImage from this PR, and also commented out all lines of `path_option.sh` that were not necessary in my opinion:

<details>
<summary>My version of <code>path_option.sh</code></summary>

```sh
#!/usr/bin/env sh

# pkill -i flameshot
# flameshot &
# 
# sleep 4
# 
# # Create a directory for file output
# rm -rf _out 2>/dev/null
# mkdir -p _out
# cd _out
# 
# # NOTE THIS
# # The dir _out will be have as the home directory
# HOME="$PWD"
# mkdir -p subdir
# 
echo ">> Nonexistent directory is not a valid path"
flameshot screen -p ~/blah/blah

sleep 2
echo ">> Saved relative to PWD"
flameshot screen -p relative.png

sleep 2
echo ">> Absolute paths work too"
flameshot screen -p ~/absolute.png

sleep 2
echo ">> Redundancy in the path will be removed"
flameshot screen -p ~/subdir/..///redundancy_removed.png

sleep 2
echo ">> If the destionation is a directory, the file name is generated from strf from the config"
flameshot screen -p ./

sleep 2
echo ">> If the output file has no suffix, it will be added"
flameshot screen -p ~/without_suffix

sleep 2
echo ">> If the suffix does not match the output format, it will be overwritten"
flameshot screen -p ~/wrong_suffix.jpg

sleep 2
echo ">> If the destination path exists, it will have _NUM appended to the base name"
flameshot screen -p ~/absolute.png

sleep 2
echo ">> We will do that again but without a suffix"
flameshot screen -p ~/absolute

sleep 2

# sleep 3
# 
# pkill flameshot
```

</details>

And I got the following:

<details>
<summary>The output I got after running the <code>path_option.sh</code></summary>

> ```
> >> Nonexistent directory is not a valid path
> Invalid path, it must be a real path in the system. See flameshot --help.
> >> Saved relative to PWD
> Invalid path, it must be a real path in the system. See flameshot --help.
> >> Absolute paths work too
> Invalid path, it must be a real path in the system. See flameshot --help.
> >> Redundancy in the path will be removed
> Invalid path, it must be a real path in the system. See flameshot --help.
> >> If the destionation is a directory, the file name is generated from strf from the config
> >> If the output file has no suffix, it will be added
> Invalid path, it must be a real path in the system. See flameshot --help.
> >> If the suffix does not match the output format, it will be overwritten
> Invalid path, it must be a real path in the system. See flameshot --help.
> >> If the destination path exists, it will have _NUM appended to the base name
> Invalid path, it must be a real path in the system. See flameshot --help.
> >> We will do that again but without a suffix
> Invalid path, it must be a real path in the system. See flameshot --help.
> ```

</details>

These show that regardless of what happens, we always get the same exact error.

src/utils/filenamehandler.cpp
veracioux: QDir should convert the delimiter automatically.

veracioux:
> I tried with the AppImage from this PR, and also commented out all lines of `path_option.sh` that were not necessary in my opinion:

Sure the lines that start the daemon and kill it are unnecessary and good on you for removing them. But I would still create an output directory so home doesn't get filled with junk. But I will change the destination to somewhere under /tmp.

I'm not sure why you are getting the invalid path errors, but did you try to run the script unmodified?

mmahmoudian:
There are two reason I removed those lines:
1. I was using the AppImage to test this, so the `flameshot &` wouldn't work, also the process name is not `flameshot` anymore and for that reason the `pkill flameshot` had to be changed to `pkill -i flameshot`
2. I didn't want to mess with the `${HOME}` variable and wanted to test with in a real environment (removing junk files is actually very easy considering their creating time, but not file was generated anyways)

I didn't try the unmodified. If you cannot figure out what the issue can be, I have to put things in a clean VM, compile your branch and then run the unmodified shell script. Let me know if you think this step is necessary for testing.

veracioux:
No idea why it doesn't work. I will try the appimage myself, and then report my findings. I will also modify the test script.

You don't need have to do anything for now. Thanks for your help.

veracioux:
@mmahmoudian When I try to download the appimage version, I keep getting a 404 error. Will try again later.

But I think I know what the problem is on your end. Both the daemon and the subcommands have to be run with the same version of flameshot. You probably ran the daemon from the appimage and the subcommands via the flameshot that was installed on your system. Please run the newest test script, following the comments in it. I emphasize that you should launch the daemon manually, using the proper version of flameshot (in your case, the appimage you downloaded). Don't rely on it being launched automatically, the versions might not match.

I didn't have this problem because I use a program to automatically manage my PATH variable. I should have disabled this and made it explicit that the daemon and subcommands need to use matching versions of flameshot. Also the test script was a bit rushed and unclear so I apologize for that.

veracioux:
@mmahmoudian I have managed to download the appimage. What I have observed is that everything works except when the specified path is relative. This is because the appimage extracts itself somewhere, so the directory of the appimage is not the same as the working directory of the extracted flameshot executable that gets run.

I need advice from someone who is more knowledgeable about appimages. The way I see it currently, there is no workaround that is intuitive to the user, i.e. appimage will have to be unsupported in this feature.

Still, I would like to ask you to run the test script and confirm that you are getting the same behavior.

veracioux:
@borgmanJeremy Thanks. Should be fixed now.

borgmanJeremy:
Okay great, anything remaining before merging?

veracioux:
@borgmanJeremy Yes, I still would like to do some more testing and add a few more things possibly. I was a little busy to do it lately, but it should be ready to merge in the upcoming days.

veracioux:
I made some small improvements:
- Default path is now changed only when the GUI dialog is used, the `--path` option won't change it
- Changed the help text for `--path` which partly addresses @mmahmoudian's [comment](https://github.com/flameshot-org/flameshot/pull/1815#issuecomment-895955655). The remainder I will address in another PR.
- The image format is now chosen based on the file extension

borgmanJeremy (approved):

Issue #1559, opened 2021-04-15
Clipboard save from CLI in a GUI mode


kaldown:
I want to have the same `--clipboard` feature when making rectangle selection (`flameshot gui`) and not click on an "copy" incon every time.

mmahmoudian:
It is already implemented. Use `Flameshot gui -c`.

kaldown:
No, it's not
```
➜  ~ flameshot gui -c
the option 'c' is not a valid option for the argument 'gui'. See flameshot --help.
➜  ~ flameshot --version
Flameshot v0.9.0
Compiled with Qt 5.15.2
➜  ~
```

mmahmoudian:
I have to check it out as as far as I remember it should work, but for now you can use <kbd>Ctrl</kbd>+<kbd>c</kbd>.

mmahmoudian:
I just checked, you are right, the `-c` is not implemented for the `flameshot gui` at least according to the `flameshot gui --help`

Anu-bhav:
Hi, I was searching for this exact issue else i would have posted it myself.

@mmahmoudian hello, is this feature going to be in the next release?

mmahmoudian:
@Anu-bhav I'm not sure, but considering that it is still easily doable with existing flags and shell, I suggest you try that for the time being:
```sh
flameshot gui -r | xclip -select clip
```

(I'm typing this one the phone and might have a typo in the xclip argument. Check if with `man xclip`)

kaldown:
> @Anu-bhav I'm not sure, but considering that it is still easily doable with existing flags and shell, I suggest you try that for the time being:
> 
> ```shell
> flameshot gui -r | xclip -select clip
> ```
> 
> (I'm typing this one the phone and might have a typo in the xclip argument. Check it with `man xclip`)

it won't work as well as `-c` or clicking `copy` icon.
I.e. it won't persist in a buffer for pasting in other GUI.
Using i3wm (for instance)

kaldown:
https://github.com/chhajedji/scripts/blob/master/screenshot.sh#L31

that could be solution.
Still would be quite nice to have that option by default without moving too much of an ad-hoc solutions.

mmahmoudian:
I wrote my previous comment on my phone and I couldn't test it on the spot. The idea was just to point you at the right direction (`man xclip`).

So as the @kaldown pointed out, the command should look like:
```sh
flameshot gui -r | xclip -in -selection clipboard -target image/png
```

[...]

PR #1829, opened 2021-08-20
Add --clipboard to gui subcommand

veracioux:
Tracks issue #1559.

When you run `flameshot gui --clipboard`, any action that closes the GUI (except for pressing Escape) will copy the screenshot to the clipboard.

This patch currently has a minor problem. If you click copy to clipboard in the GUI, you will get the notification "Capture saved to clipboard" twice - once from the button and once from the `--clipboard` option. Need to figure out a good way to suppress this.

### DIFF
--- a/data/dbus/org.flameshot.Flameshot.xml
+++ b/data/dbus/org.flameshot.Flameshot.xml
@@ -5,6 +5,7 @@
       <!--
         graphicCapture:
         @path: the path where the screenshot will be saved. When the argument is empty the program will ask for a path graphically.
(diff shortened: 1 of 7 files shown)


veracioux:
Done.

borgmanJeremy:
closes #1559

borgmanJeremy (changes requested):
Can you add a help string to describe this feature? 

Should be in /src/main around line 180'ish

veracioux (commented):
There is already a help string for the --clipboard option. It is the same as for the other subcommands.

borgmanJeremy (approved):

PR #1843, opened 2021-08-24
Fall back to default shortcuts

### LINKED ISSUE #1842: Deleting the config file leaves me with no keyboard shortcuts
If I delete the config file, I am left with no keyboard shortcuts. This means that new users will have to manually add the shortcuts. I think I remember that a previous version of flameshot would fall back to default shortcuts in that case. Was that feature removed intentionally or is it a bug? It is also possible that I have introduced the bug while working on some bug fixes :)

I would like to implement the following behavior:
- If you delete the config file, the default shortcuts will be active
- Only shortcuts that the user changes (or disables) will be written to the config file

As a side note: Why is image upload excluded from the default shortcuts [here](https://github.com/flameshot-org/flameshot/blob/41ca51bc1028dd475b1bcc8954f44fda46168305/src/utils/configshortcuts.cpp#L15-L19)?

veracioux:
Closes #1842.

### DIFF
--- a/src/utils/confighandler.cpp
+++ b/src/utils/confighandler.cpp
@@ -9,7 +9,6 @@
 #include <QDir>
 #include <QFile>
 #include <QKeySequence>
-#include <QMetaEnum>
 #include <QStandardPaths>
 #include <algorithm>
 #if defined(Q_OS_MACOS)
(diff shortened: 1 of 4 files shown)


borgmanJeremy (approved):

borgmanJeremy:
Generally I prefer using enums for stuff like this instead of strings but I don't think its a big deal.

veracioux:
> Generally I prefer using enums for stuff like this instead of strings but I don't think its a big deal.

I did try to use enums but the resulting code was more complicated. Also the functions in ConfigShortcuts are used only by ConfigHandler so this solution is more fitting. I am thinking about working on a refactor related to  #1801 which might make enums the more elegant solution in this case. But not very soon.

Issue #1593, opened 2021-04-28
macOS Global Hotkey Docs


karlhorky:
**Is your feature request related to a problem?**

It would be nice to be able to use a global hotkey with macOS, just like the [Linux instructions for Ubuntu / Gnome / KDE](https://flameshot.org/key-bindings/#global).

I would take over writing a short section on this in the docs, but cannot figure out a nice (fast) way to achieve this :(  If the Flameshot screenshot doesn't launch fast, then it's not as much value as it could be.

I have tried the following:

1. Running `/Applications/flameshot.app/Contents/MacOS/flameshot gui` - this appears to do nothing
2. Trying an Automator Apple Script as follows - the script below worked, but it launched super slow - took like 5 seconds to launch

```applescript
tell application "System Events" to tell process "Flameshot"
    tell menu bar item 1 of menu bar 2
        click
        click menu item "Take Screenshot" of menu 1
    end tell
end tell
```

Credit for the Apple Script: https://stackoverflow.com/a/16497564/1268612

Anyone have any ideas how to do this?

MatanRubin:
Regarding 1, seems like it's because of #1292.

My fingers aren't flexible enough to use the default Cmd-Alt-Shift-4 global hotkey for flameshot, so I used a separate tool (BetterTouchTool, not free) to map Cmd-4 to Cmd-Alt-Shift-4, it works really well. 
You can get similar results with any other free keyboard mapper.

karlhorky:
> You can get similar results with any other free keyboard mapper

Nice! Can you elaborate on the settings that you used for this? Is it just mapping <kbd>cmd</kbd>-<kbd>4</kbd> to <kbd>cmd</kbd>-<kbd>alt</kbd>-<kbd>shift</kbd>-<kbd>4</kbd>?

MatanRubin:
😄 
I got it from the configuration menu:
![image](https://user-images.githubusercontent.com/4488227/117463630-8ea39480-af58-11eb-82ad-42a16b175e69.png)

I see that README.md has the keyboard shortcuts for Linux but not macOS, you can probably contribute there

Faetu:
Hi @karlhorky 
 
 The global shortcut, Is a mess.
 Because of this, I try to solve this the same way like you do, over automator.
 And yes, it takes ~5 sec to start the action. I play a little bit with the delay function and this Is my Solution:


```applescript
ignoring application responses
	tell application "System Events" to tell process "Flameshot"
[...]

PR #1845, opened 2021-08-26
Add macOS + Windows global hotkeys

karlhorky:
Closes https://github.com/flameshot-org/flameshot/issues/1593

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,9 @@ These shortcuts are available in GUI mode:
 
 ### Global
 
-If you want use Flameshot as a default screenshot utility, chances are you want to launch it using the <kbd>Prt Sc</kbd> key. Flameshot doesn't yet offer a fully-automated option to do so, but you can configure your system to do so.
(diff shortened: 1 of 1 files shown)


mmahmoudian:
Thanks for the contribution and it is very welcome. But your changes about Linux was so wrong that got me into rant mode, so I apologize but here it is:

On Linux because of proper and correct design of the operating system and desktop environment, user is not forced to use a keyboard shortcut and they can/should configure it themselves. As the matter of fact all keyboard shortcut complaints we have had so far in Flameshot were almost exclusively from Windows and Mac users due to their extremely limited and poorly-designed operating system.

Considering that we can never implement a code that force a shortcut to user in most Linux machines, I would consider re-phrasing the sentence.

karlhorky:
Hm, I guess I would suggest not including personal views of superiority of different operating systems in documentation of software. I wouldn't know where to start to encode this viewpoint into docs.

But if that's the opinion of the whole maintainer team, then feel free to change it - edits by maintainers to this PR and flameshot-org/flameshot-org.github.io#31 are on.

Issue #689, opened 2020-02-24
[Feature] Invert Color


GetCurious:
I'd like to invert colors with a click of a button.
it'll be super convenient for me not to switch to an editor every time.

AndydeCleyre:
Since "inversion" would be handy for at least one uncolored tool (#165), maybe a unified way of triggering it would be nice, like invert if the user is holding down shift while using the tool.

Although I guess that wouldn't be any good for tablet/keyboardless users . . .

void-m4110c:
Hi, is there any news about this feature? I really need it because my shells all have a dark background and in reports etc. it looks better if they are bright/white...
Something like "Invert Color" in Greenshot...

Thanks in advance

mmahmoudian:
@void-m4110c how about using imagemagic for the time being?

```sh
flameshot gui -r | convert png:- -invert -channel RGB png:- | xclip -selection clipboard -t image/png
```

I'm typing this on the phone and haven't tested it, but technically it should work. Just make sure you have ImageMagic and xclip installed.

void-m4110c:
Yep, this really is a viable solution, just need to map a key for that (don't want each and every screen shot to be inverted).

Thanks for the suggestion and fast reply!

mmahmoudian:
I'm glad that it helped. I myself have a small shell script that runs a `dmenu` (i think you can do the same with `rofi` as well) that it asks what type of weird screenshot I want. Normal <kbd>prntscrn</kbd> is mapped to `flameshot gui` but <kbd>Shift</kbd>+<kbd>prntscrn</kbd> opens that dmenu and asks me what type of screenshot I want. You can do the same if this inversion of colour is not something that you would need every day. This way you can use one shortcut for all your non-regular screenshots.

void-m4110c:
I'm using i3-wm and have mapped `Winkey`+`prntscrn` to `flameshot gui`, and now additionally `Winkey`+`Shift`+`prntscrn` to the given command above.

Your `dmenu` solution is also a great idea, as I use it already as a password-manager and program starter. Just need look into setting it up that way, but that shouldn't be a big deal.

void-m4110c:
Would you mind sharing your dmenu script?

mmahmoudian:
@void-m4110c I cleaned the code and put it in a repo:
https://codeberg.org/mehrad/dmenu_shot

Enjoy and feel free to make pull requests ;)

void-m4110c:
Thanks mate, really nice!

[...]

PR #1849, opened 2021-08-28
Add a tool to take an inverted screenshot

Cr4ckC4t:
## Why invert colors?

When using flameshot in dark environments on a regular basis it would be very useful to invert the colors in the capture, making the screenshot white(ish) background and dark text. Thus, enhancing the usability of those screenshots in reports and prints without having to use additional tools or workarounds.

Also, it's been a feature request for quite a while (#689).

## How does it work?

Basically, I copied the `savetool` and made an `inverttool`, that inverses all pixels before saving the screenshot. The actual consists of four lines:
```
    QPixmap inverted = context.selectedScreenshotArea();
    QImage img = inverted.toImage();
    img.invertPixels();
    inverted.convertFromImage(img);
```

However, it turned out that flameshot requires to update quite a few places in order to add a tool. Without much of a documentation I tried to stay close to how the `savetool` is setup.

## Before merging

I'd advise that someone with a way better understanding of this project looks at my modifications in order to spot potentially missing or even unnecessary code for this new tool to be integrated.

## Example

See the above mentioned issue for a video demo. Here's a screenshot of the newly added button.
invert-example

## Usage

After choosing the area and drawing markers, text, arrows etc. - use `Ctrl+i` or the *Invert* button to save the screenshot with inverted colors.

## Notes

This isn't a perfect solution. People might want to invert colors while still drawing stuff or see a review of the inversion before saving the capture. Additionally, more filters could be added maybe ([comment by @mmahmoudian](https://github.com/flameshot-org/flameshot/issues/689#issuecomment-907575403)). However, my Qt knowledge isn't by far good enough to be able to help with those things.

From my point of view, this Invert tool is a good solution for the time being and could still be extended at a later point.

---
Alright, that's it - let me know if I missed something or should update anything.


borgmanJeremy:
What do you think about changing it so a rectangular region is inverted instead of the entire section? I think that will allow it to be more flexible

Cr4ckC4t:
That's an interesting idea. For my workflow, saving the inverted screenshot is good enough but I can see how that would make the tool more flexible.

Currently I have no idea how to do that but I will try and see if I can implement your suggestion and report back here.

borgmanJeremy:
I'm happy to provide guidance if you have questions or want help! Maybe start by looking at the pixelate tool, this will be similar.

Cr4ckC4t:
Haha, that was exactly my thought. (Thanks for the offer.)

So I came up with this:

https://user-images.githubusercontent.com/63863112/131228085-8ff89150-3278-40a9-b2ad-7bebe469bc3e.mp4

Does that match your idea?

It's basically just a stripped down version of pixelate with inversion instead of bluring.

borgmanJeremy:
Perfect!

Cr4ckC4t:
So I'm unsure about one thing. The pixelate tool used a scaled selection by querying the pixel-aspect ratio of the screen before copying the screen area. However, the normal rectangle tool and selection tool don't do that. I've left that out for now and used the basic selection area (`point().first` to `point().second`) which seems to work just fine. I don't know if the scaled selection maybe is necessary for different screen resolutions or if that was just for pixelating- so maybe check that (`inverttool.cpp`).

Other than that, I've updated the butten order of the invert tool (next to pixelate). While doing that I also noticed that the numbering tool is probably supposed to be with the other drawing tools as well, so I positioned it next to the inverter too (let me know if that's not wanted).

To match the other drawing tools I also updated the description and the shortcut (now just `i`).

Here's what it looks like now:
invert-example

src/tools/invert/inverttool.cpp
```diff
@@ -0,0 +1,84 @@
...
+    QImage img = inv.toImage();
+    img.invertPixels();
+
+    painter.drawImage(selection, img);
+}
+
+void InvertTool::drawSearchArea(QPainter& painter, const QPixmap& pixmap)
+{
+    Q_UNUSED(pixmap)
+    Q_UNUSED(painter)
```
borgmanJeremy, line 62: Reference the implementation in the Pixelate tool for this. Currently since you marked Q_UNUSED(painter), this new feature does not tie into the "movable" feature. IE I cannot move the selection once placed. 

If you duplicate the implementation that is in the pixelate feature that will resolve this.

borgmanJeremy (changes requested):

data/translations/Internationalization_ca.ts
```diff
@@ -1303,6 +1303,19 @@ You may need to escape the &apos;#&apos; sign as in &apos;\#FFF&apos;</source>
         <translation>Guarda la captura</translation>
     </message>
 </context>
+<context>
```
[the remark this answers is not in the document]
Cr4ckC4t: ✅ Changes reverted in `973aef74` and `de41c958`.

src/tools/invert/inverttool.cpp
Cr4ckC4t: ✅ Done in `9889eadf`, although I didn't fully understand. I was able to move the screen selection without those lines too. 

Regardless, it's now the same as in the `pixelate tool`.

data/img/material/black/invert.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" aria-hidden="true" focusable="false" role="img" width="1025" height="1024" preserveAspectRatio="xMidYMid meet" style="transform: rotate(360deg);"><rect id="backgroundrect" width="100%" height="100%" x="0" y="0" fill="none" stroke="none"/><g class="currentLayer" style=""><title>Layer 1</title><path d="M896.428 1024h-768q-53 0-90.5-37.5T.428 896V128q0-53 37.5-90.5t90.5-37.5h768q53 0 90.5 37.5t37.5 90.5v768q0 53-37.5 90.5t-90.5 37.5zm0-768q0-53-38-91l-120 121q-44-45-102.5-69.5t-123.5-24.5q-87 0-160.5 43t-116.5 116.5t-43 160.5q0 65 24.5 123.5t69.5 102.5l-121 121q38 37 91 37h512q53 0 90.5-37.5t37.5-90.5V256zm-384 576q-133 0-226-94l452-452q45 44 69.5 102.5t24.5 123.5q0 87-43 160.5T672.928 789t-160.5 43z" fill="#000000" id="svg_1" class="selected" fill-opacity="1"/></g></svg>
```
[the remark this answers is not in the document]
Cr4ckC4t: I did not. I don't actually remember the site I got it from.

Instead, I've seen that flameshot already uses Material Design icons so I updated the icon to [this one](https://materialdesignicons.com/icon/invert-colors) (black and white). The current license file covers this already if I'm not mistaken.

✅ `c420842b`

src/tools/invert/inverttool.cpp
borgmanJeremy: Thanks, what I meant was I could not move the inverted box the way other objects are moveable. This should fix it.

Creteil:
Hi all,

Sorry for my ignorance, but from where I can fetch this source code to have access to this latest features commits ?

Creteil:
Look like it must appear after « change request » approval in the master branch...

True ?

If so, sorry to bogus on the question...

Cr4ckC4t:
@Creteil you could clone my fork from here: https://github.com/Cr4ckC4t/flameshot/tree/invert (branch: `invert`). Building this flameshot allows you to use the `invert` tool.

However, maybe this gets merched soon (thanks to @borgmanJeremy for the quick responses and reviews) - so I'd say keep an eye on the original flameshot:master.

borgmanJeremy (approved):
Looks great! Thanks for the contribution!

Issue #1740, opened 2021-07-22
The config parser should throw error if there are issues


mmahmoudian:
## The issue

We have got multiple issues reported that Flameshot does not behave as expected and it has been narrowed down to mistakes in config file. the most recent issue was #1739.

## Suggestion

The part of the code that reads and parses the config file should:
1. throw error (in notification, in the log, and also as a layer to when the `flameshot gui` opens)
2. falls back to some defaults until the next trigger of reading the config

veracioux:
I propose that we check the config file:
- at each launch of the daemon
- each time the config file changes a la `QFileSystemWatcher`

These are all the checks I could think of (I invite you to contribute more):
# Syntax checks are handled by `QSettings` already
Errors are indicated by a status flag.
# Perform semantic checks on the config data
Are numbers, dates, file extensions, etc correctly formatted.
# Check if there are unrecognized entries
If we add this check, I think it should be on demand by the user. Extra
entries in the config are not a big problem. The only thing they can really
do is clog up the config. But even this is unrealistic to expect.
# Check if there are multiple instances of an entry
I don't know how we would implement this tbh.

We should notify the user that there was an error using the means that
@mmahmoudian mentioned and then attempt to fix the error.

- Should we ask for confirmation before fixing it?
- Should we always show what the error was before fixing it?
- What errors can be fixed at all?

I have seen that spdlog is a dependency, but haven't seen anything useful
done with it. Do you use it for anything yet?

mmahmoudian:
@veracioux 

> I propose that we check the config file:
> - at each launch of the daemon
> - each time the config file changes a la `QFileSystemWatcher`

I agree with your first bullet point, but the second one is overkill imho as [there will be unnecessary overhead load](https://doc.qt.io/qt-5/qfilesystemwatcher.html#details). I think a better alternative would be to run it whenever the daemon is about to perform an action  (basically all subcommands of `flameshot`, plus the launcher)

> **Perform semantic checks on the config data**
> Are numbers, dates, file extensions, etc correctly formatted.

As a related, but off-topic comment: I think it would be nice to add comment lines above each item in the `flameshot.ini` to guide the user.

> **Check if there are unrecognized entries**
> If we add this check, I think it should be on demand by the user.

I didn't understand this.

> **Check if there are multiple instances of an entry**
> I don't know how we would implement this tbh.

Disclaimer: I don't know C++ and QT good enough to have solid suggestions.
[...]

PR #1859, opened 2021-08-31
Implement config checking

veracioux:
Tracks issue #1740.
@borgmanJeremy I have completely redesigned `ConfigHandler` and added a class `ValueHandler` (`src/utils/valuehandler.{h,cpp}`). I think it is more maintainable this way, and works better with the config checking mechanism.
In essence, those are the only substantial changes I made. All the other files are mostly a result of changing the names of the config value getters/setters in `ConfigHandler`.

NOTE: This is a draft, but will leave it open for checks to run on it.

### TODO Checking API
- [X] Unrecognized config options
- [X] Shortcut conflicts (multiple shortcuts with the same binding)
#### - Semantic checks
Whether the setting values have the correct format (number, string, bool, etc.)
- [X] boolean
- [X] color
- [X] color list
- [X] existing directory
- [X] button list (`buttons` option)
- [X] bounded integers
- [X] filename pattern
- [x] key shortcuts
### TODO Checking procedure
- [X] On daemon launch
- [X] Config changed while daemon running
       Consider another point from mmahmoudian which I don't really understand.
### TODO Error reporting
- [X] Notification
- [X] Log file
       For this one I added a TODO comment. We should wait for logging to be implemented properly, and then implement config error logging in another PR.
- [X] GUI overlay
       @mmahmoudian and I agreed on a message box before a capture but I thought your original proposal is a better solution. The message appears in the bottom left corner and it automatically disappears if the mouse is over it and a tool is active.
- [X] Config menu indicator
       @mmahmoudian's proposal for a check again button is unnecessary because the
       config is watched for changes
- [x] Reporting the exact error

### TODO Additional things

- [x] remove obsolete config entries
- [x] fall back to paths `~/Pictures`, `~`, `/tmp` in that orders
## Testing
Please test the error checking by introducing errors in the config for various config options and value types. You should get notified every time the error state changes.
- Whenever you introduce an error in the config, the change should be detected and get notified. You will get the notification only the first time an error occurs.
- If you resolve the error, you will get notified as well
- Now if you introduce an error again, you will get notified. And so on...
- If you have another monitor like me, you might be able to edit the config file while the the capture GUI is active to see what happens with the bottom left message. You should get a message that says that the config was resolved. If that isn't possible, then you could create two files - one correct and one erroneous - and a script that applies them to the active config with a timer.
- A special case is when you delete the config file. I had to bypass Qt's `QFileSystemWatcher` behavior here, so this case should be more thoroughly tested.
- I think this should be tested on multiple platforms just in case. I got Linux covered, and will eventually do Windows too.


src/utils/confighandler.cpp
```diff
@@ -4,380 +4,175 @@
...
+    OPTION("drawThickness"               ,LowerBoundedInt(1  , 3             )),
+    OPTION("drawColor"                   ,Color              ( Qt::red       )),
+    OPTION("userColors"                  ,UserColors         (               )),
+    OPTION("drawFontSize"                ,LowerBoundedInt(1  , 8             )),
+    OPTION("ignoreUpdateToVersion"       ,String             ( ""            )),
+    OPTION("keepOpenAppLauncher"         ,Bool               ( false         )),
+    OPTION("fontFamily"                  ,String             ( ""            )),
+    OPTION("setSaveAsFileExtension"      ,String             ( ""            )),
+      // clang-format on
+  };
```
veracioux, line 115: Please review this even if you are not a programmer to see if I have messed up some default values.

mmahmoudian:
@veracioux I haven't got to the end of it, but while reading "TODO Checking API" I realized we should also check if same shortcut is defined multiple times (regardless of having the same binding or not). For example if we have the following case in the `flameshot.ini`.

> ```
> ...
> [Shortcuts]
> TYPE_ARROW=A
> TYPE_ARROW=B
> TYPE_ARROW=A
> ...
> ```

veracioux:
@mmahmoudian Addressed that in **Notes about QSettings**.

src/utils/confighandler.cpp
veracioux: Great idea!

veracioux:
@mmahmoudian Sorry about the failing builds. Unless you use ubuntu 18.04, you should be good to test the program now. I'm going to have to setup a VM to resolve issues on 18.04.

mmahmoudian:
I downloaded the [AppImage](https://github.com/flameshot-org/flameshot/pull/1859/checks?check_run_id=3475343376) which was built against 90541b441deeecba2e60648ff684a301f1d0e80d and it says that my config has error:

image

I don't understand what is the issue, nothing is reported to the stdout and there is no log file:

image


<details>
<summary>Here is my config file</summary>

> ```
> [General]
> contrastOpacity=188
> disabledTrayIcon=false
> drawColor=#800080
> drawFontSize=10
> drawThickness=6
> saveAfterCopyPath=/home/mehrad/Pictures
> savePath=/home/mehrad/
> savePathFixed=false
> setSaveAsFileExtension=.png
> showStartupLaunchMessage=true
> startupLaunch=true
> uiColor=#ff7300
> undoLimit=100
> uploadHistoryMax=25
> useJpgForClipboard=false
> 
> [Shortcuts]
> TYPE_ARROW=A
> TYPE_CIRCLE=
> TYPE_CIRCLECOUNT=
> TYPE_COMMIT_CURRENT_TOOL=Ctrl+Return
> TYPE_COPY=Ctrl+C
> TYPE_DELETE_CURRENT_TOOL=Del
> TYPE_DRAWER=D
> TYPE_EXIT=Ctrl+Q
> TYPE_IMAGEUPLOADER=
> TYPE_MARKER=M
> TYPE_MOVESELECTION=Ctrl+M
> TYPE_MOVE_DOWN=Down
> TYPE_MOVE_LEFT=Left
> TYPE_MOVE_RIGHT=Right
> TYPE_MOVE_UP=Up
> TYPE_OPEN_APP=Ctrl+O
> TYPE_PENCIL=
> TYPE_PIN='
> TYPE_PIXELATE=B
> TYPE_RECTANGLE=R
> TYPE_REDO=Ctrl+Shift+Z
> TYPE_RESIZE_DOWN=Shift+Down
> TYPE_RESIZE_LEFT=Shift+Left
> TYPE_RESIZE_RIGHT=Shift+Right
> TYPE_RESIZE_UP=Shift+Up
> TYPE_SAVE=Ctrl+S
> TYPE_SELECTION=S
> TYPE_SELECTIONINDICATOR=
> TYPE_SELECT_ALL=Ctrl+A
> TYPE_TEXT=T
> TYPE_TOGGLE_PANEL=Space
> TYPE_UNDO=Ctrl+Z
> ```

</details>

veracioux:
@mmahmoudian There was a bug, should be fixed now. If you get an error again, check if the paths in your config are existing directories.

borgmanJeremy:
I have an issue if I try to remap a key via the gui: 
Peek 2021-09-01 06-59

veracioux:
@borgmanJeremy Are you compiling in debug mode? If you are not, could you do that and tell me if you are getting an exception when trying to change a shortcut?

borgmanJeremy:
@veracioux  Since there is a bunch of discussion on this topic I thought we could try github "discussions" I posted my reply here: https://github.com/flameshot-org/flameshot/discussions/1862

mmahmoudian:
I build from source against b6641ea commit and:

- having `drawFontSize=A` didn't raise an error
- having `uploadHistoryMax=2.5` (with decimal points) would automagically correct it to `uploadHistoryMax=0` which should not become 0
- having`TYPE_RESIZE_RIGHT=Shift+Right:` (extra colon at the end) would not generate error, it is not automagically fixed and the shortcut does not work anymore. even the colon shows up in the config window:
image

veracioux:
Alright, new stuff:
- Added directory fallback scheme as we agreed: `Pictures`, `$HOME`, `/tmp`
- Added a `flameshot config --check` option that reports errors in the config
- Added the GUI version of the above - just introduce an error, open the config window and click the "Show errors" button.

mmahmoudian:
Looks great 👍🏼  . The parts that got me really excited were:
1. The CLI that gives nice explanation (although it can be improved but for the time being it is definitely good)
2. The notification that states that the problem is successfully resolved.

⚠️ During my testing there was one type of error that it didn't capture. it considered `#ff7300aaa` a valid color which as far as I know a hex color code should have either of 3,6, or 8 hexadecimal characters.

veracioux:
@mmahmoudian I just used Qt's color parser. Turns out that it accepts `#RRRGGGBBB` and `#RRRRGGGGBBBB` as well. I'll manually mark these as invalid. And also they use `#AARRGGBB` instead of `#RRGGBBAA`. Just to verify with you - we want to use the latter, right?

mmahmoudian:
> Turns out that it accepts #RRRGGGBBB and #RRRRGGGGBBBB as well. 

🤯  I didn't know that. if QT really accept them and they can turn into a color and they do not break Flameshot, I guess we can keep it in.

mmahmoudian:
@veracioux Is it possible to as the QT's color parser to only accept #RRGGBBAA ?

veracioux:
> What is your preference?

I agree, I don't expect it to cause any issues with the users. And it's less work for me :)

> Is it possible to as the QT's color parser to only accept #RRGGBBAA ?
I've implemented that case manually.

veracioux:
@mmahmoudian @borgmanJeremy 
I think I have implemented everything we discussed, except for file logging. I would like to leave that for another time, when we figure out a proper application interface (I don't want to implement a half-cocked solution only to have to change it later). I have been working on some ideas to create a common interface for sending notifications, CLI messages (stdout/err) and file logging, and will be sending a PR for that sometime.

Please gloss over the TODO entries at the top and let me know if I missed something.

@mmahmoudian If you have ideas on how to improve the `config --check` option, please share.

I say we do some more testing, and then release it into the wild and see how it performs :wink:

veracioux:
> @veracioux Is it possible to as the QT's color parser to only accept #RRGGBBAA ?
> 
> This is because then the behavior of Flameshot becomes deterministic and we can handle bug-reports easier

Wait, you mean we should only accept `#RRGGBBAA` out of all possible ways to specify the color? In other words, drop `#RGB`, `RRGGBB`, and the others?

mmahmoudian:
@veracioux I want what ever that is easiest for you now and the least confusing for the user and maintainers in the long run. I have no preference here. Basically I'm totally fine with whatever QT accepts. But we should probably include a link to the QT color parser documentation of the user has a wrong setting.

What is your preference?

veracioux:
@mmahmoudian I say we allow #RRGGBB, #RRGGBBAA and by color name (e.g. red). And ban the other formats. These are the most common, most intuitive and they cover basically every use case. I wouldn't want to redirect the user to the Qt doc in any circumstance.

veracioux:
@mmahmoudian What else is left to do?

mmahmoudian:
Imo this is very much ready 👍🏼
I have one suggestion but that might be too much trouble as it is just for aesthetic, so I just leave it here:

The `flameshot config -check` can return the error in a colored form. For example the name of the key in yellow and the error after the : in red.

mmahmoudian:
Also, is it possible to have comment in the ini file? For example leading the line with # makes that line comment?

veracioux:
> The `flameshot config -check` can return the error in a colored form. For example the name of the key in yellow and the error after the : in red.

Because I use the same code for both CLI and GUI error reporting, I can't do that without some trouble. Basically: elegant CLI interface => inelegant code in this case.

> Also, is it possible to have comment in the ini file? For example leading the line with # makes that line comment?

Problem is that QSettings removes comments when it writes settings to the file. I think it's enough to have the example config which is well documented as it is. Although I think it would be good to install the example config in a standard location like `/usr/share/flameshot` or `/usr/share/doc/flameshot` (on linux at least). I'm not volunteering to do that since I'm not currently in the mood to dig into makefiles and the like.

veracioux:
Alright, then we are waiting for your final word @borgmanJeremy

borgmanJeremy:
Thanks for the hardwork! merging it now.

mmahmoudian:
Thanks @veracioux for working on this 👍🏼 🚀

veracioux:
@borgmanJeremy @mmahmoudian Thank you! It was an absolute pleasure.

PR #1869, opened 2021-09-04
Color grabber zoom preview

### LINKED ISSUE #1808: Feature: Color grabber preview
There are many times when I want to grab a color from the screen and the region with the color is quite small (e.g. a color from text). I would like to implement a small widget that shows a zoomed-in region around the color grabber when the mouse is being held down. So everything about the color grabber would work the same except for when the mouse is held down.

Maybe it would be good to add a timer that waits a certain time after pressing the mouse before the widget appears.

veracioux:
Closes #1808.

- Added ColorGrabWidget class
- Refactored SidePanelWidget to use ColorGrabWidget

Note: The PR is still a draft but will keep it open for testing.

### DIFF
--- a/src/widgets/panel/colorgrabwidget.cpp
+++ b/src/widgets/panel/colorgrabwidget.cpp
@@ -27,6 +27,7 @@ ColorGrabWidget::ColorGrabWidget(QPixmap* p, QWidget* parent)
   : QWidget(parent)
   , m_pixmap(p)
   , m_mousePressReceived(false)
+  , m_extraZoomActivated(false)
 {
     if (p == nullptr) {
         throw std::logic_error("Pixmap must not be null");
@@ -50,7 +51,7 @@ void ColorGrabWidget::startGrabbing()
     OverlayMessage::push(
       "Press Enter or Left Mouse Button to accept color\n"
       "Press and hold Left Mouse Button to precisely select color\n"
-      "Press Space to toggle magnifier\n"
+      "Press Space or Right Mouse Button to toggle magnifier\n"
       "Press ESC to cancel");
 }
 
@@ -73,13 +74,13 @@ bool ColorGrabWidget::eventFilter(QObject*, QEvent* event)
         } else if (key == Qt::Key_Return || key == Qt::Key_Enter) {
             emit colorGrabbed(m_color);
             finalize();
-        } else if (key == Qt::Key_Space && !m_mousePressReceived) {
+        } else if (key == Qt::Key_Space && !m_extraZoomActivated) {
             setVisible(!isVisible());
         }
         return true;
     } else if (event->type() == QEvent::MouseMove) {
         // NOTE: This relies on the fact that CaptureWidget tracks mouse moves
-        if (!m_mousePressReceived) {
+        if (!m_extraZoomActivated) {
             // Update only before the user clicks the mouse, after the mouse
             // press the widget remains static.
             updateWidget();
@@ -95,8 +96,13 @@ bool ColorGrabWidget::eventFilter(QObject*, QEvent* event)
     } else if (event->type() == QEvent::MouseButtonPress) {
         m_mousePressReceived = true;
         auto* e = static_cast<QMouseEvent*>(event);
-        if (e->buttons() == Qt::LeftButton && !isVisible()) {
-            QTimer::singleShot(500, this, [this]() { show(); });
+        if (e->buttons() == Qt::RightButton) {
+            setVisible(!isVisible());
+        } else if (e->buttons() == Qt::LeftButton) {
+            m_extraZoomActivated = true;
+            if (!isVisible()) {
+                QTimer::singleShot(500, this, [this]() { show(); });
+            }
         }
         updateWidget();
         return true;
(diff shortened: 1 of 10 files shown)


mmahmoudian:
I tried the AppImage built against 5cea60398127e85e29930809e7549770198b87f8 and it worked just fine on multi-monitor (extended and Nvidia).

I just have a suggestion but I don't know how hard it is to pull off: when click and hold and while user is waiting for the zoom window to open, it would be nice to load a small pie timer to tell user that something is happening. Since this is only loaded when user clicks on "Grab color" button, it would not be too much of a bloat.

image

veracioux:
@mmahmoudian I don't see how that would help with discoverability. Because users will normally just want to click and release immediately. They won't have time to see the icon, they would already have grabbed the color.

Perhaps it would be better to replace the "Press ESC to cancel" text of the grab button with a help overlay similar to the one that appears when the GUI displays initially.

veracioux:
@mmahmoudian I have added the hex color label under the color wheel. You can use
it to edit the color too. You can even write a standard color name like "red"
and it will be converted to #RRGGBB automatically.

I also made it so that a right click shows the widget instantly, and left click with a delay, like before. This is friendlier for touchpad users (of which I am one).

borgmanJeremy:
What you you all think of making it a toggle button. Once it is enabled the magnified widget will follow the cursor. This is how greenshot on windows works.

veracioux:
@borgmanJeremy I'm sorry about my hiatus. I've been really busy lately. I've made some edits and tweaks, please let me know what you think. I have implemented what you asked for if I understood you correctly. I am quite satisfied with the overall result.

As a bonus, I have added an overlay that shows the user what options they have available for the color grabbing. In the process, I have added a common class `OverlayMessage` and made `CaptureWidget` use that implementation instead of the old one.

mmahmoudian:
@veracioux the overlaymessage is not positioned correctly. it is in the middle of all visible areas, meaning when we tile all the monitors and create a large canvas, the message will be in the middle and center  of this large canvas and not the "active" monitor where the mouse is in.

I also found it at first annoying that the zoomed square stays in place even when I move my mouse away. But then I think I found an unintended behavior: while holding down the left-click and have the zoom open, I can move my mouse around to other parts of monitor and then right-click or middle-click and cause the position of the zoom square to get updated and jump to my mouse pointer's position.

I also quite liked @borgmanJeremy 's idea about button. For example holding down Ctrl can bring this zoomed version. This would make is easier for touchpad users but makes it harder for tablet users. So I suggest keeping both "left-click&hold", and "Ctrl&hold".

veracioux:
@mmahmoudian
I will look into the bugs and unintended behaviors you mentioned.

> I also found it at first annoying that the zoomed square stays in place even when I move my mouse away.

The point of the zoom widget remaining static after mouse press is so that you can precisely select the color. This way for every real device pixel you move your mouse, you move only one fifteenth of a pixel inside the zoom widget. A zoom widget that always follows your mouse around allows you to see the area around the pixel better, but it doesn't help you select the pixel more precisely.

Also, from your comment I am not sure if you are aware that you can press `Space` to display the zoom widget and it will follow your mouse around.

> So I suggest keeping both "left-click&hold", and "Ctrl&hold".

Can you please explain what exactly you want these shortcuts to do? I'm not sure I understand.

src/widgets/panel/colorgrabwidget.cpp
```diff
@@ -0,0 +1,184 @@
...
+#include <stdexcept>
+
+// Width (= height) and zoom level of the widget before the user clicks
+#define WIDTH1 80
+#define ZOOM1 8
+// Width (= height) and zoom level of the widget after the user clicks
+#define WIDTH2 165
+#define ZOOM2 15
+
+// NOTE: WIDTH1(2) should be divisible by ZOOM1(2) for best precision.
```
veracioux, line 24: @mmahmoudian Btw, you can tweak these if you feel so inclined.

mmahmoudian:
@veracioux 

> The point of the zoom widget remaining static after mouse press is so that you can precisely select the color. This way for every real device pixel you move your mouse, you move only one fifteenth of a pixel inside the zoom widget. A zoom widget that always follows your mouse around allows you to see the area around the pixel better, but it doesn't help you select the pixel more precisely.

Perhaps I could have been a bit more constructive in my criticism 😅 sorry. I somehow expected that the zoom window disappear when the mouse leaves it's area, but I guess releasing the left-click is intuitive enough and my expectation was wrong.

> Also, from your comment I am not sure if you are aware that you can press Space to display the zoom widget and it will follow your mouse around.

No I again didn't rtfm 🤦🏼‍♂️ 😅  This is great. I love this feature that comes with the space.


I can also propose a simple change: right-click would also toggles this zoom widget that follows the mouse. The current behavior of showing the color wheel is useless while we want to grab a color from screen), and leaving the area of the zoom closes the zoom window.

src/widgets/panel/colorgrabwidget.cpp
mmahmoudian: I think these are good enough. I don't have a 4k display though, so I don't know how small they look in such monitor.

borgmanJeremy:
> @borgmanJeremy I'm sorry about my hiatus. I've been really busy lately. I've made some edits and tweaks, please let me know what you think. I have implemented what you asked for if I understood you correctly. I am quite satisfied with the overall result.

No problem, I also get quite busy and need to take breaks.

I tested the latest version and it seems pretty great!

veracioux:
@mmahmoudian Please check if the latest changes fixed the issues you mentioned, and tell me if you are satisfied with the new features you asked for.

@borgmanJeremy As far as I'm concerned this is it. I'll gladly add any suggestions you two might have, but personally I have nothing to add.

borgmanJeremy:
I think its ready to merge. Can you fix the conflicts and then I'll merge this next?

veracioux:
@borgmanJeremy Will do. But we should wait for @mmahmoudian's response before merging.

borgmanJeremy:
Yup good call. I was just skimming the comments and didn't see we were waiting for his reply.

mmahmoudian:
Perfection

Issue #1216, opened 2021-01-03
[Feature Request] Support RTL text 


elad2109:
For hebrew and arabic.

borgmanJeremy:
Would you expect this to happen automatically for particular locales or languages? Or do you and @mmahmoudian think this should be a dedicated config option for all languages. 

Is there a gif or video of this working well in another program? It's not something I am familiar with as an English speaker but I 100% support adding this.

mmahmoudian:
@borgmanJeremy  Perian is my mother tongue but I'd never tried it in Flameshot before. I kinda trusted that the RTL is not supported as usually that's the case in most under development softwares, but I just checked and it worked perfectly:

![flameshot_rtl](https://user-images.githubusercontent.com/390889/103573948-c41b7100-4ed7-11eb-90f4-2aa94560a7b3.png)

I guess the case is close. If @elad2109 you have anything else to add please feel free and re-open this issue.

elad2109:
Try to put the cursor somewhere.
Write in RTL, one letter.

Exit text mode. See how the letter moved to a whole different place.
One has to guess the final position of the text.

mmahmoudian:
I confirm the bug, but it has nothing to do with RTL support as also you can see that the position of "g" also changes when I click elsewhere. My hypothesis is that Flameshot by default creates a borderless textbox and when we type in an RTL language, it automatically get switched to right-alignment and when the text box goes out of focus, it's size gets shrinked to the text width, and this would move the text to the left:
https://user-images.githubusercontent.com/390889/104309733-4b21a800-54db-11eb-9062-414c8800ff28.mp4

elad2109:
My bad, I just use the tool so much that I thought it would show very easily my intention.
Plus I'm always happy to add more details, feel free to ask if still not clear.

Also, multi lines text alignment is not to the right.
![image](https://user-images.githubusercontent.com/1093841/104315818-3564b080-54e4-11eb-83a5-c4693e53b074.png)

mmahmoudian:
I think the multiline issue is because of the same textbox behaviour. When the alignment is practically lost when the textbox goes out of focus and for that reason the textbox get collapsed to the left and that moves the single-character texts.

elad2109:
Can you provide a supportive screenshot?

![image](https://user-images.githubusercontent.com/1093841/104317387-92616600-54e6-11eb-8b08-315fb06335b6.png)

borgmanJeremy:
[...]

PR #1892, opened 2021-09-13
Added text alignment

borgmanJeremy:
Added the ability to adjust text alignment.


borgmanJeremy:
Example
alignment

borgmanJeremy:
@mmahmoudian  This is for #1216

elad2109:
Beautiful!

mmahmoudian:
@borgmanJeremy As far as I can tell, this by default works fine but the moment user wants to edit the text it gets confusing:


https://user-images.githubusercontent.com/390889/133245766-f9b1ea81-bea7-473d-9a6d-9e88d05eec55.mp4

borgmanJeremy:
Oh wow that's really interesting! I think I'll file this upstream with Qt because its really counter intuitive. *left* align should always mean *left* no matter the font.

borgmanJeremy:
Dropping this comment here for future reference.....might be possible to detect actually: https://doc.qt.io/qt-5/qt.html#LayoutDirection-enum

PR #1918, opened 2021-09-23
Simple maintenance improvements

a1346054:


borgmanJeremy:
Thanks. Are any of these changes something should be forced in the CI like we do for clang-format

borgmanJeremy (approved):

PR #1933, opened 2021-10-02
Unify the badge style in README.md

PeterDaveHello:
There are two styles of badges in the README.md, which is inconsistent.

I found that the style change made in #339, which is a README clean up pull request, seemed to be an accident, as the clean up works should not change the style things.

If the "flat-square" style is preferred, I can also see all the badges can use it, but the default one is simple and easy, and I didn't see a PR/commit indented to change the style since the beginning, so I think the default would be better.


borgmanJeremy:
Great thanks!

Issue #1901, opened 2021-09-16
Painful performance with multiple high resolution displays


syeopite:
**Flameshot version**

Flameshot v0.10.1
Compiled with Qt 5.15.2

**Describe the bug**

Flameshot has horrible performance with multiple high resolution displays. For reference, I have four 4k monitors powered by an RTX 2080 TI — so the hardware shouldn't be causing this: 

https://user-images.githubusercontent.com/70992037/133621874-005a2621-7ecd-48fd-abec-2ec126d17d90.mp4

If it helps this was exactly the performance I had with Gpick, but did not with Gcolor3. 

**To Reproduce**

1. Have a computer connected with multiple high resolution displays 
2. Start flameshot
3. Select a region and try to drag it around

**Expected behavior**

For flameshot to have fluid performance.

**System Information**

```
OS: Arch Linux x86_64 
Kernel: 5.13.13-arch1-1 
Uptime: 11 days, 9 hours, 3 mins 
Packages: 1164 (pacman) 
Shell: zsh 5.8 
Resolution: 3840x2160, 3840x2160, 3840x2160, 3840x2160 
DE: Budgie 
WM: Mutter(Budgie) 
Theme: Materia-dark [GTK2/3] 
Icons: Papirus [GTK2/3] 
Terminal: kitty 
CPU: AMD Ryzen Threadripper 2990WX (64) @ 3.000GHz 
GPU: NVIDIA GeForce RTX 2080 Ti Rev. A 
GPU: NVIDIA GeForce RTX 2080 Ti Rev. A
```

borgmanJeremy:
Are you by any chance a c++ developer? It would be great to get some profiler information as I don't have the hardware to replicate this.

syeopite:
Unfortunately not, my skills are limited to crystal and python. Though if you could provide some instructions, I'll be happy to try.

borgmanJeremy:
Here is a script that will build flameshot and run it under a profiler for 30 seconds. You will need to install the flameshot build dependencies first (from the README).

You can see it runs for 30 seconds before terminating but feel free to tweak that. When its done can you upload two files?

1. tmp/Flamegraph/out.perf-folded
2. tmp/Flamegraph/perf.svg

This might not capture the issue but its a good starting point. You can open the SVG in firefox if you want to see what it looks like: 
![image](https://user-images.githubusercontent.com/46930769/133667558-92bb710b-7272-412d-98c8-8ce8b585e304.png)

```
[...]

PR #1939, opened 2021-10-04
Capture GUI refactor

veracioux:
Changes:
- Removed CaptureToolButton::ButtonType and use CaptureTool::Type (previously ToolType) exclusively
- Moved selection widget code from CaptureWidget to SelectionWidget. This makes
for a leaner CaptureWidget and therefore more maintainable. Also, since
SelectionWidget handles its own paintEvent, selection performance is increased
- Tweak all CaptureWidget::update() calls so that only the minimally required
regions are affected. This has boosted performance throughout the GUI.
- Improved some aspects of the selection widget and move tool along the way

I started this off as a maintainability refactor, but then I fiddled around with
the SelectionWidget and realized I could improve the GUI performance. On my
device the improvement is not drastic but it's definitely noticeable. I think
that on larger/multiple monitors the performance increase might scale a lot. This
could help with #1901


borgmanJeremy:
This is looking good! I'm still reviewing the code but all my local testing looks good.

PR #1940, opened 2021-10-04
Reduce max thickness

mgalgs:
The thickness slider is fairly small, so it's difficult to make small
adjustments. The max value of the thickness slider is currently 100, which
is actually pretty huge.

Since most common use cases involve smaller pixel values (closer to the
0-10 range, I suspect), reduce the maximum slider value to 25, allowing for
easier fine grained control of the drawing thickness.

Although fine grained control is already possible via the scroll wheel,
this change improves the overall usability of the tool, especially for new
users who many not know that thickness can be adjusted with the scroll
wheel (or for users who don't have a scroll wheel).

### Screenshots

Maxed out thickness, before this change:

flameshot_max_active_thickness_before

Maxed out thickness, after this change:

flameshot_max_active_thickness_after


borgmanJeremy:
@mmahmoudian this looks good to me. Any concerns?

borgmanJeremy (approved):

Issue #1531, opened 2021-04-02
Remove irrelevant buttons when CLI arguments are used


mmahmoudian:
When user uses `-r`, `-c`, or `-p` then having the buttons related to other functions are confusing and misleading.

The first time this confused me is when I ran `flameshot gui -c` and I didn't know what button I should press. This is even worse when using the `-r`.

My suggestion is regardless if which argument, uniformly removed the "copy", "save", and "upload" buttons and instead show a "capture" button (for example with a camera icon).

Alternatively we can hide only the irrelevant buttons. This would be most ideal but the implementation might not be as easy as the previous.

mmahmoudian:
I just realized that `flameshot gui -r` would also overwrite the clipboard with the selected area AND sensing to stdout. This is confusing and also causes data loss of clipboard despite the `-r`.

I believe implementing this new "capture" button is super necessary.

mmahmoudian:
@veracioux tbh I personally prefer the KDE shortcuts to run commands than to be a DBus call, simply because of issues similar to #1875 are less likely to happen, plus the shortcuts can be a good source for learning subcommands.

Thanks for the efforts🦾

I'll make the PR and will let you know.

PR #1953, opened 2021-10-08
Kde shortcuts dbus2cmd

mmahmoudian:
Based on @veracioux [suggestion](https://github.com/flameshot-org/flameshot/issues/1531#issuecomment-938254485) for #1531, I have changed the DBus calls to commands in KDE shortcuts. I also realized that the "proper" file extension is ".khotkeys" and I also added that.


borgmanJeremy:
Will this break shortcuts on other DE's?

mmahmoudian:
@borgmanJeremy considering that this is not inside the installation process or in Flameshot code, but rather an additional extra step user should do, this PR should not break anything for anyone (even users with v0.8.x should also be fine with it. The same perhaps cannot be said about the PR @veracioux is cooking.

PR #1955, opened 2021-10-08
Expand role of CaptureRequest and add Accept tool

### LINKED ISSUE #1336: Allow flameshot gui to work with --clipboard
**Is your feature request related to a problem?**

It would be nice if I could use `flameshot gui` to automatically copy to the clipboard like `flameshot screen -c` can do.

It's only one more click to copy to the clipboard but command consistency seems weird to support it on some subcommands and not others.

### LINKED ISSUE #1416: Enter key no longer submits the selection in raw mode
**Flameshot version**

v0.9.0

**Describe the bug** 

The enter key no longer works for submitting a selection in raw mode (`flameshot gui -r`).

**To Reproduce**

Run `flameshot gui -r`, select an area and press the enter key. Nothing happens.

**Expected behavior**

Flameshot exits when pressing the enter key and prints the image to `stdout`.

**System Information**

Arch Linux, KDE Plasma on X11

### LINKED ISSUE #1531: Remove irrelevant buttons when CLI arguments are used

### LINKED ISSUE #1813: Revamp the helpTxt content
Considering that the Enter key is not by default bounded to anything and also considering that the shortcut keys are now customizable by user, I believe it would be good if the helpTxt (in the center of the image below) is changed to accommodate the following suggestions:

1. The shortcut keys are inserted into the text dynamically so that it matches the config file.
2. When #1531 is implemented, it only shows the text relevant to the argument provided to the command. For example if the `flameshot gui --raw` is used, the message should be "Press Enter to capture the selection" (of course "Enter" should be dynamically replaced by whatever shortcut that is configured)

image

This can potentially become a very minimalistic version of [AwesomeWM helper](https://res.cloudinary.com/practicaldev/image/fetch/s--xozRmdEM--/c_limit%2Cf_auto%2Cfl_progressive%2Cq_auto%2Cw_880/https://dev-to-uploads.s3.amazonaws.com/i/kwmwkdi8rcnw32qqafhb.png) or [Unity keyboard shortcuts](https://upload.wikimedia.org/wikipedia/commons/9/93/Ubuntu_Unity_Keyboard_shortcuts_-_En.png)

### LINKED ISSUE #1897: Keyboard shortcuts not working
**Flameshot version**

Flameshot v0.10.1
Compiled with Qt 5.15.2

**Describe the bug**

The keyboard shortcuts (e.g. <kbd>Enter</kbd> for capturing the image or <kbd>Esc</kbd> for exiting the capture are not working for me when I start `flameshot gui`.

**To Reproduce**
Not sure why this is the case, so not sure what the exact steps are to reproduce. However, when I press those keys, they do not perform their stated function. Instead, nothing happens.

**Expected behavior**
I expect the keyboard shortcuts to work.

**System Information**
Arch Linux, running xmonad as a window manager.

veracioux:
- `CaptureRequest` class and `CaptureRequest::ExportTask` have been repurposed so they can be used for all actions that can potentially close the GUI in the future.
- Added accept button that is used to accept all capture requests that were specified from the CLI. After that the GUI closes. This will provide an elegant way to specify any action as CLI arguments (e.g. --pin, --upload, etc)
- Save, copy and imgur-upload buttons are disabled while the accept button is active
- Redesigned help text (OverlayMessage)
  - More readable format
  - Tracks configured shortcuts
- Removed `graphicCapture`, `fullScreen` and `captureScreen` DBus methods and replaced them all with a single `requestCapture` method. This method takes a byte array representation of a `CaptureRequest`. This way we don't have to recreate `CaptureRequest`s in `FlameshotDBusAdapter` which was a duplication of effort.

Closes #1336.
Closes #1416.
Closes #1531.
Closes #1813.
Closes #1897.

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -219,7 +219,8 @@ CaptureWidget::CaptureWidget(uint id,
                          QGuiAppCurrentScreen().currentScreen()->geometry());
 
     if (m_config.showHelp()) {
-        pushHelpMessage();
+        initHelpMessage();
+        OverlayMessage::push(m_helpMessage);
     }
 
     updateCursor();
@@ -317,9 +318,9 @@ void CaptureWidget::initButtons()
     m_buttonHandler->setButtons(vectorButtons);
 }
 
-void CaptureWidget::pushHelpMessage()
+void CaptureWidget::initHelpMessage()
 {
-    static QList<QPair<QString, QString>> keyMap;
+    QList<QPair<QString, QString>> keyMap;
     if (keyMap.isEmpty()) {
         keyMap << QPair(tr("Mouse"), tr("Select screenshot area"));
         using CT = CaptureTool;
@@ -342,7 +343,7 @@ void CaptureWidget::pushHelpMessage()
                         tr("Open side panel"));
         keyMap << QPair(tr("Esc"), tr("Exit"));
     }
-    OverlayMessage::pushKeyMap(keyMap);
+    m_helpMessage = OverlayMessage::compileFromKeyMap(keyMap);
 }
 
 QPixmap CaptureWidget::pixmap()
@@ -1022,8 +1023,8 @@ void CaptureWidget::initSelection()
         }
     });
     connect(m_selection, &SelectionWidget::visibilityChanged, this, [this]() {
-        if (!m_selection->isVisible() && ConfigHandler().showHelp()) {
-            pushHelpMessage();
+        if (!m_selection->isVisible()) {
+            OverlayMessage::push(m_helpMessage);
         }
     });
 }
--- a/src/widgets/capture/capturewidget.h
+++ b/src/widgets/capture/capturewidget.h
@@ -107,7 +107,7 @@ private:
     void initSelection();
     void initShortcuts();
     void initButtons();
-    void pushHelpMessage();
+    void initHelpMessage();
     void updateSizeIndicator();
     void updateCursor();
     void updateSelectionState();
@@ -173,6 +173,7 @@ private:
     NotifierBox* m_notifierBox;
     HoverEventFilter* m_eventFilter;
     SelectionWidget* m_selection;
+    QString m_helpMessage;
 
     SelectionWidget::SideType m_mouseOverHandle;
 


veracioux:
Closes #1336.

mmahmoudian:
While testing the "accept" functionality, I faced a problem which does not happen with the b650231f commit.

```sh
./Flameshot-0.10.1.x86_64.AppImage gui --raw | display
```

> ```
> display: no decode delegate for this image format `' @ error/constitute.c/ReadImage/572.
> ```

I don't know what the issue is, but it seems something is the image format is not the same as the older commit.

veracioux:
@mmahmoudian
I didn't mind the current appearance of the message box so I didn't bother to change anything. On my system, I do get a border around the message - maybe this has to do with the system theme. And I do get some transparency. I'll try to make it consistent. Rounded corners are fine and will take but a few seconds to implement.

I'll investigate the accept issue.

mmahmoudian:
I compiled against 1d334d372bbe9423f0a2b2252a9930f4dd5afc74 and I can confirm that the issue with `--raw` disappeared. I haven't looked into the AppImage and I'm significantly under-qualified to comment on what is happening there.

But found few issues:
1. the save button and copy to clipboard button are disappeared despite of being marked to be shown:
    image
2. the <kbd>Enter</kbd> terminates the capture when `flameshot gui` is running. In my head (I might be wrong) it should work so that in case of `flameshot gui` it shows the "final action buttons" (save to file, upload, copy to clipboard, ...), but when some "final action flags" are provided (e.g `--raw`, `--clipboard`, `--path`), then all those "final action buttons" should be disabled and also disappear and instead the "accept capture button" be shown.

What do you think?

veracioux:
@mmahmoudian

- Fixed the button issue when launched from systray
- The message now disappears only after you make a selection
- Changed the overlay message style as per your suggestions

Is that it?

veracioux:
@mmahmoudian Done. Ready for review?

PR #1969, opened 2021-10-12
Refactor tool size handling

### LINKED ISSUE #1952: Delay/lag when updating the object size/thickness
**Flameshot version**
Flameshot v0.10.1 (aca0db96)
Compiled with Qt 5.15.2
linux: 5.10.68-1-MANJARO

**Describe the bug**

I'm not sure if it is a bug or intentional behavior (for example for optimizing the load on the system), but there is a lengthy delay from the time I scroll until the time the thickness is updated. This delay does not exist when I scroll while drawing the object (while the left-click is still pressed during drawing the object).

In this video at first I show that the size/thickness can change immediately while drawing the object, but it is laggy afterwards:

https://user-images.githubusercontent.com/390889/136531475-869e061f-9713-4000-ac86-1723fd192884.mp4

The delay feels like a lag rather than actually programmed delay.

**To Reproduce**
1. `flameshot gui`
2. draw something
3. de-select the tool
4. click on the drawn object to select it
5. scroll & wait for the size to get updated.

**Expected behavior**

the size/thickness changes immediately as the value of the size/thickness changes.

**System Information**

> ```
> System:
>   Host: precisiontower Kernel: 5.10.68-1-MANJARO x86_64 bits: 64 
>   Desktop: KDE Plasma 5.22.5 Distro: Manjaro Linux 
> Graphics:
>   Device-1: Intel CometLake-S GT2 [UHD Graphics 630] driver: i915 v: kernel 
>   Device-2: AMD Lexa XT [Radeon PRO WX 3200] driver: amdgpu v: kernel 
>   Display: x11 server: X.Org 1.20.13 driver: loaded: amdgpu,ati,modesetting 
>   resolution: 1: 1920x1080~60Hz 2: 1080x1920~60Hz 
>   OpenGL: renderer: AMD Radeon Pro WX 3200 Series (POLARIS12 DRM 3.40.0 
>   5.10.68-1-MANJARO LLVM 12.0.1) 
>   v: 4.6 Mesa 21.2.2 
> ```

### LINKED ISSUE #1965: Better placment of "autoincreminating counter bubble" button
I think the order if buttons should get slightly modified to have a more logical order. The "autoincreminating counter bubble" button at the moment is at the end of the "final action" buttons, but it is part of the annotation tools:

image

veracioux:
- Object thickness is now correctly updated while drawing
- Refactored thickness handling in CaptureWidget and SidePanelWidget. Now uses less confusing method and sigslot names
- Added ConfigHandler::toolSize and setToolSize as wrappers around drawThickness/setDrawThickness and drawFontSize and setDrawFontSize
- Renamed thickness to tool size where generic tools are concerned

Closes #1965.
Closes #1952.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,8 @@ Now every time you press <kbd>Prt Sc</kbd> it will start Flameshot GUI instead o
 
 - If you are using Gnome you need to install the [AppIndicator and KStatusNotifierItem Support](https://extensions.gnome.org/extension/615/appindicator-support/) extension in order to see the system tray icon.
 
+- If you are using Gnome41 or later you need to install the [Unsafe Mode Extesnsion](https://extensions.gnome.org/extension/4517/unsafe-mode/). This may be changed in the future but as of now Flameshot using a private DBus interface. See [this issue](https://github.com/flameshot-org/flameshot/issues/1910) for more details
+
 - Press <kbd>Enter</kbd> or <kbd>Ctrl</kbd> + <kbd>C</kbd> when you are in a capture mode and you don't have an active selection and the whole desktop will be copied to your clipboard. Pressing <kbd>Ctrl</kbd> + <kbd>S</kbd> will save your capture to a file. Check the [Shortcuts](#keyboard-shortcuts) for more information.
 
 - Flameshot works best with a desktop environment that includes D-Bus. See this [article](https://wiki.archlinux.org/index.php/Flameshot#Troubleshooting) for tips on using Flameshot in a minimal window manager (dwm, i3, xmonad, etc).
--- a/cmake/modules/MacOSXBundleInfo.plist.in
+++ b/cmake/modules/MacOSXBundleInfo.plist.in
@@ -23,9 +23,9 @@
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>0.8.5</string>
+	<string>0.10.1</string>
 	<key>CFBundleLongVersionString</key>
-	<string></string>
+	<string>0.10.1</string>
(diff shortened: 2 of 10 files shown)


veracioux:
@mmahmoudian Can you confirm everything works correctly?

veracioux:
@mmahmoudian I shall remove the slider in those tools it doesn't affect.

I don't think the slider does anything for the invert tool. One thing that comes to mind that we could implement is a contrast setting that can be changed using the slider. Can't think of a solid use case though. I don't know why I'm even saying this...

Regarding the color grabber, I think the best course of action is to just disable scrolling when it's active. If we added a use for it, then we would have to decide whether it should change size or zoom. This can be confusing for the user. And for the sake of consistency, we would have to add a config setting to save the size/zoom. All of which is not worth it IMO. I think the current default sizes for the magnifier are suitable for most (if not all) use cases.

mmahmoudian:
> for the sake of consistency, we would have to add a config setting to save the size/zoom

Or as per my suggestion, alt key can toggle between the two, so without alt it increases the size and with alt it zooms.

> All of which is not worth it IMO

Unlike you, I have zero idea on how much work it would need 😅. You know best and let's go with that 👍🏼

> I think the current default sizes for the magnifier are suitable for most (if not all) use cases.

I agree, but there are two reasons that we haven't heard any feedback:
1. If I don't make mistake it has not been in an stable release 
2. Most users don't know about this

I will think of a way to promote new features

Regarding this PR though, everything seems to be perfectly in order.

veracioux:
@mmahmoudian I can't solve the useless-thickness-slider issue until I do some other, major refactoring. Well, I could, but in a very inelegant and frustrating way. So I'm going to merge this, and put that on my TODO list.

PR #1970, opened 2021-10-12
Add --pin, --upload, --accept-on-select to CLI

### LINKED ISSUE #526: (Enhancement) Quick Mode - Capture as soon as mouse released
Mostly in the title, haha.

---

But for those times where you need to cancel/not save immediately, maybe you can press right click after you made your selection with left click and it can either keep the current behavior (keep the selection on screen) or just delete it entirely.

Additionally, it would require a default option/list of options. Back when I used Windows, I used ShareX to copy the image to the clipboard, but also save the screenshot to a folder in case I needed to go back to it.

### LINKED ISSUE #1920: Feature Request: Screenshot a region without GUI
It would be fantastic to have the ability to screenshot a region, but skip the GUI and having to hit [ENTER] to confirm the screenshot. ShareX has this ability and it's fantastic for quick region based screenshots.

### LINKED ISSUE #1956: Support for export actions from CLI
I would like to complement the existing CLI options `-p, -c, -r, -g` to `flameshot gui/screen/full` with the following options:

- `--pin`, ~`-P`~; When the screenshot is accepted, pin screenshot. [#1834](#1834)
- `--upload`, `-u`; When accepted, upload to imgur
- `--accept-on-select` ~`--done-on-select`~, `-s`; Accept the capture as soon as a selection is made (on mouse release). [#642](#642)

To understand what I mean by **accept**, please look at #1955.

Note: Maybe the pin option should not be available in `flameshot full`. It could have weird behavior. Thoughts?

Please let me know what you think about these options. Do you have additional suggestions?

Another thought I'd like to drop here:
We could merge `--path` and `--raw` into a single `--output/-o` option. When used as `--output -` or `-o -` it should behave exactly as `--raw`. Otherwise it should behave the same as `--path`. This is more in line with the usual behavior of CLI programs. And it would allow `--pin` to be shortened to `-p`.

### LINKED ISSUE #1964: Allow `flameshot screen/full` with no options
If no options are specified to `flameshot screen/full` subcommands, I think we should open a file dialog, instead of printing an error message as we do currently. The dialog should behave just like the one triggered by the Save button in gui mode.

veracioux:
TODO:

- [X] Open file dialog when --accept-on-select is the only option to `flameshot gui`
- [x] --print-geometry not working

Closes #526.
Closes #1956.
Closes #1964.
Closes #1920.


veracioux:
@mmahmoudian

1. Fixed the --delay problem.

2. When I try
```
flameshot gui --accept-on-select --upload --clipboard
```
the "Latest uploads" entry is created just fine. I tried to upload in all the variants I could think of (various option combinations, from gui without CLI options) and it still works. Please try it again - maybe it was a temporary connection problem or something.

You are right about the copy/upload clipboard conflict. But it only arises if the config option `copyAndCloseAfterUpload` is set. This option, if true, copies the URL to the clipboard and closes the upload progress widget. If false, the widget turns into a dialog with some actions you can do with the uploaded screenshot.

I made it so that:
- If `copyAndCloseAfterUpload` is false, there is no conflict and everything behaves as it was before
- If true, then... Well, see for yourself. It's easier than to explain it.

I believe this way there is no need for a warning. If you agree with this decision, I'll merge this.

NOTE: The failing builds are due to a server error. Will try to rerun later.

mmahmoudian:
> The `copyAndCloseAfterUpload` is true by default and therefore not included in the config file by default.

This is even getting confusing for me 😆 I guess we should either up our game in documenting the config options, or include them all in the config file along with their default value.

> When both `--clipboard` and `--upload` are specified, the image is always saved to the clipboard. A dialog opens that allows you to copy the URL if you want to.

Makes total sense. Let's merge this.

> I think the upload issue is either specific to you or a bug that goes beyond the scope of this PR

I don't think it is a bug. I thing as our users are growing, there are less and less available quota and the higher chance that we get error from Imgur. For example I literally just tried and it worked just fine: https://i.imgur.com/I835ijQ.png
Basically this is an indication that we should start prioritizing the custom uploads and custom imgur API keys.

PR #1979, opened 2021-10-15
Fix autostart write path

### LINKED ISSUE #1975: Autostart doesn't work on NixOS
**Flameshot version**
Tested on both master and v0.11.1

**Describe the bug**

Autostart doesn't work.

On NixOS, if the directory `~/.config/autostart` doesn't exist, Flameshot will try to write the desktop file to somewhere in `/nix/store`. However, `/nix/store` is read-only, so that fails.

**To Reproduce**

Install Flameshot on NixOS. Launch it with `strace flameshot 2>&1 | grep open` so that we can observe where it tries to write files. When flipping `Launch at startup` in configuations, strace will print some like this:

```
openat(AT_FDCWD, "/nix/store/nkgiwqxc5i6fhkzm4lj8p11gjpclm5cj-plasma-workspace-5.21.5/etc/xdg/autostart/Flameshot.desktop", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = -1 EROFS (Read-only file system)
```

**Expected behavior**

The desktop file should be written to `~/.config/autostart`, creating the directory if necessary.

**System Information**

NixOS 21.05

yizhepku:
Closes #1975.

### DIFF
--- a/src/utils/confighandler.cpp
+++ b/src/utils/confighandler.cpp
@@ -250,7 +250,9 @@ void ConfigHandler::setStartupLaunch(const bool start)
                    << process.readAll();
     }
 #elif defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
-    QString path = QStandardPaths::writableLocation(QStandardPaths::GenericConfigLocation) + "/autostart/";
+    QString path =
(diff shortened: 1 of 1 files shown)


veracioux:
Thanks!

PR #1987, opened 2021-10-17
fixing MacOs version and documenting process

borgmanJeremy:


veracioux:
Can't it be done automatically using cmake `configure_file` somehow?

borgmanJeremy:
Yup it can i figured this was a good first step, I didnt feel like subjecting my self to CMAKE torture this morning ;)

PR #1995, opened 2021-10-21
Add default file extension configuration

### LINKED ISSUE #1304: Feature request 1of2: preferred default jpg/png/gif extension
In the settings could a setting for "preferred image extension" (ie: jpg/png/gif) be added?  
  
It seems to default to PNG atm - which means changing it every use.

Lyqst:
Closes #1304
There was already a config option for this, called `setSaveAsFileExtension`, but it wasn't exposed in the configuration panel. I've changed it to `defaultFileExtension` so it has a clearer name, and added it to the general options tab in the configuration panel as a ComboBox.

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -415,7 +415,7 @@ void GeneralConf::initSaveAfterCopy()
             SLOT(setSaveAsFileExtension(QString)));
 
     extensionLayout->addWidget(m_setSaveAsFileExtension);
-    m_layout->addLayout(extensionLayout);
+    vboxLayout->addLayout(extensionLayout);
 }
 
 void GeneralConf::historyConfirmationToDelete(bool checked)


veracioux:
@Lyqst Thank you for your contribution. I didn't test the feature in detail, but I have a few small criticisms:
- When I restart flameshot, the spin box should load the extension from the config file. Currently it displays an empty extension every time.
- I don't think you should change existing option names. This will break people's existing configs. At least until we implement a way for people to resolve configs between flameshot versions (I'm not aware of any ongoing discussions about this)
- I think the "Preferred save file extension:" text would look better if you put it inside the "Save Path" group (to make it indented).

Lyqst:
@veracioux Thanks for your comments!

> When I restart flameshot, the spin box should load the extension from the config file. Currently it displays an empty extension every time.

Not sure what you mean with spin box, but when I test these changes on my machine (Manjaro), it loads whatever option has been previously saved. Maybe I'm missing something here, please let me know.

> I don't think you should change existing option names. This will break people's existing configs. At least until we implement a way for people to resolve configs between flameshot versions (I'm not aware of any ongoing discussions about this). I agree that `setSaveAsFileExtension` is a poorly chosen name.

I see your point, makes total sense to keep compatibility with existing configs. I'll change this back to the old name.
On a side note, the previous config was only working with extensions beginning with `.` (`.png` instead of `png`), so to keep full compatibility I'll append the `.` to each options in the ComboBox.

> I think the "Preferred save file extension:" text would look better if you put it inside the "Save Path" group (to make it indented).

I agree, I'll make this change.

PR #2013, opened 2021-10-27
Add --region option

### LINKED ISSUE #479: Support position and size as input parameter
As others have suggested (#5, #334 and #395) there is no way to capture the current focused window or any window really. Instead you have to use the GUI tool and draw a rectangle yourself.

If I understand the creator correctly they do not want to add this feature if it means that migration to another system suffer because of it.

To solve this, why no introduce new parameters where the user themselves can say what positions and size a window has and then capture that region? With these new parameters a user could easily use another tool to determine the position and size and subsequently Flameshot will support taking screenshots of a window.

I am thinking something like this:
`flameshot capture xPosition yPosition xSize ySize`

Example script (this would get all parameters without any decoration):
```shell
#!/bin/bash

# vars
windowData=$(xdotool getwindowfocus getwindowgeometry)
windowPosData=$(echo $windowData | grep -Eo "[0-9]+,[0-9]+")
windowSizeData=$(echo $windowData | grep -Eo "[0-9]+x[0-9]+")
xPosition=$(echo $windowPosData | cut -d ',' -f 1)
yPosition=$(echo $windowPosData | cut -d ',' -f 2)
xSize=$(echo $windowSizeData | cut -d 'x' -f 1)
ySize=$(echo $windowSizeData | cut -d 'x' -f 2)

# exec
flameshot capture $xPosition $yPosition $xSize $ySize
```

To get decoration info you can tinker with:
`xprop _NET_FRAME_EXTENTS -id $(xdotool getActiveWindow)`

Finally you could bind a keyboard shortcut to the above script and presto, we can now capture the current active window.

### LINKED ISSUE #523: Feature: Select all on GUI start

### LINKED ISSUE #2010: Add --region option to select an exact screenshot region
**NOTE: This issue is here just so we can track it in milestones.**

### Discussed in https://github.com/flameshot-org/flameshot/discussions/1974

<div type='discussions-op-text'>

<sup>Originally posted by **veracioux** October 14, 2021</sup>
I propose a common solution to the following issues:
- #479
- #523
- #273

(The discussions in the issues became too complicated to follow)

**The problem in a nutshell**

You can't select the whole screen or a specific region without complicated and hacky scripts.

**Proposed solution**

Add a `--region` option to the `gui` and `screen` subcommands. This option can be used in the following ways:
- `--region FORMAT` (see the table below)
- `--region all` (select the entire screen, not available with the `screen` subcommand)
- `--region screen0`, `--region screen1`, etc. (select an entire screen by number, not available with the `screen` subcommand)

| Format | Pros | Cons |
| --- | --- | --- |
| X,Y,W,H | intuitive | non-standard |
| WxH+x+y | used by other CLI tools (xrandr, imagemagick) | may confuse new shell-users (minor IMO) |
| WxH,x,y | N/A | N/A |

\* *x and y are the coordinates of the top-left corner of the region*
\* *w and h are the width and height*

We would also modify the `--print-geometry` to use the format we agree upon (currently uses `W H X Y`). We can even make `--region` accept all of the formats.

:bangbang: Please vote for **one or more** of these formats in the comments below. Please vote even if you are indifferent. :bangbang:

**Other isues that may benefit from this:**
- #334
- #417
- #1509
- #1902

**What I'd like everyone to do**
- If you have other suggestions, post them as comments. We will discuss them, and if they can't be included into the existing proposal, I may ask you to open a new issue.
- Please vote for this post if you agree with the proposal.
- Vote for one of the suggested region formats
- If you prefer someone else's suggestion, vote for it</div>

veracioux:
Closes #479.
Closes #523.
Closes #2010.

Changes & TODO:

- [x] Add --region option to gui subcommand
- [x] Add --region to full
- [x] Add --region to screen
- [x] Allow specifying negative width and height
- [x] Test invalid regions (out of bounds, etc)

### DIFF
--- a/src/core/capturerequest.cpp
+++ b/src/core/capturerequest.cpp
@@ -54,8 +54,8 @@ QByteArray CaptureRequest::serialize() const
     QDataStream stream(&data, QIODevice::WriteOnly);
     // Convert enums to integers
     qint32 tasks = m_tasks, mode = m_mode;
-    stream << mode << m_delay << tasks << m_data << m_forcedID << m_id
-           << m_path;
+    stream << mode << m_delay << tasks << m_data << m_forcedID << m_id << m_path
+           << m_initialSelection;
     return data;
 }
 
@@ -71,6 +71,7 @@ CaptureRequest CaptureRequest::deserialize(const QByteArray& data)
     stream >> request.m_forcedID;
     stream >> request.m_id;
     stream >> request.m_path;
+    stream >> request.m_initialSelection;
 
     // Convert integers to enums
     request.m_tasks = static_cast<ExportTask>(tasks);
@@ -103,12 +104,22 @@ CaptureRequest::ExportTask CaptureRequest::tasks() const
     return m_tasks;
 }
 
+QRect CaptureRequest::initialSelection() const
+{
+    return m_initialSelection;
+}
+
 void CaptureRequest::addTask(CaptureRequest::ExportTask task)
 {
     if (task == SAVE) {
         throw std::logic_error("SAVE task must be added using addSaveTask");
     }
     m_tasks |= task;
+}
+
+void CaptureRequest::removeTask(CaptureRequest::ExportTask task)
+{
+    ((int&)m_tasks) &= ~task;
 }
 
 void CaptureRequest::addSaveTask(const QString& path)
@@ -121,6 +132,11 @@ void CaptureRequest::addPinTask(const QRect& pinWindowGeometry)
 {
     m_tasks |= PIN;
     m_pinWindowGeometry = pinWindowGeometry;
+}
+
+void CaptureRequest::setInitialSelection(const QRect& selection)
+{
+    m_initialSelection = selection;
 }
 
 void CaptureRequest::exportCapture(const QPixmap& capture)
--- a/src/core/capturerequest.h
+++ b/src/core/capturerequest.h
@@ -44,10 +44,13 @@ public:
     QVariant data() const;
     CaptureMode captureMode() const;
     ExportTask tasks() const;
+    QRect initialSelection() const;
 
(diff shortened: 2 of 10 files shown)


borgmanJeremy:
Does this just need reviewed and tested? Or does more work need completed?

veracioux:
@borgmanJeremy There's not much that can go wrong, but testing is always welcome.

@mmahmoudian I'd like you to test the negative W/H stuff we discussed in #2010.

borgmanJeremy:
Seems to be working well for me after some testing.

borgmanJeremy (approved):

borgmanJeremy:
@veracioux can you fix the conflicts and then lets get this merged before more conflicts arise.

veracioux:
@borgmanJeremy Yeah. Merging now.

nachovizzo:
Amazing job! Thanks a lot

veracioux:
@nachovizzo My pleasure!

PR #2023, opened 2021-10-29
Changed initilization to remove deprecated warnings

borgmanJeremy:
Cleaning up some compiler warnings. All the deprecation warnings need to be solved before we can build against Qt6.


veracioux:
@borgmanJeremy Why was `// clang-format off` ignored?

borgmanJeremy:
good question, let me figure that out and fix the formatting.

PR #2061, opened 2021-11-11
Bug report form

mmahmoudian:
This is the first try and I don't know how it will look like simply because I could not test it on my fork.


veracioux:
@mmahmoudian Try merging this branch into your fork's master branch. Then you'll be able to preview it. Works on my fork anyway...

mmahmoudian:
I thought the cleanest way would be to create a temporary repo and test them there:

https://github.com/mmahmoudian/cuddly-train/issues/new/choose

Everything seems to be working and in order. Checkout the link above to try it in action and for easy assessment.

borgmanJeremy:
Hmm doesn't seem to work on the mobile interface

mmahmoudian:
Interesting! I never tested it as I thought GitHub have already had implemented it in the phone app somehow. We can postpone the merge until it is integrated there.


In the meantime we can fine-tune the structure and text to have the least papercuts for users.

borgmanJeremy:
It's working great on desktop so I think we can merge it whenever you are ready. I only checked on mobile because I was away from my PC when the PR was updated.

I would guess nearly all bug reports come from Desktop.

mmahmoudian:
I'll do one quick test and then I'll ask you to merge

mmahmoudian:
I tried to see if we can somehow have the Markdown and YAML formats side-by-side so that it shows the Markdown when the YAML is not available, but it seems it is not possible. Users that use the Github app can still submit bug report or feature request but they will not have a template.


So let's merge this.

borgmanJeremy (approved):

PR #2126, opened 2021-11-27
[Cloud Support] Droplr

uttam-hulkapps:
* Added New Tab to manage cloud on Configuration window.
* Added support to accept droplr's username & password.

**Why to do this?**
Droplr is really amazing, but they do not have desktop app for Ubuntu users and moreover their Windows app is also not responsive as Mac version. Flameshot can provide extensive support for Droplr using REST APIs.


src/CMakeLists.txt
```diff
@@ -218,7 +218,7 @@ else()
 endif ()
 target_compile_definitions(flameshot PRIVATE APP_PREFIX="${CMAKE_INSTALL_PREFIX}")
 target_compile_definitions(flameshot PRIVATE APP_VERSION="v${PROJECT_VERSION}")
-target_compile_definitions(flameshot PRIVATE IMGUR_CLIENT_ID="313baf0c7b4d3ff")
+target_compile_definitions(flameshot PRIVATE IMGUR_CLIENT_ID="43467546c66193c")
```
mmahmoudian, line 221: Is this necessary? Why?

src/tools/droplr/droplruploader.cpp
```diff
@@ -0,0 +1,206 @@
...
+    m_NetworkAM = new QNetworkAccessManager(this);
+    connect(m_NetworkAM,
+            &QNetworkAccessManager::finished,
+            this,
+            &DroplrUploader::handleReply);
+
+    setAttribute(Qt::WA_DeleteOnClose);
+
+    upload();
+    // QTimer::singleShot(2000, this, &DroplrUploader::onUploadOk); // testing
```
mmahmoudian, line 69: Should this be removed if it was only for testing?

mmahmoudian (commented):
I put two comments one of which was regarding the Imgur ID which I don't understand why it had to be changed.

mmahmoudian:
For the sake of transparency, I put this explanation here:

Processing this PR will wait until @thepurple, @borgmanJeremy, and @veracioux give the green light considering [the discussion on Slack](https://flameshotworkspace.slack.com/archives/C019W6EPYMN/p1638181028001200).

src/CMakeLists.txt
uttam-hulkapps: Hey @mmahmoudian 
It's not necessary at all, I apologise for this change, I was testing multiple image upload and I was facing too many requests error from Imgur so I changed client ID for my test cases, I will just remove this right away.

src/tools/droplr/droplruploader.cpp
uttam-hulkapps: @mmahmoudian 
Actually I copied imgur.cpp file and this was the part of that file, I will certainly remove it from code. :)

uttam-hulkapps:
Thank you! @mmahmoudian

borgmanJeremy:
Hello, I really appreciate the work you put into this. However, we have been working for some time to clarify how we want to handle different backends for uploading images. Some time ago we settled on wanting to simply support ShareX files. I put a draft of our requirements here: 

https://github.com/flameshot-org/flameshot/discussions/1929#discussioncomment-1399520

I would like to work towards that design rather than having to keep adding more and more code to flameshot for every backend service.

Thanks again for the contribution and effor.

PR #2142, opened 2021-12-03
NC: Added UploaderManager for further multiple Storage for images support

thepurple:
No changes in logic, just code refactoring for further multiples Storage support


src/tools/imgupload/imguploadermanager.cpp
```diff
@@ -0,0 +1,70 @@
...
+    //    m_imgUploaderPlugin = "imgur";
+    //}
+    m_urlString = "https://imgur.com/";
+    m_imgUploaderPlugin = "imgur";
+}
+
+ImgUploaderBase* ImgUploaderManager::uploader(const QPixmap& capture,
+                                              QWidget* parent)
+{
+    // TODO - implement ImgUploader for other Storages and selection among them,
```
borgmanJeremy, line 39: This TODO does not reflect how we will implement more uploaders. Rather than being hardcoded in the c++ source we will parse config files.

borgmanJeremy (approved):

borgmanJeremy:
Should I merge this or is there more changes coming?

src/tools/imgupload/imguploadermanager.cpp
thepurple: Probably "TODO" is not a perfect marker here, sorry for confusing.

It works for both options:
- implement plugin system here (we'll parse config files later)
- if we won't find a time for that it is also possible to add just C++ classes for other providers here (I did such approach for Namecheap fork).

The main purpose of this PR is to remove "Imgur" code from Capture, History and other classes and move to a single place.

Please merge it if you have no strict objections.

thepurple:
Merge it please

Issue #1214, opened 2021-01-03
Allow restarting numeration from a specific number


elad2109:
Sometimes I want to numerate a few series on the same screenshot.
I would love to be able to start a new numeration with "1" again.

Also sometimes I have a saved screenshot with numeration on it.
I would love to add a new number starting from the last number already appears.

mmahmoudian:
For continuing a number series, this is what I do at the moment: click outside of the capture area enough times to get to the number I want and then I continue as usual.

elad2109:
Thanks, would still love to have a built-in feature.

borgmanJeremy:
Now that we have customizable hot keys this is actually much easier to add. We can bind it to dedicated keys.

elad2109:
> Now that we have customizable hot keys this is actually much easier to add. We can bind it to dedicated keys.


You mean "We can bind it" - as the developers or the users?

borgmanJeremy:
Sorry for the confusion, I mean as in the developers.

Creteil:
Actually, using the « Auto incrementing counter bubble », it is impossible to start at a manually chose number when using it.

Why not assigning the « CTRL + Scroll wheel » to set the starting number of the bubble ?

This actually already show numbers increasing/decreasing, just add the logic :

When used with « CTRL + Scroll wheel », **just set the bubble number to the one show in upper left and don't change size of the bubble**...

Look like there not much to change since the big part of the code already exist :

https://user-images.githubusercontent.com/27310214/111832251-757f4f80-88e8-11eb-9727-d8986abb9a8b.mp4

borgmanJeremy:
Yes I would like to add this now. I was mostly waiting for custom shortcuts to be merged. Now we can add this feature and let users customize how to trigger it.

Drarig29:
Are you still working on this @borgmanJeremy?

Drarig29:
I see you are working on other things so I made a PR.

Drarig29:
> Yes I would like to add this now. I was mostly waiting for custom shortcuts to be merged. Now we can add this feature and let users customize how to trigger it.

This custom shortcut is still a good idea though!

And I think it is easier to implement with my PR (#2162) because it would only consist in resetting `m_context.circleCount` (and show the notifier box to have a visual feedback).

aquatic7:
What is the status on this request after almost 3 years?
It would be so useful to be able to edit the number like you're able to in fx Snagit.

milnadar:
[...]

PR #2162, opened 2021-12-13
Set circle count on "Ctrl+Scroll Wheel"

Drarig29:
Related to: #1214

This PR implements this idea: https://github.com/flameshot-org/flameshot/issues/1214#issuecomment-803057997

I had to have some kind of global variable to keep track of the current circle number, so I added a `m_context.circleCount` context variable.

Maybe there is a better way, I know that something which is tool-related maybe shouldn't be in the general context.

I moved the "incrementation of the series" logic elsewhere: now the increment is done when adding a new circle.

You are not forced to [click outside of the capture area](https://github.com/flameshot\-org/flameshot/issues/1214\#issuecomment\-753593207) to continue the number series anymore.

And you can go back to:

- Previous numbers
- Reset to 1
- Reset to 0
- Negative numbers
  - I kept this because maybe there would be a use-case for this
  - If we were to actually bound the value to a minimum, this minimum should be 0. I had this need this morning because I wanted to display IDs, which were starting from 0.

### DIFF
--- a/src/tools/capturecontext.h
+++ b/src/tools/capturecontext.h
@@ -27,6 +27,8 @@ struct CaptureContext
     QPoint mousePos;
     // Size of the active tool
     int toolSize;
+    // Current circle count
+    int circleCount;
     // Mode of the capture widget
     bool fullscreen;
     CaptureRequest request = CaptureRequest::GRAPHICAL_MODE;
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -800,6 +800,20 @@ void CaptureWidget::setToolSize(int size)
     }
 }
 
+/**
+ * Show notifier box displaying the selected circle count
+ */
+void CaptureWidget::setCircleCount(int count)
+{
+    m_context.circleCount = count;
+
+    QPoint topLeft =
+      QGuiAppCurrentScreen().currentScreen()->geometry().topLeft();
+    int offset = m_notifierBox->width() / 4;
+    m_notifierBox->move(mapFromGlobal(topLeft) + QPoint(offset, offset));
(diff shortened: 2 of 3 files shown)


Drarig29:
Here is a demo:

demo

affirVega:
I like this idea.
Your PR kinda relates to №[2108](https://github.com/flameshot-org/flameshot/pull/2108). We agreed that circles must not change their number when their position in the object list is changed. I have implemented circle counter similarly to yours, but your feature is better by adding ability for user to change circle count.

---

Also, please tell what happens if you
- open a screenshot
- place a circle counter
- press Ctrl-Z to undo circle counter
- place a circle counter

Does this circle counter start with a 2 after the undo?

thepurple:
I cannot agree with such approach, it breaks the idea of Numbered Circles at all. It should be sequential numbers to be able to numerate steps one by one.
If someone really needs it, it should to be available as an option, not a replacement.
On removing some Numbered Circle in the middle, the rest should be renumbered.
Don't break existing features please.

Drarig29:
> I cannot agree with such approach, it breaks the idea of Numbered Circles at all. It should be sequential numbers to be able to numerate steps one by one. If someone really needs it, it should to be available as an option, not a replacement. On removing some Numbered Circle in the middle, the rest should be renumbered. Don't break existing features please.

Good point! As I can see I'm not a power user of flameshot, I didn't know you could select a circle and remove it!

You are right then. To be honest, my only need of this PR was to be able to reset the counter to a specific number (#1214), and specifically to 0 because one time I needed to annotate with IDs which started from 0, and I wanted to use the "Numbered circles" because it was easier (I thought) than writing text directly.

I'm closing the PR :wink:

Issue #2167, opened 2021-12-17
Alternate Hot Key Screen Capture needed for windows


kmeekva:
### Feature Description

Windows ships with snip and sketch bound to the PrtScr key.   

Flameshot is programmed to bind to that key to launch graphical screen capture interface.

But this hotkey will not work if windows is already using it.

Therefore the only way  for me to take a screen shot is to click the icon in the windows tray to activate a screen grab.

While best solution would be to allow the user to 'customize' the hot key.  I'd propose just binding both PrtScr and Alt-PrtScr -- as the Alt-PrtScr  does not conflict with any default windows keybindings.

I will create a pull request to add this simple alternative which will make flameshot imensely easier to use for windows users.

ngocphamm:
I tried Flameshot and like it, but this is what keeping me from using it, due to my nature of using a remote Windows machine using RDP from my mac. My mac keyboard just doesn't have the Print Screen key...

umernalla:
+1 to this request. 
I really would prefer to be able to choose which key I use to capture the screen - similar to GreenShot.
I actually prefer the interface of FlameShot - i.e. in situ editing - which is so much better than having to use a separate editor.
Thanks.

mmahmoudian:
I believe this is duplicate of #1341, so I close this in favor of that. If you believe these two are different, please let me know and we will discuss

PR #2168, opened 2021-12-17
Added Alt-PrtScn as alternate method for initiating screen capture

kmeekva:
Simple fix to allow Alt-PrtScr as alternative way of initiating screen capture on windows 10.
This should provide workaround for using keyboard to capture screen in windows when other WINDOWs defaults Bind to the PrtScr Key.

### DIFF
--- a/src/core/globalshortcutfilter.cpp
+++ b/src/core/globalshortcutfilter.cpp
@@ -38,7 +38,7 @@ bool GlobalShortcutFilter::nativeEventFilter(const QByteArray& eventType,
         }
 
         // Capture screen
-        if (VK_SNAPSHOT == keycode && 0 == modifiers) {
+        if (VK_SNAPSHOT == keycode && (0 == modifiers || MOD_ALT == modifiers) {
             Controller::getInstance()->requestCapture(
               CaptureRequest(CaptureRequest::GRAPHICAL_MODE));
         }


veracioux:
Closes #2167.

borgmanJeremy:
Rather than adding specifically Alt-PRNTSCREEn I think it would be better to allow users to set the hotkey.

This can be done the same way it was done on MacOS in this PR: #2098

kmeekva:
I don't think I understand how the example for MAC OS works.

I thought I could just define with initial values and it would allow users to set hotkey.

But my code fails when I try to open the configuration file with message saying "The configuration contains an error. Falling back to default configuration."

borgmanJeremy:
Okay no problem, I just got my windows partition set up so I can take a look. I think this is the last pr before we beta v11

borgmanJeremy:
Okay the reason it is crashing has nothing to do with your change, you just were the first to find it :) Here is a picture of the fix, I can merge it tomorrow or you can make the fix on your branch. I dont have my ssh keys properly set up in windows. 

image
 

That said, the hotkey change doesnt seem to be working

aprilkenopsia:
is changing the keybind still being worked on?

dt192:
This would be great to have, seems to be the only obvious large issue with the windows version.

borgmanJeremy:
closed with #4556

PR #2169, opened 2021-12-19
NC: Automatically fix configuration file "flameshot.ini"

thepurple:
Automatically fix configuration file "flameshot.ini" on start (removes unused variables and groups)

### DIFF
--- a/src/utils/confighandler.cpp
+++ b/src/utils/confighandler.cpp
@@ -386,7 +386,7 @@ bool ConfigHandler::setShortcut(const QString& actionName,
 
     bool error = false;
 
-    m_settings.beginGroup("Shortcuts");
+    m_settings.beginGroup(CONFIG_GROUP_SHORTCUTS);
     if (shortcut.isEmpty()) {
         setValue(actionName, "");
     } else if (reservedShortcuts.contains(QKeySequence(shortcut))) {
@@ -416,12 +416,12 @@ done:
 
 QString ConfigHandler::shortcut(const QString& actionName)
 {
-    QString setting = "Shortcuts/" + actionName;
+    QString setting = CONFIG_GROUP_SHORTCUTS "/" + actionName;
     QString shortcut = value(setting).toString();
     if (!m_settings.contains(setting)) {
         // The action uses a shortcut that is a flameshot default
         // (not set explicitly by user)
-        m_settings.beginGroup("Shortcuts");
+        m_settings.beginGroup(CONFIG_GROUP_SHORTCUTS);
         for (auto& otherAction : m_settings.allKeys()) {
             if (m_settings.value(otherAction) == shortcut) {
                 // We found an explicit shortcut - it will take precedence
@@ -469,7 +469,6 @@ QVariant ConfigHandler::value(const QString& key) const
 
 QSet<QString>& ConfigHandler::recognizedGeneralOptions()
 {
-
 #if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
     auto keys = ::recognizedGeneralOptions.keys();
     static QSet<QString> options = QSet<QString>(keys.begin(), keys.end());
@@ -492,12 +491,13 @@ QSet<QString>& ConfigHandler::recognizedShortcutNames()
     return names;
 }
 
-/// Return keys from group `group`. Use "General" for general settings.
+// Return keys from group `group`. Use CONFIG_GROUP_GENERAL (General) for
+// general settings.
 QSet<QString> ConfigHandler::keysFromGroup(const QString& group) const
 {
     QSet<QString> keys;
     for (const QString& key : m_settings.allKeys()) {
-        if (group == "General" && !key.contains('/')) {
+        if (group == CONFIG_GROUP_GENERAL && !key.contains('/')) {
             keys.insert(key);
         } else if (key.startsWith(group + "/")) {
             keys.insert(baseName(key));
@@ -514,6 +514,20 @@ bool ConfigHandler::checkForErrors(QTextStream* log) const
            checkSemantics(log);
 }
 
+void ConfigHandler::cleanUnusedKeys(const QString& group,
+                                    const QSet<QString>& keys) const
+{
+    for (const QString& key : keys) {
+        if (group == CONFIG_GROUP_GENERAL && !key.contains('/')) {
+            m_settings.remove(key);
+        } else {
+            m_settings.beginGroup(group);
+            m_settings.remove(key);
+            m_settings.endGroup();
+        }
+    }
+}
+
 /**
  * @brief Parse the config to find settings with unrecognized names.
  * @return Whether the config passes this check.
@@ -525,14 +539,37 @@ bool ConfigHandler::checkForErrors(QTextStream* log) const
 bool ConfigHandler::checkUnrecognizedSettings(QTextStream* log) const
 {
     // sort the config keys by group
-    QSet<QString> generalKeys = keysFromGroup("General"),
-                  shortcutKeys = keysFromGroup("Shortcuts"),
+    QSet<QString> generalKeys = keysFromGroup(CONFIG_GROUP_GENERAL),
+                  shortcutKeys = keysFromGroup(CONFIG_GROUP_SHORTCUTS),
                   recognizedGeneralKeys = recognizedGeneralOptions(),
                   recognizedShortcutKeys = recognizedShortcutNames();
 
     // subtract recognized keys
     generalKeys.subtract(recognizedGeneralKeys);
     shortcutKeys.subtract(recognizedShortcutKeys);
+
+    // automatically clean up unused keys
+    if (!generalKeys.isEmpty()) {
+        cleanUnusedKeys(CONFIG_GROUP_GENERAL, generalKeys);
+        generalKeys = keysFromGroup(CONFIG_GROUP_GENERAL),
+        generalKeys.subtract(recognizedGeneralKeys);
+    }
+    if (!shortcutKeys.isEmpty()) {
+        cleanUnusedKeys(CONFIG_GROUP_SHORTCUTS, shortcutKeys);
+        shortcutKeys = keysFromGroup(CONFIG_GROUP_SHORTCUTS);
+        shortcutKeys.subtract(recognizedShortcutKeys);
+    }
+
+    // clean up unused groups
+    QStringList settingsGroups = m_settings.childGroups();
+    for (const auto& group : settingsGroups) {
+        if (group != QLatin1String(CONFIG_GROUP_SHORTCUTS) &&
+            group != QLatin1String(CONFIG_GROUP_GENERAL)) {
+            m_settings.beginGroup(group);
+            m_settings.remove("");
+            m_settings.endGroup();
+        }
+    }
 
     // what is left are the unrecognized keys - hopefully empty
     bool ok = generalKeys.isEmpty() && shortcutKeys.isEmpty();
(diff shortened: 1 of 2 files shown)


veracioux:
While I don't mind flameshot automatically fixing the config errors that much even though it messes up my git managed dotfiles, @mmahmoudian might have firmer objections to this.

Maybe add a config option "Automatically fix configuration errors"? Note that currently a config error makes flameshot fall back to the default config, but we can make the new option exempt from that rule.

mmahmoudian:
I skimmed through the code, to me it doesn't seem that it fix anything. For instance if a color code is not up to standard it doesn't fix it as far as I can understand.

Additionally, I personally prefer to have the config file in a way that help me understand my mistakes, plus it help me configure the software with ease. Therefore, I prefer to be able to keep
- the unused items as they are
- the commented items as they are
- the order of items as they are (as long as a variable doesn't go to another group)

mmahmoudian:
I'm curious to know pros and cons of such approach so that we can decide with having all the opinions and information to our disposal

thepurple:
I don't see any cons for the end user. End user shouldn't know anything about configuration file and how to fix at all. It should just work. Moreover, we don't have any critical data in the configuration, even in some unknown for me case, the lost of some configuration variable is not critical.

thepurple:
@veracioux if I'm not mistaken, it was your fix (raise an error on unused variable in the configuration file).
It was a breaking change. If we added some variable and later will remove it, the application won't work as expected. 
I have such case in my company right now. I have neary 1500 users with old config files with outdated variables, and what do you propose to do in this case? Explain to all of them how to fix it?
Most of users don't want to know anythin about what is there under the hood, that is not user friendly. 
If you want to have an option to mange and handle it manually - it should be an option, but not a must have behavior for everyone.

@veracioux you did a great job with code refactoring, thank you. But you did it not very carefully, you have also remove number of other my fixes, you just removed what you didn't understand, and I have to fix it again.
A few examples:
- this fix: https://github.com/flameshot-org/flameshot/pull/2170
- remove automatic update of the screenshot history when the window is opened and new screenshot was taken

Please be more careful and do not change the default behavior which requires actions from the end user. It's not a toy, it a cool and great project which runs on Enterprise level.

borgmanJeremy:
No I strongly disagree. This project is meant to be fun for developers. If an enterprise requires such stability they should only work off a fork.

I work on this to have fun, not a second job.

If something breaks, politely discuss it here, but don't expect enterprise support.

veracioux:
@thepurple

> @veracioux if I'm not mistaken, it was your fix (raise an error on unused variable in the configuration file).

I implemented the change, but I did so after very thorough discussion in #1740 and #1859.

> It was a breaking change. If we added some variable and later will remove it, the application won't work as expected.
I have such case in my company right now. I have nearly 1500 users with old config files with outdated variables, and what do you propose to do in this case? Explain to all of them how to fix it?

This is a result of accumulated error on the part of the flameshot developers, which is understandable and inevitable. Things sometimes break. And your argument is a good one for introducing this PR.

> If you want to have an option to mange and handle it manually - it should be an option, but not a must have behavior for everyone.

You're right. I think we should add a config option that allows users to disable config auto-correction, but it should be enabled by default. @mmahmoudian @borgmanJeremy

> @veracioux you did a great job with code refactoring, thank you. But you did it not very carefully, you have also remove number of other my fixes, you just removed what you didn't understand, and I have to fix it again.

Sorry about that. I should have been more careful.

thepurple:
> No I strongly disagree. This project is meant to be fun for developers. If an enterprise requires such stability they should only work off a fork.
Sad to hear that, I was sure we were on the same page and we wanted to have a cool product and to be able to say proudly: "We did it!". But it's just a fun without much responsibility... :(
I was hoping to migrate all NC specific code to plugins and have a single repository. To got and deliver features directly without merging, solving merge conflicts, wasting time on extra PRs etc. I'd prefer to concentrate on code instead of this...

> I work on this to have fun, not a second job.
I understand your point, I really get some money for this, it's my extra job now, and not all of my work is paid now. I'm doing this "for fun" too, because I like this project. And I can earn much more on other job. Money is on the second place here.
I recommend Flameshot to my friends and proudly say that I'm a part of this project and it works well and stable enough. Sad to understand that it would change...

mmahmoudian:
@thepurple 
> I was hoping to migrate all NC specific code to plugins and have a single repository. To got and deliver features directly without merging, solving merge conflicts, wasting time on extra PRs etc. I'd prefer to concentrate on code instead of this...

That is still a viable possibility. I don't see any contradiction. The only thing is that a software will get some bugs and features, even some breaking ones and that is a fact for any software even Emacs. Due to perhaps miscommunication we didn't clearly know why certain code was there, plus during those discussions that @veracioux mentioned, we didn't see any objections/comments from your (NC) side.

> But it's just a fun without much responsibility... :(

From what I see, responsibility is there. The difference is in the angle of view. A corporate product is expected to be more stable than a FOSS casual project (which we all know typically not the case though).

> I have nearly 1500 users with old config files with outdated variables

I'm not totally clear how and in what context NC is using Flameshot (e.g have you white-labelled it or not and etc.), but I don't remember any bug report or discussion in our issues and discussions section that we haven't guide the user on how to fix it. Any of those 1500 users on any operating system would have been happily helped regardless of being NC user or a non-NC Flameshot user :)

While typing this I just realized perhaps the easiest was for you is to tell them to close Flameshot, delete the config file and tun Flameshot again. Their config would be gone but they instantly would regain functionality.

In the hindsight, in these type of cases automagical fix seems to be appealing. We should find the correct balance between automagical fixing and customisability.

thepurple:
> > Most of users don't want to know anything about what is there under the hood, that is not user friendly.
> 
> I personally would not say "most" but I agree that there are large number of users (esp. Windows users) are purely gui oriented.

I work for an IT company and trust me, nobody wants to know anything about the config file as long as Flameshot works well. Just me and a few system administrators know about this file.
Keep the product it simple and people will love it. It was the killer feature of Flameshot when I choose it for a big integration. After review (as end users, not as developers) my co-workers choose it too. And we decided to make it more stable and add some new features, like objects editing and history.
I don't think that we would choose it if it wouldn't be so simple and user friendly.
I really want to have a good product to be proud of, not just something for fun.


@mmahmoudian 
I would be very grateful if you would invite me to such discussions, unfortunately I do not have enough time to read everything here. 
I ​would provide number of use cases and how it is used by end users. I have a big statistics

mmahmoudian:
> I don't think that we would choose it if it wouldn't be so simple and user friendly.

I agree. I chose it because it not on my way to do my daily work. So I completely understand what you say.

> I would be very grateful if you would invite me to such discussions, unfortunately I do not have enough time to read everything here.
> I ​would provide number of use cases and how it is used by end users. I have a big statistics

You got it. Give me the list and I'll do my best to tag you in those discussions.

I believe based on all comments here, for the time being the consensus is to:
1. Have this auto-fixing toggleable
2. The default state can be "enabled"
3. If something automatically fixed, user should be notified.

For the third point, I suggest either of the following:
1. Give user a relatively detailed log of what has changed and hopefully why
2. Copy the old config (e.g "yyyymmdd-hhmm_flameshot.ini.old") and then fix it and notify the user. This way the more tech-savvy users can diff the files. This way if they want to investigate, they can turn this feature off and with @veracioux implemented methods solve the issues one by one.

thepurple:
> I'm not totally clear how and in what context NC is using Flameshot (e.g have you white-labelled it or not and etc.), but I don't remember any bug report or discussion in our issues and discussions section that we haven't guide the user on how to fix it. Any of those 1500 users on any operating system would have been happily helped regardless of being NC user or a non-NC Flameshot user :)
> 
> While typing this I just realized perhaps the easiest was for you is to tell them to close Flameshot, delete the config file and tun Flameshot again. Their config would be gone but they instantly would regain functionality.
> 
> In the hindsight, in these type of cases automagical fix seems to be appealing. We should find the correct balance between automagical fixing and customisability.

Just imaging, to tell it to 1500 users!!! I would spent a whole month on telling that. 

Let's forget about my case for a while.
Just explain me, why do we need it at all? Why user should fix something if it could be fixed by software? 
I remember that we had some problems with "bad" config files, let's fix it automatically and that is it.

Just imaging the case. You just upgraded your browser, some outdated features were removed (and it's a constant process) and you get a message like "You configuration file is incorrect, please fix it".
Would you be happy? I don't think so. You would prefer to have automatic configuration migration and do not know anything about it.

thepurple:
> @thepurple
> 
> > I was hoping to migrate all NC specific code to plugins and have a single repository. To got and deliver features directly without merging, solving merge conflicts, wasting time on extra PRs etc. I'd prefer to concentrate on code instead of this...
> 
> That is still a viable possibility. I don't see any contradiction.

I still hope so. But changing the default behavior which requires any action from the end user is a kind of contradiction. It's not a user friendly approach.

mmahmoudian:
> changing the default behavior which requires any action from the end user is a kind of contradiction. It's not a user friendly approach.

I would argue that not informing user and doing things (changing files) in the background is not user-friendly.

@borgmanJeremy @thepurple @veracioux 
How about we arrange a web-call and discuss this further? I think typing and discussion like this would take a long time whereas such discussion can be discussed in half an hour on a call. I can arrange the call if you all agree.

thepurple:
> > Why user should fix something if it could be fixed by software?
> 
> Because many users (including me) want to customize the software and they might have a typo in their config. Just one typo should not result in deletion of the whole thing. Letting users know the mistake is also educational imho.

So if you want a new feature which requires some user action for customization, make it as an option, not default one. It's the basics of mature software.
Moreover if you have to update your configuration file manually it means that your configuration editor is not good enough and the issues is configuration dialog, not in "have or not to have" an option to check your configuration file.
If you need something in your way it doesn't mean that everyone need it. You'd better to ask others. I don't like all of the decisions were made, but the majority said "yes" for some option and I accepted it.


> > Just imaging the case. You just upgraded your browser, some outdated features were removed (and it's a constant process) and you get a message like "You configuration file is incorrect, please fix it".
> > Would you be happy? I don't think so. You would prefer to have automatic configuration migration and do not know anything about it.
> 
> I partly agree, and that part is that there should be a migration mechanism (which I also mentioned some comments back). The current error at the moment is not just that. Users can open a window from the configuration window GUI to know what is the problem. If user doesn't want to be bothered, then there should be a button for "brutal autofix" (the name is a joke just to convey my point). Do you think adding such button would address the issues of your users?

Configuration dialog shouldn't allow to create broken configuration file, if we have some - it should be fixed.

thepurple:
> @borgmanJeremy @thepurple @veracioux How about we arrange a web-call and discuss this further? I think typing and discussion like this would take a long time whereas such discussion can be discussed in half an hour on a call. I can arrange the call if you all agree.

No objections, I guess we can schedule it in Slack.

In my opinion, the main question is to define the goal of the project. 
As I see we have two visions:
- Have a project for fun
- Have a project to be proud of

veracioux:
> I think typing and discussion like this would take a long time whereas such discussion can be discussed in half an hour on a call. I can arrange the call if you all agree.

Fine by me. I'm available on business days 18:00 to 23:00 UTC, on weekends all day.

thepurple:
> You got it. Give me the list and I'll do my best to tag you in those discussions.

@mmahmoudian any update changing the default behavior. 
I would be very happy to take a part in such discussions, thanks a lot for your care.

borgmanJeremy:
If we set aside the meta conversation for now and return to the specific change:

@veracioux and @mmahmoudian do you want this merged as is? Or do you think another fix is more appropriate.

veracioux:
@borgmanJeremy That would be fine by me, but I'd like to have the config option implemented ASAP. Since the option should be read even when falling back to the default config due to an error, the implementation could be trickier than just inserting a new option into the QMap. Maybe the fallback to defaults should be disabled altogether, but that's a much larger bite.

mmahmoudian:
@borgmanJeremy I'm not sure if I understand the code change as detailed as I want to. That said, I strongly believe a config option for toggling this feature is essential.

Imho considering that this doesn't affect the average user and it is almost not a breaking change (comparing with v0.10.2), I think this can be merged and we then later should come up with the config option for it when we had time to do it safely.

thepurple:
I can add an option for this, but I'd like to keep it disabled by default.

mmahmoudian:
@thepurple disabled? You just implemented this, why disabled by default?

thepurple:
> @thepurple disabled? You just implemented this, why disabled by default? This solves your problem and to my understanding does not hurt average user in any shape. Imo you can have it enabled so that your 1500 users have the issue solved.

My bad, it was a misprint, fixed it.

borgmanJeremy (approved):

PR #2171, opened 2021-12-19
NC: Add option for using large color palette on right button click

thepurple:


src/CMakeLists.txt
```diff
@@ -215,6 +215,15 @@ if (WIN32)
     endif ()
 endif ()
 
+# Choose default color palette (small or large)
+if($ENV{FLAMESHOT_PREDEFINED_COLOR_PALETTE_LARGE})
```
borgmanJeremy, line 219: I would prefer for this not to be a compile time option. The reason is to properly test different combinations we must build Flameshot multiple times with different flags. I think it would be better as a run time option. 

If its specifically for the NC uses you could also distribute a default config that sets this option.

borgmanJeremy (changes requested):

src/CMakeLists.txt
thepurple: This features is configurable and you can find it in the "generalconf.cpp". This flag changes just the default value.
I did it to keep the default behavior for both NC users and Community users. You shouldn't set this flag, the default behavior is the Community behavior.
If it is not acceptable I'll remove it, but I'd like to have it.
Unfortunately I cannot provide default config right now, because users already have it with their customization. I have just two options:
- Provide compile flags
- Implement mechanism for providing new default flags which differs from the Community edition.

borgmanJeremy: Okay ill merge it as is. My only concern was related to stability since I won't really know if something breaks this feature flag as it will be untested.

PR #2174, opened 2021-12-20
Implement abstract logger

veracioux:
As briefly advertised some time ago, I implemented an AbstractLogger thatserves
as a common interface for all messages/notifications/logs. I replaced all
notifications with calls to AbstractLogger. Those calls initiate notifications
just like before, but now also print those same messages to the CLI using a
standard format:

```
flameshot: info/warning/error: ...
```

And I replaced some of the `QTextStream(stderr)` calls with AbstractLogger.

**Some problems:**

- Since copy to clipboard is handled by the daemon, the CLI notification
  is also printed by the daemon. The fix for this is unrelated to the
  AbstractLogger implementation, and I'll implement it another time
- Logging to a file is not implemented. We have to discuss the dynamics of
  that first, in an issue. I'll get around to opening that discussion, or if
  someone else feels more emotionally attached to that feature they can do it.

Let me know what you think. And please test this out - I might have missed some
spots, i.e. forgot to rewrite some old messages/logging using the new interface.

Additionally, I managed to fix the annoying QPainter warnings (which were
printed at the subcommand since the DBUS PR, and not at the daemon). The fix is
inside the QtColorWidgets library. I'm going to try to push that fix upstream
as well.

### DIFF
--- a/external/Qt-Color-Widgets/src/QtColorWidgets/color_wheel.cpp
+++ b/external/Qt-Color-Widgets/src/QtColorWidgets/color_wheel.cpp
@@ -231,7 +231,11 @@ void ColorWheel::mouseReleaseEvent(QMouseEvent *ev)
 
 void ColorWheel::resizeEvent(QResizeEvent *)
 {
-    p->render_ring();
+    static bool skipFirst = true;
+    // Skip the first time in order to prevent QPainter warning messages
+    if (!skipFirst)
+        p->render_ring();
+    skipFirst = false;
     p->render_inner_selector();
 }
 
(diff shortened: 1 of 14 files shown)


borgmanJeremy:
Do you think this should go into v11?

veracioux:
@borgmanJeremy Yeah, it doesn't make any huge user-facing changes.

borgmanJeremy:
Seems to be working well on Linux, ill test macos later today and then merge. Did you test on windows? If not I can test there too.

veracioux:
@borgmanJeremy Yes please test on windows. I don't have access to a windows machine, nad will need some time to setup a VM. Good news is I got my hands on a Macbook and I'll use it for flameshot tests. As soon as I figure out how to launch flameshot on it :)

borgmanJeremy:
I tested on MacOS, im going to merge now and test on windows later since I will be doing extensive testing for v11.

Issue #2192, opened 2021-12-26
Latest V11.0 Beta does not start on macOS 12.1 


dannykorpan:
### Flameshot Version

V11.0 Beta

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

macOS 12.1 (21C52) - Apple Silicon M1

### Description

Latest Flameshot beta crashes upon start

Crash Report
```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               flameshot [3458]
Path:                  /Applications/flameshot.app/Contents/MacOS/flameshot
Identifier:            org.flameshot
Version:                (11.0.0)
Code Type:             X86-64 (Translated)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2021-12-26 10:47:51.6635 +0100
OS Version:            macOS 12.1 (21C52)
Report Version:        12
Anonymous UUID:        EC5B89F7-21E0-C15F-CA55-C65477D73543

Sleep/Wake UUID:       2806398D-C49A-4C3B-BA74-9879C12DA04C

Time Awake Since Boot: 16000 seconds
Time Since Wake:       869 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
abort() called
__cxa_guard_acquire detected recursive initialization


Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   ???                           	    0x7ff8a69c6940 ???
1   libsystem_kernel.dylib        	    0x7ff816413112 __pthread_kill + 10
2   libsystem_c.dylib             	    0x7ff816395d10 abort + 123
3   libc++abi.dylib               	    0x7ff8164060b2 abort_message + 241
4   libc++abi.dylib               	    0x7ff816408d12 __cxa_guard_acquire + 242
5   flameshot                     	       0x10026281a Controller::getInstance() + 42
6   flameshot                     	       0x10027869a SystemNotification::sendMessage(QString const&, QString const&, QString const&, int) + 90
7   flameshot                     	       0x10026901b AbstractLogger::sendMessage(QString, AbstractLogger::Channel) + 91
[...]

PR #2196, opened 2021-12-27
Fix recursive static init

borgmanJeremy:
Removed default config for abstract logger that sends a system notification, also cleaned up a useless ifdef


borgmanJeremy:
Closes #2192

veracioux (commented):
@borgmanJeremy See #2197

PR #2197, opened 2021-12-27
Send notifications from event loop on Win and Mac

veracioux:
Solves the static recursion without disabling any features.

We just have to deal with the fact that the Controller is a bit messy
and has some circular dependencies with other classes. Unfortunately,
I don't see a good opportunity to refactor, but we might get some new
ideas with the passage of time.


borgmanJeremy (approved):

borgmanJeremy:
Thanks!

PR #2202, opened 2021-12-29
Customizable ColorPicker

### LINKED ISSUE #1942: Add black and white as default-color
Would be great if black and white would be always there for the default-colors... without the need to add them to the ini-file.

deo002:
Closes #1942

## Functionality added:
- Select preset from colorpicker using spinbox.
- Press Delete to remove that preset from colorpicker.
- Choose color from color-wheel or enter it manually.
- Press Add to add that preset to colorpicker.


borgmanJeremy:
Thanks, this looks really good for a first pass. Since it touches so many place in the code and we are very close to the version 11 release I will wait to merge this until after the release.

Some comments based on testing (haven't done code review yet).
- If I add too many colors I get an error that says "Unable to add preset". This should probably state that the error stems from a cap on how many colors can be added.
- I think we should increase the number of colors to match the number that was added in #2171. @thepurple I think then we could remove the build flag for the pallet size. How does that sound?
- It would be nice if there were arrows or something that let you re-order the colors. This could be added in a future PR, not a blocking issue.

deo002:
Thanks @borgmanJeremy! Third one would be a really nice addition.

thepurple:
@borgmanJeremy it sounds great, thx.

borgmanJeremy:
Okay the version 11 issues are slowing down so I am ready to look at this / merge. Can you remind me where it left off? We just need to do a code review and some testing?

deo002:
@borgmanJeremy Yep, testing and code review are left. 
Some changes were suggested too. I will implement them along with the changes you would suggest in the code review. 
Thanks!

src/widgets/colorpickerwidget.cpp
```diff
@@ -0,0 +1,139 @@
...
+{
+    // draw the highlight when we have to draw the selected color
+    if (i == m_selectedIndex) {
+        QColor c = QColor(m_uiColor);
+        c.setAlpha(155);
+        painter.setBrush(c);
+        c.setAlpha(100);
+        painter.setPen(c);
+        QRect highlight = m_colorAreaList.at(i);
+        highlight.moveTo(highlight.x() - 3, highlight.y() - 3);
```
borgmanJeremy, line 47: Why is this "-3"? Maybe 3 as a const variable with a descriptive name so its easier to read in the future?

src/widgets/colorpickerwidget.cpp
```diff
@@ -0,0 +1,139 @@
...
+        highlight.setHeight(highlight.height() + 6);
+        highlight.setWidth(highlight.width() + 6);
```
borgmanJeremy, line 49: Same question on 6

src/widgets/spinbox.cpp
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// SPDX-FileCopyrightText: 2017-2019 Alejandro Sirgo Rica & Contributors
+
+#include "src/widgets/spinbox.h"
+#include "src/utils/confighandler.h"
+
+SpinBox::SpinBox(QWidget* parent)
```
borgmanJeremy, line 7: Im not sure SpinBox is the best name for this. Maybe put something in the name about what makes it different from a normal spin box? Like ColorSpinBox or something

src/config/colorpickereditor.cpp
```diff
@@ -0,0 +1,161 @@
...
+{
+    ConfigHandler config;
+    QVector<QColor> colors = config.userColors();
+
+    if (colors.contains(m_color))
+        return;
+
+    colors << m_color;
+
+    const int maxPresetsAllowed = 10;
```
borgmanJeremy, line 111: Make this maximum match #2171 so we can remove this compile time option.

src/config/colorpickereditor.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    if (colors.size() > maxPresetsAllowed) {
+        QMessageBox::critical(this, tr("Error"), tr("Unable to add preset."));
```
borgmanJeremy, line 114: Adjust this error to state why the preset cannot be added "IE exceeded maximum" or something.

src/config/colorpickereditor.cpp
```diff
@@ -0,0 +1,161 @@
...
+    ConfigHandler config;
+    QVector<QColor> colors = config.userColors();
+
+    colors.remove(m_selectedIndex);
+
+    const int minPresetsAllowed = 3;
+
+    if (colors.size() < minPresetsAllowed) {
+        QMessageBox::critical(
+          this, tr("Error"), tr("Unable to remove preset."));
```
borgmanJeremy, line 132: Adjust this to say why it cannot be removed.

borgmanJeremy (changes requested):
Thank you this is a great feature! 

I had some very minor comments. I had to fix some merge conflicts to do my testing but it was pretty easy. 

I'll get this merged quickly once the comments are addressed.

deo002:
Thanks for feedback! I have pushed a fix addressing all the changes. @borgmanJeremy

borgmanJeremy:
Thanks!

PR #2204, opened 2021-12-29
NC: fix - crash on Windows if "setSaveAsFileExtension" is defined

thepurple:
fix - crash on Windows if "setSaveAsFileExtension" is defined in the configuration file


veracioux (commented):
@borgmanJeremy This is a backwards-incompatible change, and will drop the value set by the user prior to this version. We have to add a release notice that `setSaveAsFileExtension` has changed to `saveAsFileExtension`. Even better would be to add a migration mechanism, which I'll try to implement.

Since we're changing the name, I suggest we change it to `defaultFileExtension` instead.

borgmanJeremy:
Agree default makes more sense. I also don't understand How this fixes a crash but I'm limited to mobile for a while. I thought we fixed the crash with the change to the static init

thepurple:
This bug appeared after the code refactoring with ConfigHandler, setter and getter was the same function. I din't understand why it not crashed on Linux.
Root cause: 
`CONFIG_GETTER_SETTER(setSaveAsFileExtension, setSaveAsFileExtension, QString)`
So I changed it to 
`CONFIG_GETTER_SETTER(saveAsFileExtension, setSaveAsFileExtension, QString)`
About changing the parameter name I decided that this one is not critical and we can just rename it. You can set the name to the previous one and we'll have no backwards-incompatible.

thepurple:
BTW Previously we have value for this parameter like `JPEG file (*.jpg)` and now it's just `jpg`. So it is not backwards-incompatible anyway, that's why I decided to change the name to the same as getter function (it's our current style). And I don't think that it's a good idea to have functions and parameters with different names. It will be confusin.

veracioux:
> BTW Previously we had value for this parameter like `JPEG file (*.jpg)` and now it's just `jpg`.

Yeah, you're right about that. @borgmanJeremy This should also be included in the release notes if it's not already.

borgmanJeremy:
I will merge as is, but I agree we might want to auto correct this issue.

Issue #417, opened 2018-12-06
Repeat last shot


sliptonic:
I make documentation for another project.  I need to be able to take multiple shots while the content of the screen changes.  I need the ability to repeat the last screenshot again without setting up the region and size over and over.  It would be nice to have an option in the main menu to 'repeat last shot' which would do exactly that.

HotShots has this feature but that project seems to be dying.

josipjelic:
Please implement this. Should be fairly easy but extremely useful when creating documentation.

miili:
Oh yess!

ardydo:
+1
 This was one of my favorite feature from shutter and I really miss it.

shelper:
Same here, please do it

nikolayhg:
Yeah, that would save us a lot of time! Similar to the "Capture last region" shortcut of Greenshot under Windows.

jeaye:
I came here specifically looking for a program which can do this. Flameshot has a lot of features which are enticing, but repeating the last shot (or remembering the last region) is a must have.

This is primarily because selecting a region and moving it to get it just right is finicky and should only need to be done once. I can set the region, take a shot, then change the contents of the window, then take another shot using the same exact region. There won't be any jitter when switching between the captured images, allowing the actual content difference to be more easily the focus.

elmerehbi:
+1 for this

lawipac:
+1 for this function.

pglira:
+1!

bart9h:
Just add an option to remember the last area, and start with that.

Or even make it the default, and get rid of the pink dialog for good.
Just start with rectangle with 1/2 of the screen at the center (or something like that) if there was no previous capture.

mfilser:
+1

alerque:
I tracked down this issue tracker to report this. Also coming from Shutter and this is a make-or-break feature.

Another way to implement this besides the ones mentioned would be to have a `flameshot region --geometry` command where X window coordinates (X origin, Y origin, Width, Height) could be used to specify a consistent region allowing multiple captures of the same coordinates.

njminchin:
This is a must-have feature - I can't use Flameshot without it :( back to Greenshot for now

borgmanJeremy:
This is now possible in the CLI in version 11 beta.

jeaye:
Thanks for the update, @borgmanJeremy! I don't think this issue should be closed, though.

The original request was:

[...]

PR #2208, opened 2021-12-31
first pass at region in launcher

borgmanJeremy:
## Todo

- [x] Size combo boxes more appropriately 
- [x] Hide region settings when in full screen mode
- [x] If there was no "last region", rather than setting 0x0+0+0, set the full resolution
- [x] Decide where to store the last shot

## Demo
Peek 2021-12-30 20-06

### DIFF
--- a/src/widgets/capturelauncher.cpp
+++ b/src/widgets/capturelauncher.cpp
@@ -8,8 +8,10 @@
 #include "src/utils/screengrabber.h"
 #include "src/utils/screenshotsaver.h"
 #include "src/widgets/imagelabel.h"
+#include <QDir>
+#include <QFile>
 #include <QMimeData>
-
+#include <QStandardPaths>
 // https://github.com/KDE/spectacle/blob/941c1a517be82bed25d1254ebd735c29b0d2951c/src/Gui/KSWidget.cpp
 // https://github.com/KDE/spectacle/blob/941c1a517be82bed25d1254ebd735c29b0d2951c/src/Gui/KSMainWindow.cpp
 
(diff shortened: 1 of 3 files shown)


borgmanJeremy:
Closes #417

src/utils/confighandler.cpp
```diff
@@ -120,6 +120,7 @@ static QMap<class QString, QSharedPointer<ValueHandler>>
     OPTION("predefinedColorPaletteLarge", Bool               ( PREDEFINED_COLOR_PALETTE_LARGE )),
     // NOTE: If another tool size is added besides drawThickness and
     // drawFontSize, remember to update ConfigHandler::toolSize
+    OPTION("lastRegion"                  ,String             ( ""            )),
```
veracioux, line 123: I am utterly opposed to the inclusion of volatile data into the config. I quote what @mmahmoudian said in https://github.com/flameshot-org/flameshot/issues/1740#issuecomment-896567222 
> Many do version control their dotfiles and I personally hate it when a software change the dotfile without my permission and knowledge (e.g [Konsole](https://konsole.kde.org/), [screenkey](https://gitlab.com/screenkey/screenkey/)) because it will ruin the `git status`.

I propose we add a `volatile.ini` as a sibling of `flameshot.ini` so that the user can still change this value manually if needed. Or you can put it in `~/.local/yadda-yadda`, but I find this to be sub-optimal.

src/widgets/capturelauncher.cpp
```diff
@@ -69,10 +71,26 @@ CaptureLauncher::CaptureLauncher(QDialog* parent)
...
             });
 
+    m_width = new QLineEdit("0");
+    m_height = new QLineEdit("0");
+    m_xOffset = new QLineEdit("0");
+    m_yOffset = new QLineEdit("0");
+
+    setLastRegion();
+
+    auto* dims_layout = new QHBoxLayout();
```
veracioux, line 85: It's not really my place to tell you this, but can you keep the naming style camelCase? I have a particular allergy to inconsistency.

src/widgets/capturelauncher.h
```diff
@@ -36,4 +38,10 @@ private slots:
     QPushButton* m_launchButton;
     QLabel* m_CaptureModeLabel;
     ImageLabel* m_imageLabel;
+
+    QRect m_initial_rect;
```
veracioux, line 42: Here too.

veracioux (changes requested):
- Label is not vertically aligned with the input boxes.
image
- When I manually change the parameters in the edit boxes, it doesn't affect the selection

I think that much more can be done with this feature and would prefer to leave this for the next release.

src/widgets/capturelauncher.cpp
borgmanJeremy: Yeah sorry, i have been doing like 70% rust coding and am in the headspace of their conventions.

src/utils/confighandler.cpp
borgmanJeremy: Hmmmm, if we move this to another file it probably won't be ready for V11. I'm okay with that either way. 

Maybe we can just use https://doc.qt.io/qt-5/qtemporarydir.html#details, the downside is it will not preserve the last screenshot between invocations.

borgmanJeremy:
Ok, Let's kick this to version 12 since we need some more discussion.

src/utils/confighandler.cpp
veracioux: @borgmanJeremy I feel that such a constraint would be felt by the users.

borgmanJeremy: Then lets use: QStandardPaths::CacheLocation

I think on linux that will be ~/.cache which also conforms to the XDG specification

Issue #390, opened 2018-10-30
[Feature Request] Add magnifying frame for corner selection


steelstrings:
I propose a magnifier frame for pixel-perfect screenshots, which appears when you drag-resize one of the corners w/ your cursor. Activate either in a setting, on keypress, or both.

`spectacle` for example has got support for this

![snapshot_2018-10-30_22 56 41](https://user-images.githubusercontent.com/24418003/47753321-71aaee00-dc97-11e8-98e3-9038106d3794.png)

However instead of a crosshair which determines the corner pixel of the screenshot being taken, I would suggest to keep everything inside the corner and cut off everything which is overlaid.

![snapshot_2018-10-30_22 57 28_2](https://user-images.githubusercontent.com/24418003/47754446-38747d00-dc9b-11e8-916b-55b159c20767.png)

k1nghat:
im not sure how it would be implemented but i love that you can arrow over px by px to get that precision but without some sort of magnifier its not really that useable.

sfire:
@steelstrings I totally agree with you that a magnifier while **placing the cursor** and **adjusting the border** will help a lot. 
I've been using a screen-capture program([Snipaste](https://github.com/Snipaste/feedback)) under Windows which has this feature, and shows **coordinates** and **pixel color value**.

![image](https://user-images.githubusercontent.com/5409756/53693867-58e74c00-3de1-11e9-88df-7f155c8b1b55.png)
Hope this will help @lupoDharkael add this feature.

densha:
Any update on this feature request? I came across flameshot after looking for a replacement for Shutter on Linux. This feature would make flameshot on par with other screenshot tools that have this feature (Shutter, Snagit, ShareX).

borgmanJeremy:
I'll take a look at it, it would be pretty useful.

cacarr-pdxweb:
> Any update on this feature request? I came across flameshot after looking for a replacement for Shutter on Linux. This feature would make flameshot on par with other screenshot tools that have this feature (Shutter, Snagit, ShareX)

I expect many people are looking for a replacement for Greenshot, which does not have a Linux version. Flameshot is very similar and a good replacement, except for the lack of this feature. 

If this could be added at some point, that would be great.

borgmanJeremy:
I started looking into it and it wasn't a quick add. I still would like to add it but right now its in the backlog as we focus on some other improvements.

[...]

Issue #328, opened 2018-08-29
Zoom on cursor


jeancastel:
Hi,
It'll be a good feature to implement a zoom box around the cursor during screen selection.
This feature will permit to take precises screenshots.

Thank you !

kyrios123:
I agree, this is really something I miss to easily select the windows border (especially when being on a host taking a screenshot of a window running in a virtual guest machine).
Something similar to the [magnifying glass of greenshot](http://www.cid-auditeur.com/images/bt_portfolio/87/original/36896cbbee8440076c0735ae21d22bea.png) for example would be awesome!

anmsh:
+1  I love this on greenshot and would love to see on flameshot now. Another vote for this.

mmahmoudian:
@veracioux considering that you have implemented the zoom thing for the color selection, and considering the suggestion of @borgmanJeremy to have a key pressed for loading the zoomed part, can we use these elements to implement this?

I'm bot saying that this feature request should be done, I'm just saying that I think the elements are there and can potentially be combined.

veracioux:
@mmahmoudian Sure, I can try to extend the feature some time. But I am really busy lately so I'll have to put it on hold or someone else can try to adapt my code from the color grabber.

PR #2219, opened 2022-01-01
Added magnifier for more precise selections

SilasDo:
related to [#328](https://github.com/flameshot-org/flameshot/issues/328) and [#390](https://github.com/flameshot-org/flameshot/issues/390)

I was missing a magnifier so I decided to add this feature.
not really happy about the way I integrated  the magnifierwidget into the capture process but its good enough for my use case.
(for example: right now it does not show the magnifier when using the move tool)

I'm open to feedback.
If this is something you wold like to see in the Flameshot I'm willing to improve the implementation. 

demo:

https://user-images.githubusercontent.com/26182436/147848831-88eb23bd-c30c-4d1f-8974-66ae097225cc.mp4

btw not sure if you already know this but the selection is bugged in v11 beta, for example the selection is no longer constrained to the screen (you can simply drag the selection offscreeen).

https://user-images.githubusercontent.com/26182436/147849148-e1f62db4-4a2e-44ca-a56c-33167c2edfc7.mp4

another bug i encountered  in v11 is that the first pixel you selected does not end up in the selection, an easy way to test this is  to select your entire screen top-left to bottom-right (you would expect the selection to  equal your screen resolution but it does not)

I compiled the v10.2 version and had neither of those issues.

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -57,6 +57,8 @@ GeneralConf::GeneralConf(QWidget* parent)
 
     m_layout->addStretch();
 
+    initShowMagnifier();
+    initSquareMagnifier();
     // this has to be at the end
     initConfigButtons();
     updateComponents();
@@ -79,6 +81,8 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
       config.historyConfirmationToDelete());
     m_checkForUpdates->setChecked(config.checkForUpdates());
     m_allowMultipleGuiInstances->setChecked(config.allowMultipleGuiInstances());
+    m_showMagnifier->setChecked(config.showMagnifier());
+    m_squareMagnifier->setChecked(config.squareMagnifier());
 
 #if !defined(Q_OS_WIN)
     m_autoCloseIdleDaemon->setChecked(config.autoCloseIdleDaemon());
@@ -618,6 +622,24 @@ const QString GeneralConf::chooseFolder(const QString pathDefault)
         }
     }
     return path;
+}
+
+void GeneralConf::initShowMagnifier()
+{
+    m_showMagnifier = new QCheckBox(tr("Show magnifier"), this);
+    m_scrollAreaLayout->addWidget(m_showMagnifier);
+    connect(m_showMagnifier, &QCheckBox::clicked, [](bool checked) {
+        ConfigHandler().setShowMagnifier(checked);
+    });
+}
+
+void GeneralConf::initSquareMagnifier()
+{
+    m_squareMagnifier = new QCheckBox(tr("Square shaped magnifier"), this);
+    m_scrollAreaLayout->addWidget(m_squareMagnifier);
+    connect(m_squareMagnifier, &QCheckBox::clicked, [](bool checked) {
+        ConfigHandler().setSquareMagnifier(checked);
+    });
 }
 
 void GeneralConf::togglePathFixed()
--- a/src/config/generalconf.h
+++ b/src/config/generalconf.h
@@ -67,6 +67,8 @@ private:
     void initUseJpgForClipboard();
     void initUploadWithoutConfirmation();
     void initPredefinedColorPaletteLarge();
+    void initShowMagnifier();
+    void initSquareMagnifier();
 
     void _updateComponents(bool allowEmptySavePath);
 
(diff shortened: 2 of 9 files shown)


veracioux:
@SilasDo I appreciate you taking the time to implement this. But you should have first discussed this via issues (the devs only agreed that this feature is useful but not how it should be executed). Right off the bat I'd like to say that you should take what I say with a grain of salt because @borgmanJeremy and @mmahmoudian might disagree with me.

We were looking at a solution that reuses the color grab zoom widget I implemented in #1869, in order to prevent unnecessary code duplication.
Would you be willing to combine your effort with my implementation so that we use the same code for both functionalities? You could adapt my code to reuse yours, or you can adapt your code to reuse mine. I personally prefer the second option, obviously because I am biased (that's why @borgmanJeremy should give his opinion), but I can also list the following reasons:
- It already basically implements the magnifier widget, it just needs a crosshair and a round variant
- It implements the color grabber, so it doesn't require any adaptation on this front
- It is decoupled from `CaptureWidget`. Basically the recent refactoring trend is to move functionality outside of `CaptureWidget::mouseMoveEvent` and elsewhere

**Some criticism**
- May be personal preference, but I don't see a need to include both a circular and a square magnifier. I really don't like including the "Square shaped magnifier" config option. It strikes me as arbitrary, inelegant, and bloats the config and corresponding GUI (@borgmanJeremy might disagree with me on this)
- The magnifier should IMO be toggleable using the keyboard and/or mouse
- See the following recording, which happens when the `FLAMESHOT_DEBUG_CAPTURE` option is `ON` (this is what I always use when debugging flameshot, so I'd like it to work properly):
Peek 2022-01-01 15-22

If we collectively decide that you should merge this with my implementation, then please hold off on working on it until I fix a bug on MacOS regarding the `ColorGrabWidget` (I'll inform you when it's done). But in any case, this feature should not be included in the v11 release.

P.S. I was able to reproduce the existing bugs you mentioned, and I've started working on them. Thanks for testing out the beta!

mmahmoudian:
@SilasDo thank you for taking the time to implement such feature. It's good to know there are users who can contribute to the codebase.

My suggestion is to have them both using the same code as it would increase the maintainability.

Regarding the shape, I myself love to see the crosshairs to show the center pixel, plus I feel like the circle shape is more smoother to the eye and intuitive as IRL magnifiers are typically circular.

Also regarding in which version to release this in, I would also go with the idea that let's not delay the v11 but add this to the next minor/major version.

So to sum up, I completely agree with the points @veracioux made.

borgmanJeremy:
@SilasDo  now that v11 is calming down we have bandwidth to help get this integrated into v11. I think @veracioux has good points on merging the implementation with the existing magnifier feature. Are you willing to take on that work?

If not we can take your code and port it over. Overall I really like the demo.

SilasDo:
I'm currently quite busy with another project, i could probably find some time in about a week.

i took a quick look at the colorgrabwidget (really like the hold left click for precise selection feature) 

but I'm not really sure how to best merge both implementations since they are currently not very similar.
e.g: magnifier on top of cursor vs at an offset

things like that should be agreed upon in order for me to merge features.

that being said If someone else has time to implement this, that would be totally great as well!

mmahmoudian:
> but I'm not really sure how to best merge both implementations since they are currently not very similar.
> e.g: magnifier on top of cursor vs at an offset

I think the one with offset is good for when user wants to find the region of interest, and then the rightist can freeze the zoom window do the precise pixel can be selected. For me the hardest part is to find the region, for instance in the current implementation if I want to get the color of a green tongue in 🤑, I should move the mouse quite a bit to find the region of the tongue, so the offset can help because it doesn't cover my field of view.

borgmanJeremy:
@veracioux when you say "merge the implementations" do you mean you want this PR merged as is?
 
Or are you going to merge this PR to another branch, combine the zoom classes, and then submit a new PR?

veracioux:
@borgmanJeremy I meant the latter. But you could also merge this PR into master if you think it is mature enough, so `flameshot-git` users don't miss out on it. And then I can make my changes in another branch.

borgmanJeremy:
@veracioux okay I think the best way to approach this will be to merge as is and iteratively improve it.

borgmanJeremy (approved):

PR #2220, opened 2022-01-01
Version 11 housekeeping

veracioux:
A branch meant for house keeping changes for the v11 release (e.g. editing release notes). I recommend to keep it open until v11 is released.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.13)
 # cmake_policy(SET CMP0076 OLD)
 
-set(FLAMESHOT_VERSION 11.0)
+set(FLAMESHOT_VERSION 11.0.0)
 
 # Flameshot-org
 set(GIT_API_URL "https://api.github.com/repos/flameshot-org/flameshot/releases/latest")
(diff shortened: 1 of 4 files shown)


veracioux:
@borgmanJeremy Actually, I don't think I understand the build process on Windows enough to make the change from 11.0.0 to 11.0. I forgot to ask - do we even want to make this change? :D

borgmanJeremy:
> @borgmanJeremy Actually, I don't think I understand the build process on Windows enough to make the change from 11.0.0 to 11.0. I forgot to ask - do we even want to make this change? :D

I'm okay either way. It might be less risky to wait until 12.0. My main concern is messing up the windows package and making sure that https://github.com/flameshot-org/flameshot/blob/fcadc02a293f14a3e88a85744fae45decdf79915/src/core/controller.cpp#L174 still works

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.13)
 # cmake_policy(SET CMP0076 OLD)
 
-set(FLAMESHOT_VERSION 11.0.0)
+set(FLAMESHOT_VERSION 11.0)
```
xvitaly, line 4: Not semver compliant.

borgmanJeremy: I don't intend to follow semver... each release will bump the first number... patches will bump the second

PR #2223, opened 2022-01-03
Fix flameshot not exiting when it should

veracioux:


src/core/controller.h
```diff
@@ -35,12 +35,9 @@ class Controller : public QObject
 
     Controller(const Controller&) = delete;
     ~Controller();
-    void operator=(const Controller&) = delete;
```
borgmanJeremy, line 38: Why is this line removed? It seems like the copy assignment for this class might need to be explicitly marked delete or we should implement the copy/ assignment operators. I don't think the default ones will be correct. (Nor do I think this class should be assigned  or copied).

borgmanJeremy (changes requested):

src/core/controller.h
veracioux: https://doc.qt.io/qt-5/qobject.html#no-copy-constructor-or-assignment-operator

borgmanJeremy: Oh, thanks.. Today I learned.

borgmanJeremy (approved):

PR #2226, opened 2022-01-05
Fix broken colorgrabber

veracioux:
Fixed it on macOS, now trying to fix it on Windows.
Done.

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -106,7 +106,8 @@ CaptureWidget::CaptureWidget(const CaptureRequest& req,
 
 #if defined(Q_OS_WIN)
         setWindowFlags(Qt::WindowStaysOnTopHint | Qt::FramelessWindowHint |
-                       Qt::Popup);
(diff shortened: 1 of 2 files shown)


veracioux:
@borgmanJeremy When using the `Qt::Popup` window flag on `CaptureWidget` on Windows, opening other widgets (i.e. `ColorGrabWidget`) would close the `CaptureWidget`. So I changed it to `Qt::SubWindow`. Basically the only role of `Qt::Popup` was to hide the taskbar icon, as far as I'm aware. Debugging this issue drove me nuts :/

@borgmanJeremy @mmahmoudian If you can spare some time, test the GUI (especially color grabber) on all platforms you can conveniently access. I tested on Linux (i3wm, GNOME, Plasma), Mac, Windows and it seems to work fine.

I didn't test on any platforms with Wayland though.

mmahmoudian:
It works fine for me when built against 9fe0690

borgmanJeremy (approved):

borgmanJeremy:
I will test on gnome wayland and then merge.

PR #2238, opened 2022-01-10
Includes the mouse cursor into the screenshot

AndreaMarangoni:
Few things to consider:

- this is a draft
- only tested in Linux. I don't have a Windows or Mac machine to test it.
- I have included a configuration to enable/disable it
- I didn't go further than "make it work" at this stage. I want to see if the direction I am going is the correct one.

Once the configuration is set, start `flameshot gui`, select a region (optional), `Ctrl+s` to save screenshot.

If any dev/maintainer could have a look at it, I am open to suggestions critiques and questions.

### DIFF
--- a/src/utils/filenamehandler.cpp
+++ b/src/utils/filenamehandler.cpp
@@ -15,7 +15,7 @@ FileNameHandler::FileNameHandler(QObject* parent)
 {
     auto err = AbstractLogger::error(AbstractLogger::Stderr);
     try {
-        std::locale::global(std::locale(""));
+        std::locale::global(std::locale());
     } catch (std::exception& e) {
         err << "Locales on your system are not properly configured. Falling "
                "back to defaults";
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -182,6 +182,9 @@ target_link_libraries(
         Qt5::Widgets
         ${QTSINGLEAPPLICATION_LIBRARY}
         QtColorWidgets
+        X11
+        Xext
+        Xfixes
 )
 
 if (APPLE)
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -54,6 +54,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initAutoCloseIdleDaemon();
 #endif
     initPredefinedColorPaletteLarge();
+    initIncludeCursor();
 
     m_layout->addStretch();
 
@@ -88,6 +89,7 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
     m_screenshotPathFixedCheck->setChecked(config.savePathFixed());
     m_uploadHistoryMax->setValue(config.uploadHistoryMax());
     m_undoLimit->setValue(config.undoLimit());
+    m_includeCursor->setChecked(config.includeCursor());
 
     if (allowEmptySavePath || !config.savePath().isEmpty()) {
         m_savePath->setText(config.savePath());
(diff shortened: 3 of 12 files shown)


borgmanJeremy:
Thanks for working on this, but this implementation will not work because it is not cross platform. Can you describe the issue with using QCursor? My thought was you could record the the position of the cursor when the screenshot is taken, and then draw a cursor using: https://doc.qt.io/qt-5/qcursor.html

Maybe you can draw inspiration from ksnip or spectacle: https://github.com/ksnip/ksnip/blob/0895df40541fd103dfb5b1846a1d59d0c8e1ac8e/src/widgets/CustomCursor.cpp

AndreaSenseon:
The issue with `QCursor` is that you can get the `bitmap()` or the `pixmap()` only if you constructed your `QCursor` object from a `QBitmap` or `QPixmap` respectively. 

> My thought was you could record the the position of the cursor when the screenshot is taken, and then draw a cursor using: https://doc.qt.io/qt-5/qcursor.html

The code does that (get the cursor image and draw it on top of the screenshot), but the cursor is taken from the OS. 
I mean we can get an image of a cursor and draw it on top..worst case scenario we can literally draw a cursor on top of the screenshot..
Thoughts ?

borgmanJeremy:
Yeah I think we should literally draw a picture of a cursor.

borgmanJeremy:
I think it should also be implemented like most other tools so it can be moved

borgmanJeremy:
I would look at the circle counter for a place to start, should be pretty similar

AndreaMarangoni:
@borgmanJeremy could you have a look now? This looks more like what you were expecting I think. Again, I am open to any comment or suggestion. (Actually I might be able to test on Mac soon).

AndreaMarangoni:
Testing on Mac OS and it is working fine. Not sure if you guys are aware though, but on Mac the preview of the tools looks very nasty for big 'toolSize'. The preview looks like a big red rectangle that follows the mouse. Only circle counter looks fine as it is probably meant to have a preview.

borgmanJeremy:
Can you open an issue with that macOS problem?

Btw thanks for the good work, I've been busy triaging the v11 release but should get to this soon.

AndreaSenseon:
> Can you open an issue with that macOS problem?

Sure!

>Btw thanks for the good work, I've been busy triaging the v11 release but should get to this soon.

No worries. I'll convert the Draft into a proper PR soon.

AndreaMarangoni:
@borgmanJeremy FYI https://github.com/flameshot-org/flameshot/issues/2280

src/utils/filenamehandler.cpp
```diff
@@ -15,7 +15,7 @@ FileNameHandler::FileNameHandler(QObject* parent)
 {
     auto err = AbstractLogger::error(AbstractLogger::Stderr);
     try {
-        std::locale::global(std::locale(""));
+        std::locale::global(std::locale());
```
AndreaMarangoni, line 18: #2283

borgmanJeremy:
Sorry for the delay and confusion on this. My thought was not that we should add another icon to the tray to have the user manually add a cursor. I'll try explain better this time:

- In the config menu a user should enable or disable if a cursor should be drawn
- If the cursor is enabled, a cursor should be drawn at the location the cursor was _when the screenshot was taken_
- The cursor implementation should stay the same as it is now. IE behind the scenes it is a tool, and users can move, scale, or delete it.

TheRedXD:
I happen to have a windows machine. Could test it if I have time, tho use linux as a daily driver lol

PR #2284, opened 2022-01-19
working on wayland clipboard

borgmanJeremy:


borgmanJeremy:
Added the ability to build against KSystemClipboard. This is a clipboard provided by the KDE project that resolves wayland issue. Unfortunately it is only present in very new versions of KF5GuiAddons. Therefore I have only enabled it for flatpak and snapcraft. 

If Fedora, Debian, and Ubuntu want this fix they need to figure out how to package the correct version of KF5GuiAddons.

PR #2295, opened 2022-01-20
Major Controller refactor

veracioux:
Tasks:
2. [X] Extract systray icon code from Controller into its own Widget class
    (currently called SystemTray but will rename to SysTrayIcon)
3. [X] Remove obsolete methods from Controller
4. [X] Add a utils module with an `os` constant
1. [ ] Change Controller interface to completely mimic the CLI (**IN PROGRESS**)
5. [ ] Use the `os` constant where ever appropriate (**IN PROGRESS**)
5. [ ] Rename SystemTray to SysTrayIcon (more appropriate name)
6. [ ] Rename Controller to Flameshot

**Note: I broke check for updates, but will restore it to its old functionality
before this PR is merged.**

**Rationale for item no. 4:**

Currently we check for the OS using preprocessor #if directives, but this has the
following drawbacks:
- Refactoring and usage detection does not work properly for those parts of
  code enclosed in inactive #if blocks
- Errors in inactive #if blocks are not detected until you compile on each OS

Therefore, I added a `Utils::os` constexpr variable that tells us the OS type.
I think this will greatly reduce the number of accidental errors, mismatches
between different systems, code readability. And it comes with a negligible
performance drawback (if any). Of course, preprocessors will still have to be
used in some places (I think I don't have to explain where).

To add more rationale to this PR, here are some things I plan to add in future
PRs:
- Add support to use multiple configs (System, User, custom file via a
    `--config` option which I'll create an issue for)
- Refactor the CLI parsing to make `int main()` more comprehensible, and move
    something to the `Flameshot` class (previously `Controller`)
- Introduce an `Option` class/structure that can be passed to methods inside
  Controller (which will be called Flameshot after this PR). It will also in
  some way replace or encapsulate the current `CaptureRequest`.

I think this PR is also a step towards creating a cleaner API which will make it
easier to devise a plugin API.

Let me know what you think @borgmanJeremy.


src/utils/utils.h
```diff
@@ -0,0 +1,39 @@
...
+ * @brief Operating system type.
+ * Always test for a match using '&' instead of '=='.
+ * @example
+ * ```
+ * Utils::Os os = Utils::MACOS;
+ * bool isUnix = os & Utils::UNIX;  // true (correct usage)
+ * bool isUnix = os == Utils::UNIX; // false (incorrect usage)
+ * ```
+ */
+enum Os
```
borgmanJeremy, line 16: I generally like using a constexpr value for this instead of #ifdefs scattered everywhere. However, I think the API should change to use ==. Testing with & does not feel natural.

borgmanJeremy (changes requested):

src/utils/utils.h
veracioux: @borgmanJeremy On the other hand it allows us to test for `os & UNIX`.
If `os` is `MACOS` then `os == UNIX` will be false.

borgmanJeremy: @veracioux something like this maybe? https://godbolt.org/z/Y4n39qr48

veracioux: @borgmanJeremy I abandoned this idea altogether. It didn't make sense in those places where system-specific includes were needed. With those cases taken out of the picture, the few that were left don't warrant the trouble. Btw, this PR is ready to merge as far as I'm concerned.

borgmanJeremy (approved):

borgmanJeremy:
Overall all it looks good. I like the new interface / way the files are split. Did you confirm that the update checker is fixed?

veracioux:
@borgmanJeremy Yes, the update checker is fixed. And it runs only in the daemon, and if it's enabled in the config.

Issue #564, opened 2019-06-21
Fails when fractional scaling <> 100%


fzaninotto:
Ubuntu 19.04 supports fractional scaling on Wayland and Xorg. When using a factor that is not the default (like 150%, for instance on a HiDPI display), and trying to capture a screenshot with flameshot, the screen flickers an changes resolution. and flameshot can only capture a quadrant of the screen.

dszryan:
please fix this !

flantel:
Any update on this? Since the demise of Shutter, Flameshot is the best tool, but without fractional scaling it is no good to me.

Mart-Bogdan:
@fzaninotto @flantel 
Hello, I tried to make a fix.
Can you please test patched version released here as downloadable built archive with x64 executable
https://github.com/Mart-Bogdan/flameshot/releases/tag/tmp-fix-scaling

flantel:
Hi,

No, unfortunately this does not work for me. The screen keeps flipping around between different resolutions. This is on Ubuntu 19,10/Gnome

Mart-Bogdan:
@flantel you are using 200% scaling? indeed it wasn't working on 2x only on fractional, like 1.5x
New patch should work https://github.com/Mart-Bogdan/flameshot/releases/tag/tmp-fix-scaling-2
I've tested it on Gnome too.

flantel:
Hi @Mart-Bogdan I am using 150% on a Pixelbook, running ubuntu 19.10/Gnome/X

Result is the same - screen will flip between small and normal and continues to do so until I press escape.

adampiskorski:
Same issue on Pop_OS! on both an Eve V (surface pro look-alike) and an Matebook X Pro, both at 150%.

CumpsD:
Same here on Pop_OS Surface Laptop 2 at 150% Impossible to take a screenshot, it just toggles constantly big/small/big/small/...

karolbe:
@Mart-Bogdan Hi Bogdan, can you please confirm if your fix for fractional scaling was indeed incorporated into the main flameshot branch? Occasionally I get flameshot updates from OpenSUSE and it is breaking it, so clearly at least SUSE does not have that patch.

Mart-Bogdan:
[...]

PR #2305, opened 2022-01-22
Wayland kde clipboard

borgmanJeremy:


SFBB:
Hello, I found an issue with this merge.

```C++

#ifdef USE_WAYLAND_CLIPBOARD


        mimeData->setImageData(formattedPixmap.toImage());


        mimeData->setData(QStringLiteral("x-kde-force-image-copy"),


                          QByteArray());


        KSystemClipboard::instance()->setMimeData(mimeData,


                                                  QClipboard::Clipboard);


#else
```

This code assumes all Linux Wayland DE like gnome to use a KDE addon, which doesn't exist at all. And it will make them not be able to save the screenshot to clipbord as a result.

Other users also met similar issue.
https://github.com/flameshot-org/flameshot/issues/564#issuecomment-3243250890

borgmanJeremy:
If this flag is enabled, we expect this dependency to be installed: https://github.com/KDE/kguiaddons

If a distro using gnome is enabling that flag, but not adding kguiaddons as a dependency thats an issue with the distro.

Issue #2270, opened 2022-01-17
Saving screenshot is the wrong format (the one above in the list)


mellon85:
### Flameshot Version

Flameshot v11.0.0 (ad1bf28)
Compiled with Qt 5.15.2
darwin: 20.6.0
osx: 11.6

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Big Sur 11.6.2

### Description

When saving a screenshot and I have to select the format it's always off-by-one. Starts with tiff selected and if saved it would save it as PNG, if I select jpeg it will be jp2 instead, etc etc

if I select the topmost (bitmap), it keeps using bitmap.

### Steps to reproduce

1. take any kind of screenshot
2. press on save
3. select a format, it will use the one above the selected 

### System Information

MacOS Big Sur 11.6.2

borgmanJeremy:
Thanks, I can confirm this.

AndreaMarangoni:
I can take this

borgmanJeremy:
Great thanks for the help

AndreaMarangoni:
This looks like a bug on Qt. Even the smallest of the sample has that issue.
The only fix I found is to don't use the native dialogs setting the option `QFileDialog::Option::DontUseNativeDialog`. 
Thoughts?

borgmanJeremy:
If you have a very minimal example can you try building against a newer Qt version? Perhaps even Qt6? I dont really want to use the DontUseNativeDialog option, it makes the Dialog box look terrible on most platforms.

AndreaMarangoni:
I have tried against QT 6.2.2 and it has the same behaviour unfortunately

borgmanJeremy:
I wonder if it's a bug related to auto generating the file formats. In 10.2 we specifically enumerated a few types, but it 11 we auto generate all possible types. #2307 uses the non-native dialogs which we used once before to fix a small bug and got a significant amount of negative feedback.

This is the commit that changed the file format behavior: 3cad8084c73e1a3cbde96a5e0ab8766bdf2bbef5

AndreaMarangoni:
Don't think so. I have created a new project which replicates what we are trying to do in production and tested with different versions of Qt.

This is the sample used.

```
#include <QFileDialog>
#include <QImageWriter>

int main(int argc, char *argv[])
{
    QApplication a(argc, argv);
    const QString extension{"png"};
    QString mimeFilter("image/png");
[...]

PR #2307, opened 2022-01-22
QFileDialog misbehaves when choosing images format

AndreaMarangoni:
It looks like the behaviour described in #2270 is a Qt BUG. I can see it in `Qt v5.15.2`. 
Using non-natives dialog seems the only way of fixing this at the moment.

Also fixed the following warning:
```
.../src/utils/screenshotsaver.cpp:142:16: Don't call QList::first() on temporary [clazy-detaching-temporary]
```


borgmanJeremy:
Can you try #2379

PR #2333, opened 2022-01-27
reworked info window

borgmanJeremy:


borgmanJeremy:
@veracioux Here is the simplest menu reworked with qt ui, overall I really like how it declutters the cpp files. 

I think i might restructure the folders a little so all the .ui files are in the same folder. I also would like to merge this in stages rather than one monster commit. 

Any thoughts?

veracioux (approved):
Yep, much cleaner IMO. I tried it out and it looks and works as it did before.

borgmanJeremy:
Sweet, I'll start doing the other widgets, saving the config widget for last since there are a couple open PR's that affect that one.

PR #2336, opened 2022-01-28
Opencv

vohefwdskcmigw:
Making Flameshot full app, something like paint. I do not know is it even possible but that was just my idea. You could add adding own images.


borgmanJeremy:
Can you help me understand the purpose of this change? It's not mergable as is since the branch point is nearly 4 years old.

PR #2357, opened 2022-02-03
Add alternative shortcuts file for KDE Flatpak installs

Proton-459:
Alternative shortcuts file for KDE installs from Flatpak - CommandURL needs to be org.flameshot.Flameshot, rather than flameshot. Update readme to include a link to the alternative file - step by step instructions left unchanged to avoid reducing the clarity.


mmahmoudian:
I wonder if it is easier for the user to create a symlink (for example in `/usr/share/bin`) that links to `org.flameshot.Flameshot`. This way they can use flameshot normally and we don't need to maintain more than one shortcut file for KDE

mmahmoudian (approved):

Proton-459:
Thanks!

Proton-459:
Sure, I can add it to the installation document.

PR #2378, opened 2022-02-06
Zooming in/out happens at different speed

AndreaMarangoni:
What I have done: 
- Removed magic number `ex. +15/-15` 
- Calculating the necessary steps (in pixels) from the `QWheelEvent`.
- Using same aspect ratio during zooming in/out
- Refactoring

I have tested it on `Ubuntu 18.04` and `Mac OS Monterey`, but if someone could kindly test this on other devices that would be great.


borgmanJeremy (approved):
Looks good. Thanks!

PR #2379, opened 2022-02-06
fix bug on macos with save dialog

borgmanJeremy:
There is a bug on on MacOs. When the mime types are populated Qt reports HEIC and HEIF as seperate types. Really they are the same thing except HEIF also supports video. 

The MacOs native dialog always shows heif/heic, so it was causing an off by one error. This fix removes HEIF from the list.


borgmanJeremy:
Closes #2270

PR #2382, opened 2022-02-06
better zsh code completion

mmahmoudian:
I reviewed multiple code completions including `arch-audit`, `pamac`, and `nvm`. I ultimately ended up finding the `pamac` style and flow much more cleaner and more compatible with our style of CLI. So I adopted that and modified it according to Flameshot's CLI subcoimmands and args.

Unfortunately this is not compatible with bash and therefore I renamed the one @veracioux made to be only for bash, and added this one for zsh.

I'm [as always] open to any suggestion/comments/feedback 😃


borgmanJeremy (commented):
Thanks!

Issue #1875, opened 2021-09-09
The shortcut to take a screenshot is broken on KDE


rapiz1:
**Flameshot version**

 0.10.1-1

**Describe the bug**

Pressing the shortcut to take a screenshot doesn't work on KDE.
**To Reproduce**

1. Open KDE Settings / Shortcuts
2. Add Flameshot
3. Set "Take a screenshot" with key PrintScr
4. Press PrintScr. Nothing happens.

**Expected behavior**

A screenshot will be taken
**System Information**

Operating System: Arch Linux
KDE Plasma Version: 5.22.5
KDE Frameworks Version: 5.85.0
Qt Version: 5.15.2
Kernel Version: 5.13.13-arch1-1 (64-bit)
Graphics Platform: X11
Processors: 4 × Intel® Core™ i7-6600U CPU @ 2.60GHz
Memory: 19.4 GiB of RAM
Graphics Processor: Mesa Intel® HD Graphics 520

mmahmoudian:
Would you please post here a screenshot of the command tab of the shortcut you created?

rapiz1:
![Screenshot_20210909_155703](https://user-images.githubusercontent.com/20105061/132646423-7e0aa0d9-64c6-4ad2-9cf6-13469b7dc842.png)

rapiz1:
BTW, it was working until recently update, like 3 months? However, I'm not using Arch much during that time so I'm not sure which update broke this.

mmahmoudian:
Try this:
![image](https://user-images.githubusercontent.com/390889/132672031-dc720b56-008b-47dc-9d5b-42aaf8c1e063.png)

rapiz1:
Thanks. This works.
However, I think some users may still face the original issue.

mmahmoudian:
I don't know why that issue happened to you as we don't/can't create global shortcuts for the window managers (e.g KDE Plasma).

But since sunrise you had that, I don't see why it has stopped working as we haven't had any breaking changes in our CLI. If you open a terminal and run `flameshot gui` it should work and this form of command has existed perhaps since versions < 0.5.x

mmahmoudian:
Since this is the first time such thing is reported and as it is not clear hire such simple thing has gone wrong, I suggest closing this issue for now. If anyone faces the same issue, they can reopen this thread and add a bit more information so that we can track is down and fix it if it is Flameshot's issue.

mindyourlifeguide:
[...]

PR #2383, opened 2022-02-06
KDE Plasma khotkeys

mmahmoudian:
At the moment we have a bug when `flameshot gui --path ''` is invoked and that caused one of the shortcuts to fail. Considering that in [the description of that shortcut](https://github.com/flameshot-org/flameshot/blob/a87d4469d4015142a26715db628dabe23d94865f/README.md?plain=1#L240) there was no mention of saving to disk, I removed that from the command.


borgmanJeremy:
Thanks! Does this also close #1875

PR #2395, opened 2022-02-08
fix: arrow tool glitches

UnkwUsr:
Demo (on the left is fixed version, on the right is original flameshot):

https://user-images.githubusercontent.com/49063932/153045211-3ddb3d0a-fa93-48d0-a132-c5043e35e5de.mp4


UnkwUsr:
Also check comment in code. The problem is that if we set length to 0, then "line" will be just 1 "pixel" (but real size is much bigger, because we set pen size before, in function ArrowTool::process (code block attached)), so we lose information about angle.

https://github.com/flameshot-org/flameshot/blob/bcb4041e0e433e6ac0d65e87bd50c194459f5684/src/tools/arrow/arrowtool.cpp#L148

borgmanJeremy:
Thanks!

PR #2495, opened 2022-03-20
Make KDE use Freedesktop portal

### LINKED ISSUE #2436: Wayland does not support QWindow::requestActivate() on KDE Plasma
### Flameshot Version

Flameshot v11.0.0 (-)
Compiled with Qt 5.15.2
linux: 5.16.8-200.fc35.x86_64
fedora: 35

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...), Flatpak from Flathub

Fedora 35 with KDE Plasma 5.23.5 on Wayland

### Description

The application starts without an itch, but it fails to take any screenshot afterward. Problem is similar to issue #619, and as far as I can see was fixed under Sway, but is still encountered under KDE Plasma on Wayland.

### Steps to reproduce

1. Run `flameshot`
2. Click on the tray icon to take a screenshot
3. Visually nothing happens, but on the terminal you can see the following messages:
```
Unable to get current screen, starting to use primary screen. It may be a cause of logical error and working with a wrong screen.
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
```

I also tried running `flameshot full`, it seems to take the screenshot but fails when saving the file with the following message:
```
kf.kio.widgets.kdirmodel: No node found for item that was just removed: QUrl("file:///home/snyssen/Pictures/28382-gunar-streu-midwinter-morning.jpg")
```
(the mentioned file is the first that appear inside the folder when ordered alphabetically, and it has nothing to do with the actual screenshot)
I doubt the bugs are correlated, but I wanted to include it in case it could be.

_No response_

### System Information

Output from Neofetch:
-------------- 
OS: Fedora Linux 35 (KDE Plasma) x86_64 
Host: XPS 15 7590 
Kernel: 5.16.8-200.fc35.x86_64 
Uptime: 1 hour, 6 mins 
Packages: 2187 (rpm), 28 (flatpak) 
Shell: zsh 5.8.1 
Resolution: 3840x2160 
DE: Plasma 5.23.5 
WM: kwin 
Theme: Layan [Plasma], Adwaita [GTK2], Breeze [GTK3] 
Icons: [Plasma], Tela [GTK2/3] 
Terminal: konsole 
CPU: Intel i7-9750H (12) @ 4.500GHz 
GPU: NVIDIA GeForce GTX 1650 Mobile / Max-Q 
GPU: Intel CoffeeLake-H GT2 [UHD Graphics 630] 
Memory: 17899MiB / 31721MiB

greenfoo:
This fixes #2436

It looks like in the newest KDE, using the Fredesktop portal is the only way that works (at least that seems to be the case with Fedora 35, where I have tested this change).

Having said so... it **could be** that this ticket breaks compatibility with older versions of KDE (after all, the special KDE code this PR is replacing was originally put in there for a reason!).

So... there are two options here: either merge this PR and forget about backwards compatibility or create a more sofisticated logic which (for the KDE case) tries with different approaches until one succeeds (KDE Screenshot / KDE Screenshot2 / Freedesktop / ...).

I'm not personally a fan of looking backwards... but you decide :)

After this change, all 3 supported desktops (GNOME, KDE and Sway) use the Freedesktop portal... and I *suppose* we could even go as far as dropping the Desktop check from this function and always try to use the Freedesktop portal as this will probably also work with others.

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -100,25 +100,8 @@ QPixmap ScreenGrabber::grabEntireDesktop(bool& ok)
         QPixmap res;
         // handle screenshot based on DE
         switch (m_info.windowManager()) {
-            case DesktopInfo::GNOME: {
-                freeDesktopPortal(ok, res);
-                break;
-            }
-            case DesktopInfo::KDE: {
-                // https://github.com/KDE/spectacle/blob/517a7baf46a4ca0a45f32fd3f2b1b7210b180134/src/PlatformBackends/KWinWaylandImageGrabber.cpp#L145
-                QDBusInterface kwinInterface(
-                  QStringLiteral("org.kde.KWin"),
-                  QStringLiteral("/Screenshot"),
-                  QStringLiteral("org.kde.kwin.Screenshot"));
-                QDBusReply<QString> reply =
-                  kwinInterface.call(QStringLiteral("screenshotFullscreen"));
-                res = QPixmap(reply.value());
-                if (!res.isNull()) {
-                    QFile dbusResult(reply.value());
-                    dbusResult.remove();
-                }
-                break;
-            }
+            case DesktopInfo::GNOME:
+            case DesktopInfo::KDE:
             case DesktopInfo::SWAY: {
                 freeDesktopPortal(ok, res);
                 break;


borgmanJeremy:
Couple of questions since I'm having trouble testing on wayland right now.

1. When using the portal do you get the prompt to "share" the screen with flameshot?
2. Why not use https://invent.kde.org/plasma/kwin/-/blob/master/src/effects/screenshot/org.kde.KWin.ScreenShot2.xml

greenfoo:
> 1. When using the portal do you get the prompt to "share" the screen with flameshot?

No :) It just takes the screenshot without asking.
This is also the case in Sway (ie. wlroots based compositors).

Out of curiosity I've been looking at the source code of xdg-desktop-portal and this is how it seems to work:
1. There is a "frontend" component in charge of receiving dbus requests (ie. "org.freedesktop.portal.Desktop")
2. When a request to take a screenshot is received, it is forwarded to a "backend" implementation depending on the running compositor:
     1. The ["backend" implementation for GNOME](https://github.com/flatpak/xdg-desktop-portal-gtk) seems to [**always** show the pop-up](https://github.com/flatpak/xdg-desktop-portal-gtk/blob/347fcd1f61e126ae77950b5d4b66ad9522aec1f2/src/screenshot.c#L197) asking the user whether he wants to allow a screenshot to be taken (only "native" GNOME applications, such as the built in screenshot tool, [can bypass this dialog](https://github.com/flatpak/xdg-desktop-portal/issues/649) using an internal GNOME API (?)).
     2. The ["backend" implementation for KDE](https://github.com/KDE/xdg-desktop-portal-kde) seems to [**never** show the pop-up](https://github.com/KDE/xdg-desktop-portal-kde/blob/master/src/screenshot.cpp#L97). I've read (can't find the link now) that it shouldn't be possible and that you shoud grant explicit permissions to the app trying to take the screenshot by placing a special string ("X-KDE-DBUS-Restricted-Interfaces=...") on the app .desktop file, but I found no trace of this limitation in the code and was able to run flameshot without any extra pop-up (maybe it only applies to the older org.kde.KWin.ScreenShot{,2} interfaces?)
     3. The ["backend" implementation for Sway/wlroots](https://github.com/emersion/xdg-desktop-portal-wlr) seems to [**never**  show the pop-up](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screenshot/screenshot.c#L25) either. It simply ends up running "grim" which uses one experimental Wayland protocol (only supported by wlroots it seems) to take the screenshot without asking for confirmation.

> 2. Why not use https://invent.kde.org/plasma/kwin/-/blob/master/src/effects/screenshot/org.kde.KWin.ScreenShot2.xml

It looks like the most recent "consensus" is to move to the Freedesktop interface, which is considered "the way to go in the future" (or, at least, that's what they [end up doing in Gimp's screenshot tool](https://gitlab.gnome.org/GNOME/gimp/-/issues/5785#note_1100355)):

gimp

Notice, however, that [at the end they still kept the KDE portal](https://gitlab.gnome.org/GNOME/gimp/-/commit/1ba2e11c91780ea99b51eb3dff4d4f9fb6f15cb9) in case the Freedesktop one failed (instead of completely dropping it). Something similar could be done here... or not, considering the Freedesktop portal should be available in all except very old distros.

borgmanJeremy:
Thanks for documenting all that! It's really helpful to have it all summarized here. 

Since this code path only is activated on Wayland, it makes sense to me to just use the Freedesktop implementation. If users are on old distros they are likely using X11.

@mmahmoudian any objections before I merge this?

mmahmoudian:
> there are two options here: either merge this PR and forget about backwards compatibility or create a more sofisticated logic which (for the KDE case) tries with different approaches until one succeeds (KDE Screenshot / KDE Screenshot2 / Freedesktop / ...).
> 
> I'm not personally a fan of looking backwards... but you decide :)

Well, in spite of my preference on being on the latest version of software, we should see which if the functionality breaks in major KDE-based point-based release distros i.e kubuntu, Fedora KDE spin. I will try these two distros today and will report back.

One question that comes to mind is that of this backwards compatibility can be mitigated by using Flatpak or Snap.

In general I have no objections and I support the cause, but we should identify which major popular distros will break and we create a clear announcement to avoid confusion and avalanche of new bug reports.

borgmanJeremy:
Thanks I tested this on Fedora36 beta and this change is required.

borgmanJeremy:
This also worked on Arch, but importantly I had to uninstall gnome first. There was some conflict while having the xdg portals of both gnome and kde installed at the same time.

PR #2503, opened 2022-03-25
Allow users to override imgur key

borgmanJeremy:


mmahmoudian:
This is the error I got when changed one character of the default API key (the one in upper-case is what I changed to: `313baf0c7b4F3ff`):

image

Is there a way to validate an API key?

veracioux:
@

mmahmoudian:
User gets this message after they have selected the screenshot area and potentially annotated it. Basically it causes the user to lose all the progress. What I suggest is that we validate the key upon change on the settings so that the user gets the error right at that time and before potentially losing valuable information.

borgmanJeremy:
> User gets this message after they have selected the screenshot area and potentially annotated it. Basically it causes the user to lose all the progress. What I suggest is that we validate the key upon change on the settings so that the user gets the error right at that time and before potentially losing valuable information.

Yeah I can do that. I'll have it try to upload a one pixel image when the key changes and then return an error if that fails.

borgmanJeremy:
> > User gets this message after they have selected the screenshot area and potentially annotated it. Basically it causes the user to lose all the progress. What I suggest is that we validate the key upon change on the settings so that the user gets the error right at that time and before potentially losing valuable information.
> 
> Yeah I can do that. I'll have it try to upload a one pixel image when the key changes and then return an error if that fails.

Actually its going to be more work than I expected to validate it. I'm going to punt on that given the recent plugin discussion.

PR #2518, opened 2022-04-01
Adds instruction to open on macos after instaling using homebrew

ricardovsilva:


mmahmoudian:
I would suggest adding a link to our website for troubleshooting macOS as well:

https://flameshot.org/docs/guide/troubleshooting/#macos

In general I have started feeling that our README has got super lengthy and overwhelming, and we might need to break it into some pages.

borgmanJeremy (approved):

Issue #1007, opened 2020-09-29
[Feature request] Pin tool Enhance: copy pinned picture


vio1etus:
**Is your feature request related to a problem?**
I'm always frustrated when I want to copy the pinned image.

add the copy pinned picture function

borgmanJeremy:
Related to #954

toofar:
This was implemented in #2519 via a context menu so I think this issue can be closed now!

Also I think the top post of #954 could be edited to cross out the item about this functionality and link to the merged PR.

(As an aside, that's kind of a grab bag issue, but out of everything discussed in it it seems the only thing discussed there that doesn't have its own PR is "improve the usability/discoverability/UX of pinned image".)

mmahmoudian:
@toofar thanks for pointing these out

PR #2519, opened 2022-04-02
pinwidget: allow copying to clipboard and saving to file

zhangfuwen:
add a context menu in pinwidget for copying to clipboard and saving to file

### DIFF
--- a/src/tools/pin/pinwidget.cpp
+++ b/src/tools/pin/pinwidget.cpp
@@ -6,10 +6,12 @@
 
 #include "pinwidget.h"
 #include "qguiappcurrentscreen.h"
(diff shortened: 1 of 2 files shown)


mmahmoudian:
This addresses #1007

veracioux:
@borgmanJeremy I agree that we should show purple bubbles, but the best, most consistent and maintainable way to do that is by adding a context menu item that turns the pin widget into a windowed capture GUI.

But for now, I like the context menu implemented in this PR.

borgmanJeremy (approved):

Alaskra:
When I try to save the pinned picture, the pinned picture covers the file manager and I can't move it away.

Issue #165, opened 2018-03-11
[Feature] Add inverted "focus" blur


CurtisBelt:
* Work just like current blur tool, but invert the effect.
* Handy to give complete focus on the selected area in a single click
* Can currently accomplish this manually with blur tool, but with several carefully aligned blur rectangles.

Example result:

![image](https://user-images.githubusercontent.com/4251760/37259216-808ea9dc-2551-11e8-9919-6b030d544526.png)

AndydeCleyre:
Yes, this would be very useful. In terms of UI, maybe invert while the shift key is pressed and the user is dragging the box?

ZYZ64738:
Something new to this?

instead (or additionally) of blurring a kind of lightbox effekt would be also nice:

![image](https://user-images.githubusercontent.com/9817678/166465525-0074e765-e8e8-4a04-ab1c-401984b1302c.png)

borgmanJeremy:
This needs to wait for plugins

PR #2527, opened 2022-04-06
add invert selection feature for pixelatetool

Alaskra:
close https://github.com/flameshot-org/flameshot/issues/165.
add a checkoutbox in toolpanel to toggle invert selection for pixelatetool.

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -490,6 +490,7 @@ void CaptureWidget::paintEvent(QPaintEvent* paintEvent)
     if (m_activeTool && m_mouseIsClicked) {
         painter.save();
         m_activeTool->process(painter, m_context.screenshot);
+        update();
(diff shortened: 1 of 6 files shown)


src/widgets/capture/capturewidget.cpp
```diff
@@ -490,6 +490,7 @@ void CaptureWidget::paintEvent(QPaintEvent* paintEvent)
     if (m_activeTool && m_mouseIsClicked) {
         painter.save();
         m_activeTool->process(painter, m_context.screenshot);
+        update();
```
Alaskra, line 493: If I don't add update() here, the painting outside the selection rect will not flush until mouse move through.I'm new to qt5 and don't sure it will influence performance.

borgmanJeremy:
We are earnestly starting the plugin design so I think this should be deferred to a plugin. https://github.com/flameshot-org/flameshot/discussions/2529

PR #2549, opened 2022-04-15
Fix pinwidget save

Alaskra:
fix the [bug](https://github.com/flameshot-org/flameshot/pull/2519#issuecomment-1098142508)


mmahmoudian:
Out if curiosity, wouldn't toggling the `Qt::WindowStaysOnTopHint` be feasible?

Alaskra:
Actually the `Qt::X11BypassWindowManagerHint` flag have the effect of `stay on top`. If pinned picture is partially off-screen, switch `Qt::X11BypassWindowManagerHint` off and on will cause position change. So I think hide and show is easier.

PR #2609, opened 2022-05-17
Chore: Add instructions for adding keyboard shortcut to Ubuntu 22.04

henetiriki:


mmahmoudian:
@henetiriki Thank you for the PR. I just wonder if we should emphasis on the fact that this is Gnome-specific and it would be exactly the same regardless of distro. Right?

mmahmoudian:
Let's merge it for now, and then we have to probably create sub pages for each DE and move these instructions there. Our README is getting extremely lengthy and overwhelming for users.

mmahmoudian (approved):

borgmanJeremy:
Thank you!

PR #2615, opened 2022-05-21
Added the ability to cache the last region

borgmanJeremy:


borgmanJeremy:
@mmahmoudian I still need to review the code a bit more before merging but this adds an option that enables re-using the last region if you want to test it .

mmahmoudian:
I have few points to raise, so I enumerate them to be able to easily discuss them:

1. I don't think this is a correct approach to have this enabled by default. At least I find it very disturbing that every time I click on the tray icon the same position is selected. I think the better approach is to have one of the following:
    1.1. a checkbox in the configurations to be able to toggle this feature
    1.2. only limit this to Launcher

I personally think the latter makes more sense as the user can also see the geometry.

I additionally suggest having the following (simultaneously):

- a CLI argument that let the user to spawn this way (e.g `--last-region`). This would be useful as one-off mode and to be bound to keyboard shortcuts
- if we go with the Launcher, we should add a checkbox to make the FLameshot ignore the geometry there. This way the user can choose how to use the Launcher.

2. the behavior is inconsistent between when the `flameshot gui` is executed, and when tray icon is clicked. I believe they should behave identically to avoid confusion.

borgmanJeremy:
> 1.1. a checkbox in the configurations to be able to toggle this feature

This exists today, but I'm fine to make it disabled by default:

image


> a CLI argument that let the user to spawn this way (e.g --last-region). This would be useful as one-off mode and to be bound to keyboard shortcuts

Good suggestion, I'll add this. 

> the behavior is inconsistent between when the flameshot gui is executed, and when tray icon is clicked. I believe they should behave identically to avoid confusion.

Can you describe the difference? They behave the same for me



I first had this feature confined to only the launcher, but then I added the config option that enables it globally. So the behavior is the launcher **always** remembers the last region, but flameshot gui / tray icon can be optionally enabled or disabled.

src/config/generalconf.cpp
```diff
@@ -235,6 +242,19 @@ void GeneralConf::initShowHelp()
       m_helpMessage, &QCheckBox::clicked, this, &GeneralConf::showHelpChanged);
 }
 
+void GeneralConf::initSaveLastRegion()
+{
+    m_saveLastRegion = new QCheckBox(tr("Use last region"), this);
+    m_saveLastRegion->setToolTip(tr("Uses the last region as the default "
+                                    "selection for the next screenshot."));
```
mmahmoudian, line 249: None of the others end with a full stop.

src/main.cpp
```diff
@@ -50,7 +50,7 @@ void requestCaptureAndWait(const CaptureRequest& req)
     Flameshot* flameshot = Flameshot::instance();
     flameshot->requestCapture(req);
     QObject::connect(flameshot, &Flameshot::captureTaken, [&](QPixmap) {
-        // Only useful on MacOS because each instance hosts its own widgets
+        // Only u seful on MacOS because each instance hosts its own widgets
```
mmahmoudian, line 53: "u seful" -> "useful"

mmahmoudian:
> This exists today, but I'm fine to make it disabled by default:

You are right, Somehow I unconsciously expected it to be added to the bottom of the list of checkboxes. My bad.

 > Can you describe the difference? They behave the same for me

On my computer, when I run `flameshot gui` it will not respect the new config checkbox (i.e same behavior as before), but when I click on the tray icon it does.

> So the behavior is the launcher always remembers the last region, but flameshot gui / tray icon can be optionally enabled or disabled.

I'm fine with this. Sounds logical to me, especially if we get that new CLI argument, users will be much less dependent on the launcher for this feature in one-off mode. So in my mind, this is perfect :+1:

borgmanJeremy:
@mmahmoudian  are you aborting the screenshot without copying to clipboard or saving? I only save the previous region when a final action happened. I did all my testing on MacOS but the gui cli option is behaving the same as the tray. I do further testing on Linux.

edarague:
Thank you very much, it is a feature that many people were clamoring for... finally!! Now I will just use FlameShot.  🎉🥳

Adolio:
Thanks a lot 😀 That's awesome! 🎉

PR #2619, opened 2022-05-22
improve print geometry final action

Alaskra:
#2444 remove the accept button when `flameshot gui --print-geometry`, which is confusing(compared to other final action). I think keep the accept button and other final action button maybe better. @borgmanJeremy

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -272,11 +272,10 @@ void CaptureWidget::initButtons()
 {
     auto allButtonTypes = CaptureToolButton::getIterableButtonTypes();
     auto visibleButtonTypes = m_config.buttons();
-    if ((m_context.request.tasks() == CaptureRequest::NO_TASK) ||
-        (m_context.request.tasks() == CaptureRequest::PRINT_GEOMETRY)) {
+    if (m_context.request.tasks() == CaptureRequest::NO_TASK) {
         allButtonTypes.removeOne(CaptureTool::TYPE_ACCEPT);
         visibleButtonTypes.removeOne(CaptureTool::TYPE_ACCEPT);
-    } else {
+    } else if (m_context.request.tasks() != CaptureRequest::PRINT_GEOMETRY){
         // Remove irrelevant buttons from both lists
         for (auto* buttonList : { &allButtonTypes, &visibleButtonTypes }) {
             buttonList->removeOne(CaptureTool::TYPE_SAVE);


mmahmoudian:
I don't see you point. The `flameshot gui --print-geometry` does not have the Accept button, but the  `flameshot gui --print-geometry --clipboard` does, which to me makes total sense. The `--print-geometry` is not a final action and should not be treated as such. The only thing that the `--print-geometry` would conflict with is `--raw` as both use stdout for their output.

Also you have changed an if...else... to if....else if.... without any else, which is odd to me. I haven't fully checked the whole code, but do we really have two conditions here? if yes, then the "else if" is redundant, if no, then we should also catch the other conditions. Am I missing something?

Alaskra:
I thought `--print_geometry` is final action, because the document says:
image

Alaskra:
Maybe write in this way is better:
```
if (m_context.request.tasks() == CaptureRequest::NO_TASK) {
    allButtonTypes.removeOne(CaptureTool::TYPE_ACCEPT);
    visibleButtonTypes.removeOne(CaptureTool::TYPE_ACCEPT);
} else if (m_context.request.tasks() == CaptureRequest::PRINT_GEOMETRY) {
    // do nothing if only use --print_geometry argument
} else {
    // Remove irrelevant buttons from both lists
    for (auto* buttonList : { &allButtonTypes, &visibleButtonTypes }) {
        buttonList->removeOne(CaptureTool::TYPE_SAVE);
        buttonList->removeOne(CaptureTool::TYPE_COPY);
        buttonList->removeOne(CaptureTool::TYPE_IMAGEUPLOADER);
        buttonList->removeOne(CaptureTool::TYPE_OPEN_APP);
        buttonList->removeOne(CaptureTool::TYPE_PIN);
    }
}
```
If it is final action, have additional accept button when `flameshot gui --print-geometry` maybe necessary.

mmahmoudian:
> I thought `--print_geometry` is final action, because the document says:

You are absolutely right, the website is wrong. At the moment if one wants to use this as a Final Action, they can do `flameshot gui --print-geometry --accept-on-select` which it _almost_ have the same effect.

> Maybe write in this way is better:

Perhaps, although previously there was `||` which was practically doing the same, except now in this suggestion the Accept button can coexist with all other Final Action buttons if `--print_geometry` is used. and I winder what happens when one run `flameshot gui --print_geometry --raw` and `flameshot gui --print_geometry --clipboard`

-------

I believe this is a conceptual issue and should be discussed. Practically what you showed with this PR is that (at least in my head) some of the definitions don't match the documentation, which is a good thing to know. Discussing it would potentially change some definitions, but regardless of what the decision is, some Documentations should be changed.

Let's ask for @borgmanJeremy and @veracioux opinions on this matter.

veracioux:
I agree with @mmahmoudian. I see --print-geometry more as a "helper" functionality for scripting rather than an action the user would use as part of the normal flow.

Alaskra:
I agree that `--print-geometry` is not a final action, because it actually don't export the picture to other place like other final actions.

borgmanJeremy:
Should we close this since everyone agree's --print-geometry is not a final action? We need to fix the doc's of course.

Alaskra:
ok

PR #2629, opened 2022-05-26
Note about potential freezes

hissssst:
I've stumbled across this issue several times, and some of my friends have also experienced this one. This issue was there for several months, so it's time to create a note about it, or think about the compatibility list of WMs and enviroments where flameshot works fine

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,9 @@
   </p>
 </div>
 
+### Potential freezes
 
+**Versions 10.x.x to 11.0.0 may freeze your system when you're saving the screenshot.**
 
 ## Preview
 


borgmanJeremy:
This seems to only happen to users using a DIY distro and that do not properly meet all the flameshot dependencies

hissssst:
@borgmanJeremy , where are these flameshot dependencies stated?

hissssst:
If you're talking about [these ones](https://github.com/flameshot-org/flameshot#dependencies=), than my system meets all the requirements. It's actually not an efficent way to close the discussion before investigating the problem

hissssst:
As I've stated in body of this PR, currently flameshot has no compatibility lists or anything like this.

hissssst:
That's a great idea!

mmahmoudian:
@borgmanJeremy 

> with a list of who to contact 

I'm not sure if I understood this. What do you mean by who to contact? Do you mean that for sway contact Mr. Y and for NixOS contact Ms. X?

borgmanJeremy:
Yes, I'm imagining people sign up to adopt a specific DE or distro. Then if an issue comes in for gentoo for example we can @ that person.

mmahmoudian:
It's good to have such a phone book, but I don't think people would get volunteered for that. Perhaps we can put together such list ourselves based on package maintainers of each distro (at least we know Nix and Gentoo maintainers). But getting such list for DEsis a bit though.

Do you have any approach in mind to attract these people?

borgmanJeremy:
@mmahmoudian nope, I just know that I ignore issues these days for stranger setups and focus on more standards setups.

PR #2634, opened 2022-05-30
Suggest setting XDG_CURRENT_DESKTOP if DE cannot be detected

greenfoo:
First of all, let me say this is a stupid PR, so feel free to ignore it if you want :)

Now, some context: in sway, you are responsible for setting environment variable "XDG_CURRENT_DESKTOP=sway" before running it.

The "problem" is that if you are using a logging manager such as GDM, you need to manually edit the "/usr/share/wayland-sessions/sway.desktop" *which is overwritten every time sway is updated* (at least in Arch), which happens every few months.

Well... this is the third time this has happened to me and each time I completely forgot what the issue was and had to compile a new version, attach the debugger and step through the code to figure out what the problem was (yes... I must have the worst short term memory in the world).

Anyway, in this PR I simply add a message suggesting to check environment variable XDG_CURRENT_DESKTOP when the desktop environment cannot be detected in the hope that this will save me (and probably others) ~1 hour in a few months once sway is updated again.

Thanks!


mmahmoudian:
Shouldn't the env vars be set somewhere other than .desktop file?

https://github.com/swaywm/sway/wiki/Setting-Environmental-Variables

borgmanJeremy (approved):

borgmanJeremy:
Thank you!

greenfoo:
> Shouldn't the env vars be set somewhere other than .desktop file?
> 
> https://github.com/swaywm/sway/wiki/Setting-Environmental-Variables

Thanks! I had no idea how to set environment variables for gdm (which is a systemd service). Turns out "man environment.d" is exactly what I was looking for :)

Issue #2623, opened 2022-05-24
Consider adding number bubble pointers


dsirov:
### Feature Description

I think it could be great if, in addition to regular number bubbles, we could have pointing number bubbles.
Example:
![image](https://user-images.githubusercontent.com/25965216/170062920-f0f804b4-0cf2-438c-9ab5-6346ee5ec859.png)
![image](https://user-images.githubusercontent.com/25965216/170063022-03a5519c-33cb-46a4-ab34-62a87f5feef3.png)

I should be made possible to modify the pointing arrow:
![image](https://user-images.githubusercontent.com/25965216/170062701-8aa8be04-56ec-4943-861e-685ea0967a4c.png)

Keep doing great job!

[Idea taken from [ksnip](https://github.com/ksnip/ksnip)]

mmahmoudian:
Considering this can be easily achieved with the combination of bobbles and arrows, I don't really see the point:

![image](https://user-images.githubusercontent.com/390889/170261745-ac367f92-54ee-4101-a378-ca6689b201c1.png)

Let's see what other devs think about this.

borgmanJeremy:
@mmahmoudian I do the same thing as you with the arrows.

However, for silly technical reasons the number tool inherits from the "AbstractTwoPointTool". This is the same parent class that draws lines and rectangles. We just ignore the mouse release on the number tool.

Therefore it would be pretty straightforward to implement this feature by using the ignored second click to draw the triangle. Also if a user does not not want the new behavior they can just click once with no drag.

mmahmoudian:
@borgmanJeremy is it feasible to have a prototype to test it? I'm not opposing this behavior you mentioned, but I wonder how annoying it will be if the person click duration is a bit longer. I just don't want to break a currently working tool.

vozdeckyl:
Hi @mmahmoudian @borgmanJeremy,
Can I can try to implement this, so we can see how annoying it is?

mmahmoudian:
@vozdeckyl 
Considering [the comment Jeremy wrote in the thread](https://github.com/flameshot-org/flameshot/issues/2511#issuecomment-1138953339), I'll assigned this to you

borgmanJeremy:
Yes please. Thanks for the help

vozdeckyl:
Hi @mmahmoudian, @borgmanJeremy,
[...]

PR #2638, opened 2022-06-01
Dev bubble pointers

vozdeckyl:
Issue #2623.


borgmanJeremy (approved):

borgmanJeremy:
Thanks! Code looks good and the behavior feels right. The fact that it doesnt draw until you pull the cursor outside the bubble makes the behavior the same for users who do not want to draw the arrow.

PR #2688, opened 2022-06-20
Update win-installer banner

adem4ik:
Current banner image is too small and overpixelated even on 1600x900 resolution with 96 dpi. Here is my try to make it look better.

P.S. I'll provide comparison screenshots later, after the GitHub's build.


adem4ik:
I've missed with resolution. Going to fix with the next commit
image

borgmanJeremy:
Thanks! Looks great.

PR #2675, opened 2022-06-17
Another way to fix Launchapp unexpectly close

Alaskra:
Another way to fix #2562 . Maybe this way is better compared to  #2617.

Issue #2687, opened 2022-06-20
Regression: crash on pinning screenshot to the desktop


adem4ik:
### Flameshot Version

Flameshot v12.0.0 (adb58e24)

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Windows 10 x64

### Description

Flameshot crashes after trying to pin screenshot to the desktop. I've got the installer from https://github.com/flameshot-org/flameshot/actions/runs/2520717952 . It seems like the issue may be result of this commit - adb58e247b1c733210788d95d25891160d678153 . 12.0.rc1 doesn't have such issue.

### Steps to reproduce

1. run flameshot
2. make a screenshot
3. try to pin it to the desktop

### Screenshots or screen recordings

![image](https://user-images.githubusercontent.com/4707112/174584535-7fb4812c-201f-4155-bf6b-c4d421f116c0.png)

### System Information

Windows 10, single monitor with 1600x900 resolution.

borgmanJeremy:
Thanks, I've revered that commit. I'll wait to mess with this until after the release.

adem4ik:
Thank you, checked with artifacts from https://github.com/flameshot-org/flameshot/actions/runs/2529556850 , no more crashes.

PR #2691, opened 2022-06-20
Revert "Another way to fix Launchapp unexpectly close"

borgmanJeremy:
Reverts flameshot-org/flameshot#2675


borgmanJeremy:
Yes #2687

Issue #2562, opened 2022-04-25
Open in external application (a.k.a App Launcher) closes immediately


mmahmoudian:
### Flameshot Version

Flameshot v11.0.0 (03af9d80)

### Installation Type

User repository (AUR)

### Operating System type and version

Manjaro KDE (Kernel: 5.15.32-1-MANJARO x86_64)

### Description

Regardless having the daemon running or not, the App Launcher window immediately closes.

### Steps to reproduce

0. Try once with the daemon and once without
1. `flameshot gui`
2. select some region
3. press the App Launcher button
4. if the App Launcher window opens for a fraction of a second and closes immediately

### Screenshots or screen recordings

https://user-images.githubusercontent.com/390889/164700780-668a5b40-47be-4425-9e11-bbd42673a1cf.mp4

### System Information

```console
❯ inxi --system --graphics --width=80
System:
  Host: precisiontower Kernel: 5.15.32-1-MANJARO arch: x86_64 bits: 64
    Desktop: KDE Plasma v: 5.24.4 Distro: Manjaro Linux
Graphics:
  Device-1: Intel CometLake-S GT2 [UHD Graphics 630] driver: i915 v: kernel
  Device-2: AMD Lexa XT [Radeon PRO WX 3200] driver: amdgpu v: kernel
  Display: x11 server: X.Org v: 1.21.1.3 driver: X:
    loaded: amdgpu,modesetting gpu: amdgpu resolution: 1: 1920x1080~60Hz
    2: 1080x1920~60Hz 3: 1080x1920~60Hz
  OpenGL: renderer: AMD Radeon Pro WX 3200 Series (POLARIS12 DRM 3.42.0
    5.15.32-1-MANJARO LLVM 13.0.1)
    v: 4.6 Mesa 21.3.8
```

LordMZTE:
I'm also getting this on arch with the `flameshot` packge from the arch repos.

ramseyrashid:
I'm experiencing the same on Ubuntu 22.04. I tried the build from apt, snap and latest RC. All do the same as noted above.

Alaskra:
I find it's #2003 that cause the problem using `git bisect`. I'm now trying to fix it. @mmahmoudian

Alaskra:
https://github.com/flameshot-org/flameshot/blob/0870eca81f61f233e3c96b2d7f9a7f2b9ad7a145/src/main.cpp#L48-L57
[...]

PR #2698, opened 2022-06-22
fix issue about externalWidget launchapp

Alaskra:
Sorry to introduce other bug when fixed bug #2562 . The first one is the hack that open and close window to trigger exit. It will mess window size in window manager like dwm, so I replace the hack with exit flameshot directly. But I make a mistake so it also close the daemon. This time I add the daemon judge logic, and put it in `captureTaken`'s slot under the MAC_OS code to keep consistent.

### DIFF
--- a/src/core/flameshot.cpp
+++ b/src/core/flameshot.cpp
@@ -37,6 +37,7 @@
 
 Flameshot::Flameshot()
   : m_captureWindow(nullptr)
+  , m_haveExternalWidget(false)
 #if defined(Q_OS_MACOS)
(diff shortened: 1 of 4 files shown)


adem4ik:
Well, https://github.com/flameshot-org/flameshot/issues/2687 doesn't reproduce with this PR. Also I haven't found any new issues yet.

I can test it for day or two if you need. @Alaskra mb you can suggest any specific test-cases?

borgmanJeremy (approved):

Issue #2718, opened 2022-06-24
Confusing setting name for custom Imgur key


thehunmonkgroup:
### Flameshot Version

Flameshot v12.0.0 ()
Compiled with Qt 5.12.8

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Pop!_OS 20.04 LTS

### Description

https://github.com/flameshot-org/flameshot/pull/2503 added the ability to configure a custom Imgur key.

However, I got pretty tripped up about *what* to put in there. The Imgur documentation talks about needing to use an access token to make write requests to an account, so I spent quite a bit of time generating Oauth access tokens and putting them in the `Imgur API Key` field, which of course kept failing with permission denied errors.

It was only after I looked at the actual code in https://github.com/flameshot-org/flameshot/pull/2503 that I realized it was the `Client ID` from the created Imgur app that was supposed to go in that setting.

Given the crazy amount of complexity in the Oauth sequence, and the number of different ID/secrets/tokens generated (Client ID, Client secret, Access token, Refresh token), I think a more precise label for that setting would be in order, perhaps `Imgur Application Client ID` ?

### Steps to reproduce

N/A

### Screenshots or screen recordings

N/A

### System Information

N/A

borgmanJeremy:
Great suggestion!

PR #2719, opened 2022-06-24
rename Imgur API Key to Imgur Application Client ID

thehunmonkgroup:
Per https://github.com/flameshot-org/flameshot/issues/2718

I didn't mess with the translations other than the rename.

### DIFF
--- a/data/translations/Internationalization_bg.ts
+++ b/data/translations/Internationalization_bg.ts
@@ -1154,8 +1154,8 @@ Please solve them manually in the configuration file.</translation>
     </message>
     <message>
         <location filename="../../src/config/generalconf.cpp" line="540"/>
-        <source>Imgur API Key</source>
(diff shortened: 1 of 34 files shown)


mmahmoudian:
I think the translation should be changed as well. that is it of the scope of this PR, but I think the translations should be wiped so that the get redone by the community

thehunmonkgroup:
I did change the existing English versions of the phrase, but feel
completely unqualified to actually translate the term in other languages.

I don't think that should hold up this PR, as it looks like quite a bit of
the translation for that phrase is still incomplete.

thehunmonkgroup:
Makes sense. I'd still vote that removal should be a separate PR, just to keep it clean.

borgmanJeremy:
In general we should NOT refresh the .ts files, but ill merge this as is.

PR #2737, opened 2022-06-30
Added ability to rotate pinned imaged

borgmanJeremy:
Added ability to rotate pinned image from pin context menu.


mmahmoudian:
I was thinking if we should add icons to the context menu. If you think it is a valid idea, I can open a feature request

mmahmoudian:
Shall we add this to v13 milestone or do you think it needs more work?

borgmanJeremy:
Yeah it can be added to v13, 99% done

Issue #2708, opened 2022-06-23
Can't open with Gimp


wd5gnr:
### Flameshot Version

12.0rc1 (installed from Ubuntu 20.x deb).

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

KDE Neon

### Description

Installed from .deb. Gimp is a flatpak but there is a link to the var/export for the flatpak that will start gimp from the shell (as gimp).

When I try to open with, the Open With dialog shows up and has two Gimp icons, but neither work. Okular, Gwenview, etc. are also there and they work.

### Steps to reproduce

1. Take a screenshot
2. Press Open With button
3. Observe 2 Gimp icons
4. Press either one
5. Gimp does not open

### System Information

Up to date Neon. Gimp as a flatpak with a link in ~/bin that will open the flatpak named gimp.

X11 display.

borgmanJeremy:
Are Okular or Gwenview a flatpak?

wd5gnr:
No. However, the menu shows two copies of Gimp, it just won't run them. In addition there is a script on my path: 

[alw:~] $ cat /home/alw/bin/gimp
───────┬──────────────────────────────────────────────────────────────────────
       │ File: /home/alw/bin/gimp
       │ Size: 85 B
   1   │ #!/bin/sh
   2   │ exec /usr/bin/flatpak run --branch=stable --arch=x86_64 org.gimp.GIMP "$@"

borgmanJeremy:
Can you add a debug statement to your script to print the arguments it is getting and confirm the script is executing at all?

Under the hood flameshot is using QProcess to call your script and pass an argument to a temp file: https://github.com/flameshot-org/flameshot/blob/1f3ca56b977832cbf403a336c51d56a3da55cd8c/src/tools/launcher/applauncherwidget.cpp#L113

wd5gnr:
Yes I have done that. It isn't launched at all. I tried the AppImage version -- same result. I also moved ~/bin/gimp (the script that runs flatpak) to /usr/local/bin and no change. The dialog shows two Gimps (no telling where those are coming from) and neither of them appear to launch anything.

I do have two icons in the app menu and changing the title of one of them did change the title, but zero response. Both icons are attached to something like: flatpak run ....

I deleted one of them and flameshot still shows two, however, it maybe a restart will take away the 2nd one. But nothing even appears to happen when you select one of those. 

[...]

PR #2741, opened 2022-06-30
Properly constructs external app command line

wd5gnr:
See Issue #2708

The code to launch an external tool assumes the desktop file has a command like:

/usr/bin/XXX %f

It fails if the command is more like:
/usr/bin/XXX --important-option %f

The update looks for a % in the string. If so, it replaces %. (which is probably %f; this is how the original code did it) with the file name. If there is no %, it just appends the file name. Then it builds a proper command argument for QProcess::startDetached. That involves removing the program name and passing the array of arguments. 

Because the arguments are in an array you do NOT want to quote it (as the original code does) because it is not going through the shell so trying to open "X.png" literally looks for a file with the name <quote>X<dot>png<quote>.

### DIFF
--- a/src/tools/launcher/applauncherwidget.cpp
+++ b/src/tools/launcher/applauncherwidget.cpp
@@ -96,11 +96,19 @@ void AppLauncherWidget::launch(const QModelIndex& index)
             return;
         }
     }
-    QString command = index.data(Qt::UserRole)
-                        .toString()
-                        .replace(QRegExp("(\\%.)"), '"' + m_tempFile + '"');
+    // Heuristically, if there is a % in the command we assume it is the file name slot
+    QString command = index.data(Qt::UserRole).toString();
+    // no quotes because it is going in an array!
+    if (command.contains("%"))
+        command = command.replace(QRegExp("(\\%.)"),  m_tempFile );
+    else
+    {
+         // we really should append the file name if there
+        command.append(m_tempFile); // were no replacements
+    }
 
-    QString app_name = index.data(Qt::UserRole).toString().split(" ").at(0);
+    QStringList prog_args = command.split(" ");
+    QString app_name = prog_args.at(0);
     bool inTerminal =
       index.data(Qt::UserRole + 1).toBool() || m_terminalCheckbox->isChecked();
     if (inTerminal) {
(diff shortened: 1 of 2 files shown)


wd5gnr:
Apparently, I forgot to switch branches so while I meant to do the .desktop file fix separate, it is all bundled into one. Sorry.

mmahmoudian:
> Because the arguments are in an array you do NOT want to quote it (as the original code does) because it is not going through the shell 

Sorry I'm on the phone and I don't remember this from the top of my head, so I'm just throwing it here: if the quoting is removed as you did in the first few lines, and we have a file name that has space is the name, will it be handled properly?

wd5gnr:
Yes because it is already digested into an array. I think it could be tested by setting $TMP to something with blanks. I'll try it and report back but I'm 99% sure since leaving the quotes in breaks things. Keep in mind, the quotes are for the shell not for a process.

wd5gnr:
Well, it is odd. With TMPDIR="/tmp/screen shots" I get the following: 

With quotes, Gimp (default flatpak .desktop file) tries to process ${CWD}"/tmp/blahblah"  (with quotes)

Without quotes Gimp tries to process /tmpscreen and $(PWD)/shots. 

However, putting quotes around the %U in the .desktop file doesn't help.

Of course, it is pretty rare that a temp file has spaces in it and the way it is (no quotes) works. With the quotes it breaks (looking for ${PWD}"/tmp/blahblah"

(note: the ${PWD} there is not literal. it is the working directory from when flameshot starts.

I'm going to continue looking, but I do think the way it is is correct, but that the .desktop file is now the culprit which is not exactly upstream but... what... sidestream? lol

wd5gnr:
Ok got it. The problem was the original code substituted the file name for %X (where X is anything) in the string and I kept that code. The original then threw most of that string away including the substituted file and put it back in later (losing the other arguments). In my original fix, I kept the split up string and manipulated it. The problem is, if you have spaces in the file name the split is now incorrect. The new code does the file substitution on each element of the argument array instead of the string. So if element 4 has %f or even "%f" or --file-input=%f then we correctly substitute, spaces and all, into that array element which now does NOT need quoting of its own.

In the case of gimp/flatpak, the line use @@u %U @@ (see --file-forwarding in the flatpak docs). So the "%U" gets the file name and the @@ part takes care of quoting it to Gimp.

wd5gnr:
Oh to test, it is sufficient to launch like this:

TMPDIR="/tmp/screen shots" flameshot

(Of course, don't have flameshot running before that). Then your screenshot file will be in "/tmp/screen shots" and you can test away.

mmahmoudian:
As far as I can read the code it looks ready. Thanks. Now we should wait for other devs to read it thoroughly and approve the changes.

borgmanJeremy (approved):

borgmanJeremy:
Thank you!

borgmanJeremy:
Merged via #2752

PR #2742, opened 2022-06-30
DesktopFileParser only reads .desktop files

wd5gnr:
The current code picks up anything in the application(s) directory including things like editor backups. 

Simple change, but mildly controversial since a very long time ago you might have had .kdelnk files in there.


mmahmoudian:
@wd5gnr no worries, thanks for your investigations, detailed explanations, and contributions.

PR #2749, opened 2022-07-02
remove unused whitespaces, add shell.nix and .envrc

Creator54:
shell.nix : sets up a virtual env with deps. for nix users
.envrc : for users with direnv , just cd into project and build


borgmanJeremy:
Are you planning to maintain the nix environment? I do not want to sign up to keep it up to date as we have too many packaging formats already.

Creator54:
> Are you planning to maintain the nix environment? I do not want to sign up to keep it up to date as we have too many packaging formats already.

This just sets up a build environment for nix/nixos users for manual building. So after cloning if using `direnv` they just cd into the project and start with "build instructions" else cd into project , enter `nix-shell` then build.

Yup will be happy to help if any issues arise.

borgmanJeremy:
Thanks!

PR #2752, opened 2022-07-03
External all cli

borgmanJeremy:


borgmanJeremy:
Applies clang-format for #2741

Issue #2692, opened 2022-06-20
Fail to build on Ubuntu 18.04


TANSixu:
### Flameshot Version

Flameshot v11.0.0(latest)

### Installation Type

Compiled from source

### Operating System type and version

Ubuntu 18.04

### Description

I followed the compilation steps in the instruction with Cmake 3.22.5, however the make procedure failed with the following error information:

[  1%] Automatic MOC for target SingleApplication
[  1%] Built target SingleApplication_autogen
[  2%] Building CXX object external/singleapplication/CMakeFiles/SingleApplication.dir/SingleApplication_autogen/mocs_compilation.cpp.o
[  2%] Building CXX object external/singleapplication/CMakeFiles/SingleApplication.dir/singleapplication.cpp.o
[  3%] Building CXX object external/singleapplication/CMakeFiles/SingleApplication.dir/singleapplication_p.cpp.o
[  3%] Linking CXX static library libSingleApplication.a
[  3%] Built target SingleApplication
[  3%] Automatic MOC and UIC for target QtColorWidgets
[  3%] Built target QtColorWidgets_autogen
[  3%] Automatic RCC for resources/QtColorWidgets/color_widgets.qrc
[  4%] Building CXX object external/Qt-Color-Widgets/CMakeFiles/QtColorWidgets.dir/QtColorWidgets_autogen/mocs_compilation.cpp.o
[  4%] Building CXX object external/Qt-Color-Widgets/CMakeFiles/QtColorWidgets.dir/src/QtColorWidgets/abstract_widget_list.cpp.o
[  5%] Building CXX object external/Qt-Color-Widgets/CMakeFiles/QtColorWidgets.dir/src/QtColorWidgets/bound_color_selector.cpp.o
[  5%] Building CXX object external/Qt-Color-Widgets/CMakeFiles/QtColorWidgets.dir/src/QtColorWidgets/color_2d_slider.cpp.o
[  6%] Building CXX object external/Qt-Color-Widgets/CMakeFiles/QtColorWidgets.dir/src/QtColorWidgets/color_delegate.cpp.o
[  6%] Building CXX object external/Qt-Color-Widgets/CMakeFiles/QtColorWidgets.dir/src/QtColorWidgets/color_dialog.cpp.o
[  7%] Building CXX object external/Qt-Color-Widgets/CMakeFiles/QtColorWidgets.dir/src/QtColorWidgets/color_line_edit.cpp.o
[...]

PR #2758, opened 2022-07-03
restore old QT fallback

theofficialgman:
solves https://github.com/flameshot-org/flameshot/issues/2692

this is the only thing necessary to successfully build on bionic (and other distros with Qt 5.9.X)


borgmanJeremy:
Thanks for the contribution but we intentionally dropped bionic support. We only support the latest two Ubuntu releases.

PR #2766, opened 2022-07-06
Add integrated WXH display and options

wd5gnr:
I added an optional W X H display to the lower left corner of the capture. Hope I integrated with the option system correctly, but it tests OK. There are some small hard-coded margins but the actual semi-transparent background resizes with the size of the text. Of course, it can be turned off with the option (default is on, but that's debatable). 

I know the size is on the button, but this shows while you are dragging. It could make the button superfluous or -- would take more work -- it could only be drawn during dragging operations.

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -740,14 +740,14 @@ void GeneralConf::initshowSelectionGeometry()
     selgeoLayout->addWidget(new QLabel(tr("Display Location")));
     m_selectGeometryLocation = new QComboBox(this);
 
-    m_selectGeometryLocation->addItem("None", GeneralConf::xywh_none);
-    m_selectGeometryLocation->addItem("Top Left", GeneralConf::xywh_top_left);
-    m_selectGeometryLocation->addItem("Top Right", GeneralConf::xywh_top_right);
-    m_selectGeometryLocation->addItem("Bottom Left",
+    m_selectGeometryLocation->addItem(tr("None"), GeneralConf::xywh_none);
+    m_selectGeometryLocation->addItem(tr("Top Left"), GeneralConf::xywh_top_left);
+    m_selectGeometryLocation->addItem(tr("Top Right"), GeneralConf::xywh_top_right);
+    m_selectGeometryLocation->addItem(tr("Bottom Left"),
                                       GeneralConf::xywh_bottom_left);
-    m_selectGeometryLocation->addItem("Bottom Right",
+    m_selectGeometryLocation->addItem(tr("Bottom Right"),
                                       GeneralConf::xywh_bottom_right);
-    m_selectGeometryLocation->addItem("Center", GeneralConf::xywh_center);
+    m_selectGeometryLocation->addItem(tr("Center"), GeneralConf::xywh_center);
 
     // pick up int from config and use findData
     int pos = ConfigHandler().value("showSelectionGeometry").toInt();
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -717,7 +717,7 @@ void GeneralConf::initshowSelectionGeometry()
     int timeout =
       ConfigHandler().value("showSelectionGeometryHideTime").toInt();
     m_xywhTimeout = new QSpinBox();
-    m_xywhTimeout->setRange(0, 1000000000);
+    m_xywhTimeout->setRange(0, INT_MAX);
     m_xywhTimeout->setToolTip(
       tr("Milliseconds before geometry display hides; 0 means do not hide"));
     m_xywhTimeout->setValue(timeout);
@@ -741,8 +741,10 @@ void GeneralConf::initshowSelectionGeometry()
     m_selectGeometryLocation = new QComboBox(this);
 
     m_selectGeometryLocation->addItem(tr("None"), GeneralConf::xywh_none);
-    m_selectGeometryLocation->addItem(tr("Top Left"), GeneralConf::xywh_top_left);
-    m_selectGeometryLocation->addItem(tr("Top Right"), GeneralConf::xywh_top_right);
+    m_selectGeometryLocation->addItem(tr("Top Left"),
+                                      GeneralConf::xywh_top_left);
+    m_selectGeometryLocation->addItem(tr("Top Right"),
+                                      GeneralConf::xywh_top_right);
     m_selectGeometryLocation->addItem(tr("Bottom Left"),
                                       GeneralConf::xywh_bottom_left);
     m_selectGeometryLocation->addItem(tr("Bottom Right"),
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -61,6 +61,7 @@ GeneralConf::GeneralConf(QWidget* parent)
 
     initShowMagnifier();
     initSquareMagnifier();
+    initShowIntegratedWidthHeight();
     // this has to be at the end
     initConfigButtons();
     updateComponents();
@@ -104,6 +105,8 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
 #if defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
     m_showTray->setChecked(!config.disabledTrayIcon());
 #endif
+    m_showIntegratedWidthHeight->setChecked(
+      config.value("showIntegratedWidthHeight").toBool());
 }
 
 void GeneralConf::updateComponents()
@@ -704,6 +707,18 @@ void GeneralConf::initSquareMagnifier()
     m_scrollAreaLayout->addWidget(m_squareMagnifier);
     connect(m_squareMagnifier, &QCheckBox::clicked, [](bool checked) {
         ConfigHandler().setSquareMagnifier(checked);
+    });
+}
+
+void GeneralConf::initShowIntegratedWidthHeight()
+{
+    m_showIntegratedWidthHeight =
+      new QCheckBox(tr("Show integrated width and height"), this);
+    m_showIntegratedWidthHeight->setToolTip(
+      tr("Show Width X Height in corner of capture"));
+    m_scrollAreaLayout->addWidget(m_showIntegratedWidthHeight);
+    connect(m_showIntegratedWidthHeight, &QCheckBox::clicked, [](bool checked) {
+        ConfigHandler().setShowIntegratedWidthHeight(checked);
     });
 }
 
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -712,6 +712,25 @@ void GeneralConf::initSquareMagnifier()
 
 void GeneralConf::initshowSelectionGeometry()
 {
+    auto* tobox = new QHBoxLayout();
+
+    int timeout =
+      ConfigHandler().value("showSelectionGeometryHideTime").toInt();
+    m_xywhTimeout = new QSpinBox();
+    m_xywhTimeout->setRange(0, 1000000000);
+    m_xywhTimeout->setToolTip(
+      tr("Milliseconds before geometry display hides; 0 means do not hide"));
+    m_xywhTimeout->setValue(timeout);
+    tobox->addWidget(m_xywhTimeout);
+    tobox->addWidget(
+      new QLabel(tr("Set geometry display timeout (ms)")));
+
+    m_scrollAreaLayout->addLayout(tobox);
+    connect(m_xywhTimeout,
+            SIGNAL(valueChanged(int)),
+            this,
+            SLOT(setSelGeoHideTime(int)));
+
     auto* box = new QGroupBox(tr("Selection Geometry Display"));
     box->setFlat(true);
     m_layout->addWidget(box);
@@ -744,6 +763,11 @@ void GeneralConf::initshowSelectionGeometry()
     selgeoLayout->addWidget(m_selectGeometryLocation);
     vboxLayout->addLayout(selgeoLayout);
     vboxLayout->addStretch();
+}
+
+void GeneralConf::setSelGeoHideTime(int v)
+{
+    ConfigHandler().setValue("showSelectionGeometryHideTime", v);
 }
 
 void GeneralConf::setGeometryLocation(int index)
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -722,8 +722,7 @@ void GeneralConf::initshowSelectionGeometry()
       tr("Milliseconds before geometry display hides; 0 means do not hide"));
     m_xywhTimeout->setValue(timeout);
     tobox->addWidget(m_xywhTimeout);
-    tobox->addWidget(
-      new QLabel(tr("Set geometry display timeout (ms)")));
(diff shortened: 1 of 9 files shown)


mmahmoudian:
Thanls for the PR. We have had some discussion and feature request about this. I would personally prefer to discuss such change in the issue before having a PR. I haven't tested this yet, so I have no comment on how it behaves, I'm just stating my logic on the development process.

borgmanJeremy:
Yeah I'm generally very in favor of this change. As far as feedback goes do you want to make the changes or do you want @mmahmoudian to discuss and feedback and then I'll implement the change?

Some thoughts I have:
- It would be good to also show the "origin" and print it as WxH+x+y so it matches the --region argument. 
- Someone **will** ask for the color to be configurable
- I think we should remove the button and always show this

Where should this be displayed on the screen? Testing on MacOs right now and I dont see it

wd5gnr:
I'm happy however you like to proceed. The origin should be easy and I
think by button you mean the option?

It should show in the bottom left corner of the select rectangle as part of
the widget painting. It is pretty straight Qt so I'm surprised if it
doesn't work on the mac.

borgmanJeremy:
Sorry by "button" I mean the tool that current is used to show the size. I think this is a much better way to display the size and position info. 

I'll debug the problem on Mac

wd5gnr:
Oh Gotcha. Yeah I agree but figure that will break someone's workflow lol. I thought about the color set and even transparency but didn't feel like implementing a GUI for it. Although I suppose you already have a color picker, I had enough to reverse engineer the settings :-)

borgmanJeremy:
Personally I am fine hard coding the color (or maybe just matching the user theme.) I am just certain that someone will ask for it to be customizable :) I'll defer to @mmahmoudian .

BTW there is likely a trap when getting the origin. You will need to translate between global and local coordinates or I think we will have the opposite problem of #2746

wd5gnr:
I made the change to XxY+H+W. As for removing the buttons, I imagine you have a better handle on how that might break user config etc. As for the colors, I agree it isn't critical, although I bet you are right -- someone will ask lol

wd5gnr:
rats forgot to reformat it... standby.

borgmanJeremy:
I think the problem has to do with fractional scaling / positioning. If I hard code the position it shows up. Here is an example of accounting for the pixel ratio: https://github.com/flameshot-org/flameshot/blob/eb50d8557b78d7c4aee453fe5efaebf7ed76b131/src/widgets/capture/magnifierwidget.cpp#L52

I can do it tomorrow if you can't easily test 200% scaling on linux.

wd5gnr:
You look like you are in a better position to test it. I think I'm at 175 but x11/kde may handle it different. I'd say go for it, and thanks!

mmahmoudian:
Sorry for getting unresponsive, timezone difference.

 
> @wd5gnr: I'm modifying things for my own benefit and it seems like a reasonable feature that other people might like.

Yes, the do like it, and I myself like it as well. Discussing it before implementation helps to collect all the specs needed for that feature, and consequently help saving the developer's time.

> @borgmanJeremy: It would be good to also show the "origin" and print it as WxH+x+y so it matches the --region argument.

I second this. Keeping things consistent is important

> @borgmanJeremy: Someone will ask for the color to be configurable

I think three things we will be asked: color, position, and toggling it

> @borgmanJeremy: I think we should remove the button and always show this
> @wd5gnr: Yeah I agree but figure that will break someone's workflow lol

I also think that the old bubble should be removed, as it is now redundant. I can't imagine how someone's workflow can be dependent on that non-interactive bubble that practically contained half of the info that is implemented in this PR

> @borgmanJeremy: Personally I am fine hard coding the color (or maybe just matching the user theme.) I am just certain that someone will ask for it to be customizable :) I'll defer to @mmahmoudian .

You are right, getting the color from the theme (color of the border of the region is ideal. It can also be more informative as well because it also then shows the relation between the numbers and region with color.

Two last comments:

- can it have auto-hide? At the moment it covers part of the selected region and is disruptive. If we hide it, then it would be put of the way. There can be three strategies to hide it:

    1. Time based (after few seconds) which imho can be hardcoded
    2. When mouse hovers, it gets hidden for ever
    3. When mouse hovers, it gets hidden and when mouse moves away it gets shown again

- can we have it in the middle and center of the selected region? This is the most intuitive way imho, but I don't know how it behaves in fractional scaling.

@thanks again for the PR

wd5gnr:
Honestly, I'm flexible but IMO the transparency means it really doesn't need to hide. You can draw over it (I tested this) which would have been the only deal breaker, again, IMO. But it is certainly possible to hide it. I'm not sure the contrast color is the best color for text either. Maybe a black/white config toggle?

wd5gnr:
> I see your point, but I'm sure we will get too many complaints. people want clean slate. we have already been told that the UI is to cluttered and busy. I'm curious to know @borgmanJeremy opinion on this as well :D

Yeah, I don't doubt you are right. I'm trying to think of the easiest way to implement the hiding, but I'm probably running out of time for today. It is easy enough to flag it since it is all surrounded by an if, but need something to set/clear that flag based on some logic.

wd5gnr:
Ok, here's an attempt at hiding the display. Should work for mouse or keyboard AFAIK. So what's left is to put the position configuration in, auto hide configuration (if desired), and configuration to show/hide position button or remove it all together. Note that I hooked into updateSizeIndicator, so if the button gets removed that function needs to stay although it could be reduced to one line. Could put the autohide delay in config also -- right now hardcoded at 3 seconds

veracioux:
Hey everyone. I think this PR is a welcome addition.
Here are my thoughts about the issues you discussed:
- I don't think the users should have a dying wish to customize the color of the geometry display, since it is only an editing feature and won't end up in the screenshot. With that said, it ought to fit in well with the rest of the aesthetic so setting the background color to be the primary color is the logical choice. However, using the "contrast color" (which may be poorly named anyway) for the text is not a good idea. Because it couples the user's artistic choice to the usability of the geometry display. Look at my config for example:
image
I think here we should use the same strategy we always use to get an appropriate text color based on a given background color: https://github.com/flameshot-org/flameshot/blob/b73608060130b94379f7707825abb7d9b64f8a01/src/utils/colorutils.h#L12
- As for the toggling, here is the best strategy I could think of. I'll describe it via a real-world scenario:
  - User opens flameshot
  - starts dragging the mouse to make a selection - the geometry display appears (reason: discoverability)
  - stops dragging (end of selection) - geometry display hides after 3 seconds (reason: also discoverability, timeout is there to prevent the clean-slate complaint @mmahmoudian mentioned; 3 sec seems reasonable)
  - the geometry display appears again if any of these are true:
    - User is dragging the selection widget in order to move the selection
    - A keyboard shortcut is pressed to either move or resize the selection
    - Mouse is hovering over any of the handles or the selection border (reason: I might want to see the geometry without having to hold down a button).
    - After these conditions are no longer true, the geometry indicator disappears instantly (ergo, the 3sec timeout is active only after the first time the user makes a selection). A fade in/out animation would be a nice bonus. (reason:  a timeout could be annoying to wait for, and is no longer necessary since the user knows about the indicator)

A few more considerations:
- I think the config option should be called `showSelectionGeometry` for better clarity. Or even better `hideSelectionGeometry` since the feature is on by default and I think we should aim for most options to have default values of false.
- I think the best position for the indicator is as it is (bottom right corner), but we have to consider the case when the selection is off-screen. So, if the bottom-right corner is off-screen, show the indicator in the opposite corner (i.e. top-left
- By all means, remove the existing indicator bubble.
- It would be nice to show the indicator outside of the selection if the selection is smaller than the indicator

wd5gnr:
Thanks @veracioux for the comments. I agree the contrast color is wonky so I think that's a good idea. The toggling should work almost as you suggest. I don't have it doing hover, instant disappear, and fade although none of those are bad ideas. Config option name is probably a good idea. As for position, I already have it where you can select 5 different positions (4 corners and middle) but I have not put in a configuration menu for it (help welcome ;) ). I don't have it dynamically moving at screen edges because it is always inside the selection so that would imply the selection is also off screen. Is that a common case? 

I do try to clip for small selections, but that might not be the best approach.

wd5gnr:
@veracioux maybe I did something wrong but I tried

`painter.setPen(ColorUtils::contrastColor(uicolor));`

And it looked rough. Maybe if I decrease the alpha on the background since it is going to disappear anyway...

veracioux:
@wd5gnr Sorry my bad, you should use `painter.setPen(ColorUtils::colorIsDark(uicolor) ? Qt::white : Qt::black);`. Still I think it is a good idea to increase background alpha to something like 70-80%, and make the text bold.

borgmanJeremy:
Been AFK all day, thanks for all the work here. I think this is very close to merging. We can always add some features in a later PR like customizing the location or a timeout. I think this sets the groundwork perfectly.

The change you made to the scaling fixed the issue on MacOS. The only thing I would recommend we change before merging is agreement on the config name so we don't need to change that later.

Thanks again!

wd5gnr:
I made the change but got interrupted while adding the configuration dialog
so have not pushed it yet

wd5gnr:
Ok a slew of changes. Most importantly, there is now one configuration drop down that lets you pick none/tl/bl/br/tr/center position. Someone who is good with the config system should probably check to see that I didn't do it stupidly. If I did, you can tell me ;) and fix it, too, please. The color is now set as @veracioux suggested and I think looks pretty good. I did change the config name but if there is a desire to name it something else, that's cool too.

I'm about to be gone a few days so I really wanted to drop this before I'm on the road and not working in front of a big screen.

wd5gnr:
I like the idea of moving it to a new tab although it might be quicker for someone else to do it since I'm out for a bit. I should look on the laptop because I often forget not everyone has giant screen real estate. 

As for the lag, that's interesting. I haven't noticed it but I wonder if there is a way to profile what's eating time?

A few thoughts... Does setting alpha to 255 speed it up? Is saving the painter expensive? That could be optimized since it is done in several pieces of code. Reading the configuration could be slow? Even without profiling, these could be tested with simple code changes.

mmahmoudian:
>  like the idea of moving it to a new tab although it might be quicker for someone else to do it since I'm out for a bit.

As @borgmanJeremy already said, this can be done in another PR. So I'm not so worried about the placement of the setting for now

> I wonder if there is a way to profile what's eating time

I don't know how to do it. I also don't know an easy way to check my CPU usage as flameshot tends to cover all monitors and create a static image overlay. I don't know the correct term for it, but it seems flameshot gets the highest "zindex". I have even tried to force a terminal on top of others in KDE Plasma and it is still covered by Flameshot. using a Quake style terminal also didn't help.

> Reading the configuration could be slow?

i think the config is read only once upon the initial run, and then it is not parsed/read during a screenshot.

For now the only concern I have is the lag. I wonder if it can be mitigated by some GPU acceleration

wd5gnr:
Can you try the test version I just pushed? It tries to consolidate the painter.save and restore so it is only done once if needed instead of twice (up to twice) now. Other than that, I'm at a loss to why this would be much laggier than without it. The fact that it is still laggy with none is suspicious too since that gates out most of the code during painting (although you say it improves so I'm sure it is more laggy with the code in there, of course).

Great feedback! Thank you.

mmahmoudian:
I'm not on the same computer, but on my laptop, the difference between f948e4e and a0df08c is noticeable. Thanks.

wd5gnr:
Great. Probably before merging the pound if guards need to be removed and
then a big comment at the top so that if anybody adds anything into this
code later they know to adapt the if statement on the save flag.

If someone else wants to do that that's fine or I will try to do it this
evening on the laptop.

wd5gnr:
And of course, I forgot to format yet again... this push has the faster drawing code cemented (actually, it is no faster, but it prevent this code and another piece from both pushing the painting context) and a configurable timeout (in milliseconds) and 0 will tun off timeout. @mmahmoudian will probably have a better suggestion for the option location but to prevent eating up more space, for now, I put it in the scroll area.

wd5gnr:
As a reminder to myself, I realize I need to put TR around the combo box text for internationalization. I also think if the configuration tags are stabilized we should go ahead and use access members instead of direct calls to value like I'm doing now. Other than that, I think it's pretty functional other than perhaps moving the configuration items around as has been discussed.

wd5gnr:
Also should update the example.ini -- sorry I didn't realize that earlier.

src/config/generalconf.cpp
```diff
@@ -104,6 +105,8 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
 #if defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
     m_showTray->setChecked(!config.disabledTrayIcon());
 #endif
+    // m_showSelectionGeometry->setChecked(
```
borgmanJeremy, line 108: Why is this commented out? Should it be deleted or uncommented?

borgmanJeremy (approved):
Minor feedback

wd5gnr:
That was leftover, I think, from the old check mark before the new option
set up so I am sure it can be deleted. Sorry about that.

borgmanJeremy:
Thanks for all the hard work on this. Came out great!

wd5gnr:
Thanks in general conf there are two related comments that should go too

`// m_showSelectionGeometry->setChecked(
    //   config.value("showSelectionGeometry"`

borgmanJeremy:
Ok I'll address that in my next PR. Wanted to get the feature merged ASAP to prevent merge conflict's

wd5gnr:
> 

Perfect. Thanks for all the help getting this ready to go

mikkovedru:
@borgmanJeremy @wd5gnr 
I noticed an inconsistency (Capitalization) in naming.
1. `initshowSelectionGeometry` should be called `initShowSelectionGeometry`.
2. `setshowSelectionGeometry` should be called `setShowSelectionGeometry`.
3. Perhaps `selgeoLayout` -> `selGeoLayout`?

Please merge my PR #2766 first. :-)

wd5gnr:
Most of these probably occurred when I did a global search/replace on the
old names.

borgmanJeremy:
I'll fix it in my cleanup of the comments / old tool

wd5gnr:
I just realized I did a terrible mistake on line 540 because I have the HXW+X+Y backwards (senior moment).
```
  QString xy = QString("%1x%2+%3+%4")
                       .arg(static_cast<int>(selection.left() * scale))
                       .arg(static_cast<int>(selection.top() * scale))
                       .arg(static_cast<int>(selection.width() * scale))
                       .arg(static_cast<int>(selection.height() * scale));
```

Should be:

```
  QString xy = QString("%1x%2+%3+%4")
                       .arg(static_cast<int>(selection.width() * scale))
                       .arg(static_cast<int>(selection.height() * scale))
                       .arg(static_cast<int>(selection.left() * scale))
                       .arg(static_cast<int>(selection.top() * scale));
```
Great example of being focused on one thing and not paying attention to the other. Since this is closed I don't know if I need to push again to the branch or do another PR or @borgmanJeremy, feel free to fix it in cleanup.

borgmanJeremy:
fixed via #2782

Issue #2490, opened 2022-03-16
When an object selected, arrow keys move them instead of changing the region


mmahmoudian:
### Feature Description

In the current implementation (a6144df3), when an object is selected, the user can modify it's properties (change size, color, order/layer), but if an arrow key is pressed Flameshot use that input to change the size of the selected region. I think it would be good if the <kbd>arrow keys</kbd> be used for precise movement of the selected objects. This would also be consistent with what I explained earlier as well.

I additionally propose to have the <kbd>Shift</kbd>+<kbd>arrow keys</kbd> for 10x move. This is the same behavior as in Inkscape and LibreOffice Draw. The only difference is that in LibreOffice Draw:
- the <kbd>arrow keys</kbd> move the object about 5 units/pixels
- the <kbd>Alt</kbd>+<kbd>arrow keys</kbd> move the object about 1 units/pixels
- the <kbd>Shift</kbd>+<kbd>arrow keys</kbd> move the object about 10 units/pixels

certain things are very hard to do in the current form of Flameshot (e.g having multiple arrows originating form the same point) that with the implementation of this feature would become trivial (unless snapping is introduced)

borgmanJeremy:
I like this idea quite a bit.

vozdeckyl:
Hi @mmahmoudian & @borgmanJeremy,
Can I have a go at this? Thanks!

borgmanJeremy:
Yes please 😁

mmahmoudian:
@vozdeckyl thanks for your contributions (past, present and hopefully future) 🤗

vozdeckyl:
Hi @mmahmoudian & @borgmanJeremy,
I've got [something](https://github.com/vozdeckyl/flameshot/tree/dev-lvozdeck-move-objects-with-key) that works, but I haven't implemented the "undo" functionality. Would you like to have that, too? I'm not sure at what point the object should be added onto the undo stack. When one releases the arrow key? At the moment when you press the undo button after moving the object with arrow keys, the object itself is removed (or whichever operation is the most recent is undone).

mmahmoudian:
> I've got something that works, but I haven't implemented the "undo" functionality. Would you like to have that, too

I'm on a road trip and don't have access to my computer to test. In few days I'll be back and can check it out.

> I'm not sure at what point the object should be added onto the undo stack. When one releases the arrow key?

[...]

PR #2769, opened 2022-07-07
WIP: Move objects with keyboard

vozdeckyl:
Implements #2490.

However, it still doesn't work 100%. For some reason, the first move is never recorded and I don't know why.

Steps to reproduce:
1. create an object (rectangle, for example)
2. select the object by clicking on it
3. press Shift+Left _once_
4. press the undo button
5. the object disappears instead of moving back to its original position
6. now repeat steps 1-4, it works this time!


vozdeckyl:
The issue was fixed by using a different method. The movement is initiated by a key press and stopped by a key release. In order to achieve this, the shortcuts for moving/resizing the selection had to be removed, and its functionality had to be hard-coded. The drawback here is that the user will no longer be able to use different keys for moving/resizing the selection.

borgmanJeremy:
Is this ready for review / merging?

vozdeckyl:
Hi @borgmanJeremy, I think so!

PR #2783, opened 2022-07-12
add cmake option DISABLE_UPDATE_CHECKER

### LINKED ISSUE #1706: Allow to disable update checker for the packaged builds
**Is your feature request related to a problem?**

Please allow to disable update checker for the packaged builds with a special compile-time flag. Eg. `-DDISABLE_UPDATE_CHECKER:BOOL=ON`.

Alaskra:
close #1706 .
I just notice that there already exist an option in config window to disable update checker.


borgmanJeremy:
Actually I think the intention of the CMAKE flag is to compile out the networking feature all together not just change the default.

Alaskra:
OK, I will try.

PR #2793, opened 2022-07-14
add cmake option DISABLE_UPDATE_CHECKER 

Alaskra:
close https://github.com/flameshot-org/flameshot/issues/1706 .
allow to disable update checker for the packaged builds with a special compile-time flag: `-DDISABLE_UPDATE_CHECKER:BOOL=ON`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,7 @@ option(USE_LAUNCHER_ABSOLUTE_PATH "Use absolute path for the desktop launcher" O
 option(USE_WAYLAND_CLIPBOARD "USE KF Gui Wayland Clipboard" OFF)
 option(DISABLE_UPDATE_CHECKER "Disable check for updates" OFF)
 if (DISABLE_UPDATE_CHECKER)
-  add_definitions(-DDISABLE_UPDATE_CHECKER="ON")
+  add_compile_definitions(DISABLE_UPDATE_CHECKER)
 endif ()
 
 include(cmake/StandardProjectSettings.cmake)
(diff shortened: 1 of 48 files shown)


Alaskra:
I run `clang-format -i $(git ls-files "*.cpp" "*.h")` and it changes a bunch of files...

Alaskra:
I can't figure out clang-format, the `SplitEmptyRecord` option seems don't take effort. I fallback the wrong formatting manually.

CMakeLists.txt
```diff
@@ -68,6 +68,10 @@ option(GENERATE_TS "Regenerate translation source files" OFF)
 option(USE_EXTERNAL_SINGLEAPPLICATION "Use external QtSingleApplication library" OFF)
 option(USE_LAUNCHER_ABSOLUTE_PATH "Use absolute path for the desktop launcher" ON)
 option(USE_WAYLAND_CLIPBOARD "USE KF Gui Wayland Clipboard" OFF)
+option(DISABLE_UPDATE_CHECKER "Disable check for updates" OFF)
+if (DISABLE_UPDATE_CHECKER)
+  add_definitions(-DDISABLE_UPDATE_CHECKER="ON")
```
borgmanJeremy, line 73: This is not needed. You can reference USE_WAYLAND_CLIPBOARD

borgmanJeremy (approved):

CMakeLists.txt
Alaskra: But it add definition in here:
https://github.com/flameshot-org/flameshot/blob/57c8e8d893669cd83d8b279ea3ff6db50b16aa48/src/CMakeLists.txt#L208-L211
Do I need to change to this way?

Alaskra: And I find `add_compile_definitions` is more recommend.

Issue #1309, opened 2021-02-13
Open With doesn't function on Windows


benc-uk:
**Flameshot version**
0.8.3

**Describe the bug**

The "Open With" appears to do nothing on Windows
![image](https://user-images.githubusercontent.com/14982936/107858811-0bc5de80-6e2e-11eb-9d2d-5ed84e95a8ab.png)

There's no clue what the white box or the text box do
I tried putting in a path to an executable, and hitting enter but nothing happened. There's no clue how this dialog is supposed to work

**To Reproduce**
Click the button to open application while in capture mode

**Expected behavior**
Some way to pick an application or even a path to an exe, and have it launched 

**System Information**
Windows 10

borgmanJeremy:
Thanks for reporting this. The code currently is very platform (linux) specific for "Open With". It needs to be ported to other platforms.

ElTh0r0:
Hi
I have thought about this issue and took ideas from #1868 into consideration.

First I thought about following this [ListPrograms](https://github.com/ElTh0r0/ListPrograms) approach which I found/forked from another Github repository. This really finds all installed software on Windows, but doesn't provide information about the name of the corresponding Exe file (but only the installation folder).

So I was thinking about this approach:
> I'm no coder so I'll only chime in enough to say file extensions and their associated available programs are listed in the registry here:  \HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\FileExts
For example:  .png files shown below
![image](https://user-images.githubusercontent.com/34774124/171482779-bb2cf84e-cab4-42fd-92b6-b30c8861601a.png)

_Originally posted by @resp3ct in https://github.com/flameshot-org/flameshot/issues/1868#issuecomment-1144030964_
This approach provides information about which exe files are associated with image files - but it doesn't tell me where the exe files are located.

My proposal is some kind of porting the Linux approach, which is creating the list of applications by going through all .desktop files. For Windows one could go through all apps, which are listed in the Start Menu (which consists of .lnk files). I created a ["proof of concept application"](https://github.com/ElTh0r0/ListStartMenuPrograms), which collects all global installed (for all users) start menu apps and current users start menu apps + some filtering of apps, which most probably are not interesting for image editing - e.g. Administrative or System Tools. Since Windows .lnk files don't have app categories like Linux .desktop files, I'm combining the apps from the .lnk files with the information from above registry keys, so that apps which are associated to image file extensions are shown first.

Example result of my POC app:
[...]

PR #2801, opened 2022-07-16
Fix OpenWith on Windows (#1309)

ElTh0r0:
On Windows the Applauncher Widget / OpenWith feature was not working. It did not show any application since Windows is not using *.desktop files at all. I "rebuild" the desktop file approach with Windows *.lnk files from the Start Menu.

Since Windows doesn't support app categories, I made a little workaround and fetching graphic applications from the registry, which are assigned to common image file extensions and putting all other found applications in a "misc" category. Additionally Windows doesn't support QIcon::fromTheme, so I had to add icons for the category tab.

I'm a Linux user but using flameshot on Windows at work; means I tested, that my changes are not affecting the Linux build!


data/img/material/white/apps.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px" fill="#FFFFFF"><path d="M0 0h24v24H0V0z" fill="none"/><path d="M4 8h4V4H4v4zm6 12h4v-4h-4v4zm-6 0h4v-4H4v4zm0-6h4v-4H4v4zm6 0h4v-4h-4v4zm6-10v4h4V4h-4zm-6 4h4V4h-4v4zm6 6h4v-4h-4v4zm0 6h4v-4h-4v4z"/></svg>
```
borgmanJeremy, line 1: Did you create all the SVG's yourself? I just want to make sure they are properly licensed.

src/widgets/capture/capturewidget.cpp
```diff
@@ -113,9 +113,12 @@ CaptureWidget::CaptureWidget(const CaptureRequest& req,
         m_context.origScreenshot = m_context.screenshot;
 
 #if defined(Q_OS_WIN)
+// Call cmake with -DFLAMESHOT_DEBUG_CAPTURE=ON to enable easier debugging
+#if !defined(FLAMESHOT_DEBUG_CAPTURE)
```
borgmanJeremy, line 117: Did you test this flag on windows after making this change?

src/utils/winlnkfileparse.h
```diff
@@ -0,0 +1,37 @@
...
+    DesktopAppData parseLnkFile(const QFileInfo& fiLnk, bool& ok) const;
+    int processDirectory(const QDir& dir);
+    QString getAllUsersStartMenuPath();
+
+    QVector<DesktopAppData> getAppsByCategory(const QString& category);
+    QMap<QString, QVector<DesktopAppData>> getAppsByCategory(
+      const QStringList& categories);
+
+private:
+    void getImageFileextAssociates(const QStringList &sListImgExt);
```
borgmanJeremy, line 33: Probably should read getImageFileExtAssociates (capitalize E in Ext)

src/utils/winlnkfileparse.cpp
```diff
@@ -0,0 +1,147 @@
...
+        QFileInfo fiLnk(it.next());
+        if (!regexfilter.match(fiLnk.absoluteFilePath()).hasMatch()) {
+            DesktopAppData app = parseLnkFile(fiLnk, ok);
+            if (ok) {
+                m_appList.append(app);
+            }
+        }
+    }
+
+    return m_appList.length() - length;
```
borgmanJeremy, line 86: Why the subtraction here?

borgmanJeremy (commented):
Thank you! I still need to test it on Windows but the code in general looks good! Thanks for the help, I don't do much windows development so this helps a lot.

data/img/material/white/apps.svg
ElTh0r0: Please excuse, I should have mentioned it. Since flameshot is already using "Google Material Design Icons", I took the new icons from there as well:
- https://fonts.google.com/icons?selected=Material+Icons+Round:apps:
- https://fonts.google.com/icons?selected=Material+Icons+Round:image:

src/utils/winlnkfileparse.h
ElTh0r0: :+1: 
I will change it + proper clang-format soon

src/utils/winlnkfileparse.cpp
ElTh0r0: To be honest: I don't know the exact reason for this return value! I know it returns the number of found applications in the folder, but the return value is (currently) not evaluated (see below). But since I didn't know, if the number of apps could be useful in future, I kept it.

I copied the existing DesktopFileParser and to keep compatibility I used the same return value
https://github.com/flameshot-org/flameshot/blob/57c8e8d893669cd83d8b279ea3ff6db50b16aa48/src/utils/desktopfileparse.cpp#L120-L123

In current applauncherwidget the return value is not processed.
https://github.com/flameshot-org/flameshot/blob/57c8e8d893669cd83d8b279ea3ff6db50b16aa48/src/tools/launcher/applauncherwidget.cpp#L47-L49

borgmanJeremy:
> Just a minor question about the clang-format action. On my machine clang-format (using version 14.0.6) handles this differently: https://github.com/flameshot-org/flameshot/runs/7413928932?check_suite_focus=true
> 
> I didn't use any options, just `clang-format -i [file name]` Do you use any special options?

Unfortunately clang-format has changed behavior across the different versions. I am still trying to figure out the best way to address this.

If I update the version in the CI to v14 (my preference) it makes the work harder for people on old distributions. If I leave it as v12, it makes it harder for people on MacOS / new distros.

Issue #2551, opened 2022-04-18
KDE Connect share intergration


jonibim:
### Feature Description

I think would be a nice touch to have an option on the flameshot gui that directly sends a screenshot to your phone via KDE Connect. Spectacle already has this feature implemented as shown in this image 
![image](https://user-images.githubusercontent.com/16855557/163865607-29fdb2ed-f645-4b5f-afd1-a3b3fa889be9.png)

borgmanJeremy:
Cool, I wonder if this can be integrated with the existing "Open With" feature.

BigFish2086:
Hi, I am using `glibc void-linux` and the proposed fix in #2802 didn't complete the integration with kde-connect for me (not really sure if the problem related to the disto), however, I managed to get it working by crafting this `desktop` file in `/usr/share/applications/kdeconnect_handler.desktop`

```desktop
[Desktop Entry]
Encoding=UTF-8
Name=KDE-Handler
Comment=KDE-Handler
Exec=kdeconnect-handler %U
Icon=kdeconnect
Terminal=false
Type=Application
Categories=GNOME;Application;Development;
StartupNotify=true
```
Now, it appears in the `open with` launcher

![Image](https://github.com/user-attachments/assets/d5507619-7978-4bed-999d-19755cf4804d)

BigFish2086:
another interesting approach is to use a script like this 

```sh
#!/bin/bash

tmpfile=$(mktemp --suffix=.png)
/usr/bin/flameshot gui --raw > "$tmpfile"

# Save and Send
if [[ -s "$tmpfile" ]]; then
  dest=~/Pictures/$(basename "$tmpfile")
  mv "$tmpfile" "$dest"
  device=b29641e9917c472ab4a833f410bc90e4 # Device ID
  /usr/bin/kdeconnect-cli --share "$dest" -d "$device"
else
    rm -f "$tmpfile"
    exit 0
fi

# Get the Device ID
# device=$(/usr/bin/kdeconnect-cli --list-devices | grep '(paired and reachable)' | head -n1 | cut -d: -f2 | awk '{print $1}')
```

accepting the image would save it and send it to your other device. you can bind it to some shortcut on your keyboard for easy access

ElTh0r0:
> Hi, I am using `glibc void-linux` and the proposed fix in [#2802](https://github.com/flameshot-org/flameshot/pull/2802) didn't complete the integration with kde-connect for me

[...]

PR #2802, opened 2022-07-16
Kde-connect share integration (#2551)

ElTh0r0:
By allowing desktop files of Type "Service", issue #2551 would be fixed. Kde-Connect is shown and I was able to send the screenshot successfully to my mobile. Even if the shortcut is called "Open with", the file is copied onto the mobile device.
I don't know if there was a special reason not allowing Type "Service" apps, but at least on my machine this didn't cause any issue.

Send to mobile

I successfully tested it on Manjaro with KDE. Tests on other desktop environments might be recommended.


borgmanJeremy:
Thanks!

Issue #2774, opened 2022-07-08
Can't disable tray icon on Windows


upthekhyber:
### Flameshot Version

12.1.0

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Windows 10 Enterprise 21H2 (64-bit)

### Description

I am unable to disable Flameshot's tray icon by setting `disabledTrayIcon=true` on Windows.
Has this not been implemented in the Windows version by any chance? Any workarounds?

### Steps to reproduce

1. Exit flameshot.exe
2. Open `%AppData%\flameshot\flameshot.ini`
3. Set `disabledTrayIcon=true` under `[General]`
4. Start flameshot.exe
5. Tray icon is still showing up

### System Information

OS: Windows 10 Enterprise 21H2 (64-bit)
Monitor configuration: tested w/ various setups (single 1920x1080, dual 2560x1440, etc.)

manolol1:
I've also tried that, and it starts completely normal (like if it's just ignoring that flag).

OS: Windows 11 Pro 21H2 (64-bit)

fernandofig:
Looking here: https://github.com/flameshot-org/flameshot/blob/f7e41f4d708e50eeaec892408069da25a28e04a2/src/core/flameshotdaemon.cpp#L332

It seems the option is completely ignored on Windows - it'll always create a tray icon. Glancing on some other parts of that source, this option also seems tied to a GUI setting that's hidden / not applicable on Windows builds.

It would seem easy enough to shuffle around those ifdefs and code blocks so the Windows build would respect that setting, but I have no idea if there's any behavior downstream that's specific to the the Windows build that depends on the tray icon, that would consequently break if the FlameshotDaemon::initTrayIcon method would be changed that way.

The compile instructions seems lacking on the Windows front - I'm not sure if that's because it's straightforward to the point of not needing anything but the dependencies, or it's really overlooked. On top of all that, I'm not really a C++ guy, so it would be better if more seasoned developers would have a look at this, at least to assess the difficulty in fixing this problem. @mmahmoudian ?

mmahmoudian:
> it would be better if more seasoned developers would have a look at this, at least to assess the difficulty in fixing this problem. @mmahmoudian ?

I agree, I'm not that seasoned either, so here are people who know better: @veracioux @ZetaoYang @hosiet

Issue #2118, opened 2021-11-25
Implement the same CLI args in Windows build


gabdevbr:
### Flameshot Version

![image](https://user-images.githubusercontent.com/62145041/143483078-da6ef865-e0cf-4c71-8db5-12de09c8a9a9.png)

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Windows 10 Pro

### Description

Implement the same arguments that works in Linux in Windows.

E.g, this command:

`"C:\Program Files\Flameshot\bin\flameshot.exe" gui`

Does not work in Windows 😭 

### Steps to reproduce

1. Press WinKey + R
2. Fill `"C:\Program Files\Flameshot\bin\flameshot.exe" gui`
3. Press enter

### System Information

HOW?

![image](https://user-images.githubusercontent.com/62145041/143483020-7e325257-c923-49be-8f3f-89d25f8274cf.png)

hey-thanks:
Has there been any progress on this? I am having the same issue (Windows 10 Pro, Flameshot version 11 - installed using the ready-made package on GitHub Releases). I'm willing to help out, if possible.

borgmanJeremy:
Nope, feel free to take a crack at it.

jrozner:
Reading through some of the other tickets there is some technicality that needs to be worked around. Can that get documented here? this is something straight forward I may have some time to go and do it.

mmahmoudian:
@jrozner  can you be more specific about the workaround that you want to have in documentations?

jrozner:
https://github.com/flameshot-org/flameshot/issues/2335#issuecomment-1024275873 is the comment I'm referring to. Is there something specific about the command line parsing that doesn't work on windows?

borgmanJeremy:
I dont recall the details of the issue anymore.

davidbrinton:
Just adding my voice of support to this issue. If we're going to use Flameshot in our cross platform application we would have to have the CLI piece working Windows as well as Mac. Sound like it might work in Mac already though.

mmahmoudian:
@davidbrinton it works on any unix-line OS including Linux, MacOS, and BSDs. Please consider commenting in the thread if you have something to add to the discussion (e.g new approach, potential limitations, possible implementation strategies). To show your support of a feature, you can for instance you can like the original post. Thank you.

I'll hide your comment and this response of mine hidden in few hours.

goproslowyo:
Hey @gabdevbr & @jrozner!

[...]

PR #2820, opened 2022-07-23
Fix tray icon handling on Windows (#2774)

ElTh0r0:
Currently all tray icon options are ignored on Windows with #ifdefs, which is not needed from functional point of view. This patch removes these restrictions and user is able to decide if tray icon is wanted or not.

Only disadvantage: Since CLI parsing is currently disabled for Windows, users cannot reach the flameshot config GUI anymore after disabling the tray! So one has to edit the ini manually to restore the tray.
https://github.com/flameshot-org/flameshot/blob/57c8e8d893669cd83d8b279ea3ff6db50b16aa48/src/main.cpp#L149-L153
Disabling the CLI for Windows might not be needed as well. If I find some time, I'll maybe have a look on this and test it (see #2118).

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -102,9 +102,8 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
     if (allowEmptySavePath || !config.savePath().isEmpty()) {
         m_savePath->setText(config.savePath());
     }
-#if defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
+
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
The reason it is #ifdef'ed out is without the CLI parsing in place a user can be left with a half functional version of Flameshot.

Until CLI parsing is supported I do not want to allow windows users to disable the tray. I suspect this will lead to a deluge of bug reports.

PR #2821, opened 2022-07-23
Cppcheck and clazy warning fixes

ElTh0r0:
I hope such kind of code quality improvements are wanted, since I'm touching many files with just "minor" changes.
This patch is mainly fixing warnings related to Cppcheck & Clazy passing args-by-ref / args-by-value, old style Qt signal/slots and using variable names same to function names.


mmahmoudian:
Thanks for the PR. I think this should be discussed. some of the changes seems reasonable to me, but some not so much. For instance:

imho good change:
- `void addToNameEditor(QString s);` -> `void addToNameEditor(const QString& s);`

imho not so good change:
- `const QString& value = *it;` -> `const QString& val = *it;`

ElTh0r0:
> imho not so good change:
> 
>     * `const QString& value = *it;` -> `const QString& val = *it;`

reason for above change is that there is a function with name "value". So either the variable or the function name _should_ be changed:
https://github.com/flameshot-org/flameshot/blob/d2273c657951600c581e1ac287a833b04b72c91c/src/cli/commandlineparser.cpp#L301

I did some more variable renaming due to the same reason in other files.

veracioux (changes requested):
@ElTh0r0 Thanks for contributing to Flameshot. Some of them you should revert - I left comments to let you know which. But I very much appreciate you changing the `SIGNAL(...)` and `SLOT(...)` syntax. This was something I was going to do but never got around to it.

ElTh0r0:
@veracioux many thanks for the detailed review! Please check my changes after the last commits.

veracioux (approved):
@ElTh0r0 Thanks. All looks good now.

mmahmoudian:
@veracioux should we proceed with the merge?

PR #2826, opened 2022-07-24
Added Bold Arrow tool

jkatnik:
The Bold Arrow tool is a variant of the Arrow tool. It is filled with semi-transparent gradient. It is disabled by default and needs to be selected in the configuration window to activate.

image


borgmanJeremy:
I appreciate you adding this feature but I don't think I want this to be a dedicated tool. In the past we have talked about adding a way to modify the tools like right clicking on the tool and selecting an alternate.
@mmahmoudian  what are your thoughts.

mmahmoudian:
I second what @borgmanJeremy said. We also had some discussions about having alternative arrows and arrow heads. I think this would fit that very well, but not as a dedicated independent tool

jkatnik:
I was wondering why Flamshot has over 1k forks. Now I've got a clue :)

mmahmoudian:
This is the reason we always ask our contributors to discuss the features first in an RFC or or in a issue thread. This is the way FOSS development work regardless of the CVS. You cannot wake up one day and create a PR for linux kernel and expect them to merge it. The etiquette dictates that there must be a discussion before hand to get input from maintainers and repo owners.

All these said, Flameshot is Free and Libre Opensource software, and as long as you do not violate the license, you are free to maintain your fork in whatever way you want. Enjoy!

Issue #2091, opened 2021-11-19
Set an ppplication as 'default' whe choosing to "Choose an app to open the capture"


efstratios:
### Feature Description

Hi,

It would be very helpful if the option to "Choose an app to open the capture" would have a 'default' value as an option to set our desired application to open the saved capture every time at the configuration of the application (GUI or File). 

Best regards.

mmahmoudian:
Just to extend the idea, perhaps would be nice if the click opens the default application and the right-click opens the application list.

Pros:
- the right-clicking to open more options would be consistent with the right-clicking on the other buttons

Cons:
- it would break the current behavior
- at the moment for options of other tools we open the sidebar, but for this one it takes the screenshot and opens the window to choose the application

efstratios:
Hello, just to add a comment to your reply @mmahmoudian, when you press the shortcut (Ctrl + O) it should automatically open the capture with the predefined (default) application. 

In my opinion as a user, there is no need to change something in regards to mouse clicks as I think this is not offered in other buttons/options of the flameshot application (as you have already mentioned this is offered by pressing the space key). Please correct me if I am wrong?

mmahmoudian:
> regards to mouse clicks as I think this is not offered in other buttons/options of the flameshot application

It actually does. That was one of the features that was recently added so that users can open settings for each tool by right-clicking on the button of that tool (Basically is equivalent of selecting the tool and pressing space)

lilalkor:
Just wanted to second that request.
I'm using xbackbone to save screenshots, and to upload it uses that "Open with" tool (https://xbackbone.app/clients.html).
And it would be very nice to have it set once to save that additional click.

wd5gnr:
@mmahmoudian Yeah I was going to study that and I think it should be pretty easy to integrate.

wd5gnr:
The right click doing something special needs a total architecture (which would probably be a good idea). The code in CaptureToolButton is OK -- sort of -- but from there the signals fire to CaptureWidget::handleButtonRightClick (and LeftClick). These are all the same for all buttons and it is very hard without coding a specific test in somewhere to identify the magic button.

Because I think this is too big a change, I updated the branch to have the following logic (which is still pretty wacky):
By default there is an option set to true. While true, if you click on the tool button, you get the normal launcher for external program. If you Alt+click you get the default program if set. If not set, it is just like before. And you can set the default from the external program launcher.

[...]

PR #2833, opened 2022-07-26
Allow for default external application

wd5gnr:
As discussed in #2091 this is my take on allowing a default external app. The changes are:

1. Modifier state is captured with mouse input and made available "upstream"
2. By default, there is no default program set and until there is, nothing is different than before
3. When launching an external program, there is a new checkbox to make the program the default
4. If a default program is set, by default Alt+Click on the external tool launches with no dialog; Click operates as usual
5. There is an option to swap the Alt+Click behavior with click (e.g., Alt+Click opens dialog, click opens default)

If there is no default Alt+Click and Click are the same.

We may not need to expose the run in terminal option to the user?

Could probably use better popout text on the help for the option

There are a few other misc changes in here to clean up some of the options stuff from the XYWH change; also a quick write-up on how to add a new option.

Note there is also a new request for the Alt+Click that is consumed as a click by anything else. However, it would be possible to add shift, click, control modifiers to other tools using this much more easily.

### DIFF
--- a/docs/OptionNotes.md
+++ b/docs/OptionNotes.md
@@ -1,0 +1,20 @@
+Notes on adding new options to Flameshot
+========================================
+
+Let's add extPgmTerminal as a boolena:
+
+1. Create the getter/setter in configHandler:
+
+   src/utils/configHandler.h - `CONFIG_GETTER_SETTER(extPgmTerminal,setExtPgmTerminal,bool)`
+
+2. Add the option to the file parser:
+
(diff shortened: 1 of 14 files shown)


docs/OptionNotes.md
```diff
@@ -0,0 +1,20 @@
+Notes on adding new options to Flameshot
+========================================
+
+Let's add extPgmTerminal as a boolena:
```
danielkrajnik, line 4: "boolean" typo

danielkrajnik (commented):
just one typo

Issue #2842, opened 2022-07-27
Function of copy and close window after upload will not work randomly


cavivie:
Bug: https://github.com/flameshot-org/flameshot/blob/001726a565049ebfeb237a3a39b77104899b970e/src/core/flameshot.cpp#L395

Analyze:
`tasks` will first become a random value (which may be the same as the original value), and then assign the new value to itself, so it may lead to random failure of the logic here. To fix it, allocate a new block variable to keep the `tasks` value valid, does not use a same variable name

Fix Code:
```cpp
CR::ExportTask exportTasks = (CR::ExportTask)tasks;
...
if (!(exportTasks & CR::COPY)) {
...
```

cavivie:
Will fix here: https://github.com/flameshot-org/flameshot/pull/2843

PR #2831, opened 2022-07-25
fixed seg fault

borgmanJeremy:
Closes #2830

PR #2843, opened 2022-07-27
Add OCR recognizing support and fix copyAndCloseAfterUpload feature

cavivie:
Feature:
We often need to record the text content in the capture when capturing a screenshot containing some text, so we need to provide an OCR function to recognize the text in the capture.

Addition:
I found a bug in [export task function](https://github.com/flameshot-org/flameshot/issues/2842), it will be fixed in this branch. It is the same to [known issue mention](https://github.com/flameshot-org/flameshot/pull/2831)

Screenshot
ocr-recognizer

PS:
[flameshot-ocr-server](https://github.com/cavivie/flameshot-ocr-server) to recognize text in capture.


borgmanJeremy:
I appreciate the work but I wish you would have reached out first. I personally am not comfortable using a cloud service for OCR and would prefer to see it implemented with tesseract locally (packaging issues aside).

@mmahmoudian thoughts?

cavivie:
@borgmanJeremy I know that, you can run a local ocr server, then config local ocr server url in flameshot configuration, a simple local ocr server like this: https://github.com/cavivie/flameshot-ocr-server

cavivie:
@borgmanJeremy  At the same time, server recognizer is just a implement for ocr recognizer base class, and you can implement a ocr recognizer with local trained text recognize model to work, also implement a ocr recognizer plugin in future waiting for flameshot plugin support.

mmahmoudian:
@borgmanJeremy 
> I appreciate the work but I wish you would have reached out first

I second this

@borgmanJeremy 
> I personally am not comfortable using a cloud service for OCR and would prefer to see it implemented with tesseract locally

I second this too

@cavivie 
> you can run a local OCR server, then config local OCR server url in flameshot configuration

If it is as simple as you said (I haven't yet checked you code and that repo you linked), I'd rather have it configured to use the local by default and give the user the option to opt-in for the cloud one.

@borgmanJeremy the packaging is also yet another real issue that should be thought about. I think in general this would fit nicely as a plugin as not all users need an OCR but it is good to have the option. Unfortunately we still are not there yet with the plugin system. Would it be feasible to integrate this (if the code is proven to be safe and sound) and then later migrate it to a plugin, or would it be better to wait until the plugin interface is ready? @cavivie I'm also interested to know what you think about this as well?

cavivie:
@mmahmoudian 
> I'm also interested to know what you think about this as well?

I agree with it, I think that add the feature to the plugin system is a better choice. Is there a plugin system feature preparing schedule?

mmahmoudian:
@cavivie There is an RFC about it (#2529) that has very well defined and discussed the dynamics, but from implementation perspective I don't think there has been much progress yet. If you want to give that a go, let us know.

borgmanJeremy:
I'm going to close this for now and defer until plugins are ready.

PR #2853, opened 2022-08-01
fixed capturewidget on windows with multiple monitors

ViridiVir:
### Problem:
I have two displays, but when I try to take a screenshot, only one displays

only one monitor gui

### For reproduce:
  -  Place the second display to the left of the main display
  - This will make negative coordinates for him.
  - Then try to take a screenshot and it will set the window top left corner to x: 0 y: 0. But should be in this case at x: -1920 y: 0

### System:
  -   Flameshot v12.1.0 (96c2c82e) Compiled with Qt 5.15.2
  -  windows 10.0.19044

### DIFF
--- a/src/core/flameshot.cpp
+++ b/src/core/flameshot.cpp
@@ -116,7 +116,7 @@ CaptureWidget* Flameshot::gui(const CaptureRequest& req)
         m_captureWindow = new CaptureWidget(req);
 
 #ifdef Q_OS_WIN
-        m_captureWindow->show();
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
Have you tested this on Windows with other monitor configurations to make sure it does not introduce a regression?

toelke:
I cannot reproduce the underlying issue, the window is moved to negative coordinates correctly.

vanjaz:
I have the sam issue.
image

PR #2854, opened 2022-08-03
Recognize Qtile WM

3nprob:
Without this Flameshot won't work on Qtile Wayland.

An alternative would be to stop focuing on individual WMs and replace `SWAY` with `GENERIC_WAYLAND` and assigning recognized strings to that, or something similar, which would remove the need of explicitly having each recognized WM in the enum. Happy to amend the PR according to that if desired.


alba4k:
you have a [typo](https://i.imgur.com/eynQMaP.png) in desktopinfo.h

`QTLE` > `QTILE`

3nprob:
> you have a [typo](https://i.imgur.com/eynQMaP.png) in desktopinfo.h
> 
> `QTLE` > `QTILE`

Oops, good catch. Amended.

borgmanJeremy (approved):

PR #2932, opened 2022-10-01
Add spin box to display the tool size value in the side panel

### LINKED ISSUE #2356: Feature: Show current tool size and gain option to change via text-entry
### Feature Description

It is possible to change current tool size by typing a number. But IMHO:
 - this feature is not immediately visible
 - it would be nice to see the current value
The idea is something like this:

obrázok
with text item being clickable and editable (and obviously mapped to scrollbar)

DMaroo:
This allows to change the tool size from a spin box.

Closes #2356.


borgmanJeremy (approved):

DMaroo:
Can this be merged? Thanks!

borgmanJeremy:
Thank you!

PR #2949, opened 2022-10-15
Fix documentation of `flameshot gui --help`

jfly:
In
https://github.com/flameshot-org/flameshot/commit/6432490c316baa89fe3f5a6910539a1157614577, the format of `--print-geometry` changed from `W H X Y` to `WxH+X+Y`, but it looks like we forgot to update the docs accordingly.


mmahmoudian:
Thanks for the PR

mmahmoudian (approved):

PR #2987, opened 2022-11-17
Fix typos in README

vojta-dev:
Hi!

These are some typos I found in README, not much to say other than that.

If i should change something, let me know. 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,8 +212,8 @@ These shortcuts are available in GUI mode:
 | <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Z</kbd>                                            | Redo the next modification                    |
 | <kbd>Ctrl</kbd> + <kbd>Q</kbd>                                            | Leave the capture screen                                         |
 | <kbd>Ctrl</kbd> + <kbd>O</kbd>                                            | Choose an app to open the capture                                |
-| <kbd>Ctrl</kbd> + <kbd>Enter</kbd>                                            | Commit text in text area|
-| <kbd>Enter</kbd>                                             | Upload the selection to Imgur                                      |
+| <kbd>Ctrl</kbd> + <kbd>Return</kbd>                                            | Commit text in text area|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -212,8 +212,8 @@ These shortcuts are available in GUI mode:
 | <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Z</kbd>                                            | Redo the next modification                    |
 | <kbd>Ctrl</kbd> + <kbd>Q</kbd>                                            | Leave the capture screen                                         |
 | <kbd>Ctrl</kbd> + <kbd>O</kbd>                                            | Choose an app to open the capture                                |
-| <kbd>Ctrl</kbd> + <kbd>Return</kbd>                                            | Commit text in text area|
-| <kbd>Return</kbd>                                             | Upload the selection to Imgur                                      |
+| <kbd>Ctrl</kbd> + <kbd>Enter</kbd>                                            | Commit text in text area|
+| <kbd>Enter</kbd>                                             | Upload the selection to Imgur                                      |
```
mmahmoudian, line 216: The `Return` is also something that is used in the config file. I can see you point that on most keyboards it is written "Enter", and the only outlier is Mac, but having Return there is also to some extent educational (because of the config). Perhaps we should find a way to either accommodate both, or at least for now revert the change?

mmahmoudian (changes requested):

mmahmoudian:
Thanks for the PR. I highly appreciate it. Other than the point I mentioned about Return/Enter issue, the rest looks good to me. How about you revert this change for now and I merge this PR, then in another PR or Discussion we can discuss as a community what would be the best and least confusing approach for users.

README.md
vojta-dev: Understood, reverted.👍

mmahmoudian (approved):

mmahmoudian:
@vojta-dev Thanks. Would you do the honor and create a Discussion or an issue for the Return/Enter part. I will merge this PR.

vojta-dev:
Discussion created! #2988

PR #3010, opened 2022-12-09
added functionality for saving to disk via shift+doubleclick

qoreQyaS:
* added config option
* added handler doubleclick + shift modifier in the doubleclick eventhandler

Issue #2982, opened 2022-11-14
Copying selection to clipboard is not working (neither Ctrl+C hotkey nor the button)


isak102:
### Flameshot Version

Output of `flameshot -v`:
`Flameshot v12.1.0 (-)
Compiled with Qt 5.15.5`

Also tried cloning from git and manually compiling, had the same issue

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...), User repository (AUR), Flatpak from Flathub, Compiled from source

### Operating System type and version

Arch Linux `6.0.8-arch1-1`

### Description

Selecting a screenshot in `flameshot gui` and then pressing Ctrl+C to copy the selection to keyboard doesnt work. Pressing Ctrl+C just exits flameshot. The same thing happens when clicking the button in the interface to copy to clipboard.

I tried installing flameshot as a flatpak and there it worked properly if I ran flameshot as a daemon, but running `flameshot gui` had the same behaviour, copying didn't work. 

### Steps to reproduce

1. run `flameshot gui`
2. Make selection
3. Copy the selection to clipboard, either by pressing Ctrl+C or by clicking the button on screen
4. Nothing happens

### System Information

1. Arch Linux, 6.0.8-arch1-1
2. Output of xrandr:
```Screen 0: minimum 320 x 200, current 2496 x 1404, maximum 16384 x 16384
eDP-1 connected primary (normal left inverted right x axis y axis)
   1920x1080     60.01 +  60.01    59.97    59.96    59.93
   1680x1050     60.01    59.95    59.88
   1400x1050     59.98
   1600x900      59.99    59.94    59.95    59.82
   1280x1024     60.01    60.02
   1440x900      60.01
   1400x900      59.96    59.88
   1280x960      60.00
   1440x810      60.00    59.97
   1368x768      59.88    59.85
   1280x800      59.99    60.01    59.97    59.81    59.91
   1280x720      60.00    60.01    59.99    59.86    59.74
   1024x768      60.01    60.04    60.00
   960x720       60.00
   928x696       60.05
   896x672       60.01
   1024x576      59.95    59.96    59.90    59.82
   960x600       59.93    60.00
   960x540       59.96    59.99    59.63    59.82
   800x600       60.01    60.00    60.32    56.25
   840x525       60.01    59.88
   864x486       59.92    59.57
   700x525       59.98
   800x450       59.95    59.82
   640x512       60.02
   700x450       59.96    59.88
[...]

PR #3011, opened 2022-12-10
added functionality to save to disk on shift+double click

### LINKED ISSUE #2877: Double click to save the image to the save path or copy it to the clipboard
### Feature Description

This tool is the best one I've seen on ubuntu. It would be great if this tool allows the user to save the image to the save path, or copy to the clip board by double clicking the screen. The behavior of double click can be configured in the config GUI. Such interaction is more efficient for users like me who need to take a lot of screenshots while writing reports.

qoreQyaS:
* added boolean configuration option saveOnShiftDoubleClick

* added condition for doubleclick + shift in mouseDoubleClickEvent handler which creates saveTool instance and emulates click

fixes #2877

regarding #3010 respectively tests/problems:
i tested without a running daemon and with "flameshot gui". my part was working but the normal double click functionality did not work, which has nothing to do with my change but is an open issue #2982 

compiled with libqt5core5a 5.15.6+dfsg-5 and tested on xorg/awesomewm on debian experimental

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -48,6 +48,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initShowSidePanelButton();
     initUseJpgForClipboard();
     initCopyOnDoubleClick();
+    initSaveOnShiftDoubleClick();
     initSaveAfterCopy();
     initCopyPathAfterSave();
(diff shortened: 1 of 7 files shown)


mmahmoudian:
@qoreQyaS Sorry this slipped my eye. Thanks for re-submitting a PR.

PR #3016, opened 2022-12-11
Snap to grid

Joshua-cla:
Little feature who add a grid in the screenshot and snap the tools to this grid.
The grid can be displayed or not and the step of grid can be adjusted. These two options can be edited in the side panel.

French translation updated.

### DIFF
--- a/data/translations/Internationalization_fr.ts
+++ b/data/translations/Internationalization_fr.ts
@@ -4,24 +4,20 @@
 <context>
     <name>AbstractWidgetList</name>
     <message>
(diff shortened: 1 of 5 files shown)


veracioux:
@Joshua-cla Thank you for your contribution. I have reviewed it and it looks fine. I think the functionality is a great addition. Please just fix the formatting of the code by running `clang-format -i $(git ls-files "*.cpp" "*.h")` to fix the failing check.

Helix751:
Adding 'Toggle grid'  it as a persistent option (in the main options dialogue) and also as a keyboard mapping option would be great for easy and quick toggling of this option while editing a clipped region without having to open the side-bar.

PR #3018, opened 2022-12-12
Add grim-based wayland universal screenshot adapter

jack9603301:
In the past, flameshot used a special and superfluous method to call the wayland screenshot component - dbus protocol communication. Although this method is supported by a large number of distributions, it does not take into account the actual situation of wayland (including WM custom desktop environment users). Now, we can enable the wayland universal screenshot adapter with the help of grim, just add the following compilation flags in cmake:

```
-DUSE_WAYLAND_GRIM=true
```

In addition, the patch also adds HYPRLAND type and OTHER type support


mmahmoudian:
This looks good to me. I just request another dev to check it out as well. In the meantime I would be thankful if you can handle the clang format:
https://github.com/flameshot-org/flameshot/actions/runs/3692623906/jobs/6296081652

kyechou:
I tested the patch with the additional cmake option in both Sway and Hyprland.  Almost everything works flawlessly.  There are two things:

1. There's significant delay between the cursor moving and the screen capture window (with `flameshot gui`) moving with the cursor.  I also tested the current master branch (b4300d303d956794a5ae195100e9073a43d4f095), where the issue also exists, and so **I don't think the issue is related to this PR specifically**.  The issue doesn't exist with the current release v12.1.0. I went through some recent issues and didn't find anything similar to what I experienced. @mmahmoudian: Please let me know if I missed an old issue or if I should open a new one. Thanks!
2. Another minor thing is that [src/utils/screengrabber.cpp:147](https://github.com/flameshot-org/flameshot/pull/3018/files#diff-58ac671a0c53721cfd925ead9b485523f50243ea310d0d49caf152a3b52e393fR147) won't ever be executed if I understand it correctly. It doesn't affect the functionality but just in case you care about such things.

yavko:
Wouldn't it a better idea just to use the Wayland protocol? Also grim doesn't work under GNOME, as Mutter doesn't implement the required protocols that grim uses, so I think that should be added to the messages.

yavko:
> > Wouldn't it a better idea just to use the Wayland protocol? Also grim doesn't work under GNOME, as Mutter doesn't implement the required protocols that grim uses, so I think that should be added to the messages.
> 
> The use of grim is mainly because it has implemented the implementation for wlroots. It should be used in a large number of wayland environments. Due to the wayland specification, its implementation is completely determined by the synthesizer. Therefore, for gnome, it is possible to use the dbus protocol to take screenshots is currently the way to go - until there are alternatives available for gnome-compositor

Yeah I mean use the protocol that grim uses, if flameshot detects the user to be on a compositor that implements it. Also wasn't GNOME dbus api made private? Also I meant add it to the cmake messages

src/CMakeLists.txt
```diff
@@ -216,10 +216,17 @@ target_link_libraries(
 )
 
 if (USE_WAYLAND_CLIPBOARD)
+  if(!USE_WAYLAND_GRIM)
+    message(WARNING "You activated the USE_WAYLAND_CLIPBOARD option, but did not activate the USE_WAYLAND_GRIM option, flameshot will use the dbus protocol to support wayland, if you use wayland, it is recommended to enable USE_WAYLAND_GRIM")
```
yavko, line 220: No here it says `if you use wayland`, which isn't exactly true. Is what I mean

yavko: ye

yavko:
> @yavko Now, when users are running on GNOME or KDE, the grim-based generic screenshot adapter will not be enabled

I might be wrong here, but I think kwin (might) implement it, as it implements other wlroots protocols.

yavko:
> If KDE does not support grim, it can be modified in another PR, or through the cmake flag to disable GRIM

After doing some research I think that KDE doesn't support it, but apparently Mir does lol

yavko:
> Ok, do you think I need to remove it? Or keep it for now?

Well from what I've seen KDE implements stuff like layer shell, but I cant find anything on it implementing screencopy so maybe remove it, and someone can re-add it down the line if they so please. For KDE its probably be a better idea to implement the in house solution used by spectacle, in a different PR.

yavko:
> @yavko
> 
> Replenish:
> 
> Regarding the gnome adaptation method of wayland, the currently feasible method is to serve the desktop app of upwork, which is at this address:
> 
> [MarSoft/upwork-wayland](https://github.com/MarSoft/upwork-wayland)
> 
> The author is a freelancer who prepared this for sway, and I implemented a fork that undoes more parts of sway than implements, at (since I'm not sway):
> 
> [jack9603301/upwork-wayland](https://github.com/jack9603301/upwork-wayland)
> 
> This adapter is implemented to run freelance work software such as upwork that relies on the gnome interface under wayland, but it may provide an implementation for gnome software. Although grim may not be available on gnome, we can know that gnome The status of the interface used
> 
> With the help of this simple dbus adapter, some software with the screenshot function of the gnome screenshot protocol can be used freely under wlroots of wayland - because it simulates the interface of gnome

Yeah apps can still use that dbus api, but GNOME removed it on GNOME, in GNOME 41. So if its implemented apps that were made to use it work, but it doesn't work on versions of GNOME newer than 41, 40 and below is fine tho.

mgil2:
Can this be merged as an opt-in with some make flags to enable it? Those of us not using gnome are waiting on this patch as well, have several open issues which would be resolved by it. We should not be punished because of gnome's design decisions when it comes to implementing wayland.

yavko:
> Can this be merged as an opt-in with some make flags to enable it?

That's what this PR adds.
> those of us not using gnome are waiting on this patch as well, have several open issues which would be resolved by it. 

this **ONLY** adds support for wlroots, and is not a universal solution. That's why the portal exists and mostly works fine.
> We should not be punished because of gnome's design decisions when it comes to implementing wayland.

GNOME itself doesn't have working screenshotting it has nothing to do with GNOME's design decisions. GNOME had a internal dbus api that apps leveraged that was made private, which doesn't affect other compositors in anyways. People need to stop blaming everything on GNOME

mmahmoudian:
@jack9603301 Thank you for keeping this branch up to date. I appreciate the efforts. I'm still waiting for other dev to approve the PR.

veracioux (approved):
@jack9603301 @mmahmoudian Sorry for the long wait. I have reviewed the code - looks good. I have compiled flameshot with the `-DUSE_WAYLAND_GRIM` option and it works good on sway. Although I need to mention that without this option flameshot doesn't work for me on wayland. Not that it's in any way this PR's fault, but this means I can't verify if these changes break anything on wayland without grim. So I'll defer to other people's experience with this. I can verify that nothing breaks when running flameshot under X.

mmahmoudian:
@veracioux 

>  I have compiled flameshot with the `-DUSE_WAYLAND_GRIM` option and it works good on sway

Is this something that we should also add to our AUR build instructions and also to our [docs/Sway and wlroots support.md](https://github.com/flameshot-org/flameshot/blob/master/docs/Sway%20and%20wlroots%20support.md) ?

veracioux:
@mmahmoudian To [docs/Sway and wlroots support.md](https://github.com/flameshot-org/flameshot/blob/master/docs/Sway%20and%20wlroots%20support.md) the answer is probably yes. For AUR build instructions, do you mean we should add it as a comment, or enable the option while building based on some condition?

mmahmoudian:
@veracioux 

> To docs/Sway and wlroots support.md the answer is probably yes.

Would it be possible for you to add it? This was we can ask others to test it for us. And also I get a better understanding of where you have used that flag

> For AUR build instructions, do you mean we should add it as a comment, or enable the option while building based on some condition?

Good question. Initially I thought that if it doesn't break X11, we can have it for eveyone and always build with that flag, but then I realized this might be wlroot specific as in KDE Plasma Wayland and Gnome Wayland, Flameshot is already functional. I think this needs some investigation. I have Hyprland on my machine but few others can be tested either by the community or through VM. What do you think?

jack9603301:
I think it's now necessary to sync to other distro's main repos, this includes aur and ebuilds

jack9603301:
> Good question. Initially I thought that if it doesn't break X11, we can have it for eveyone and always build with that flag, but then I realized this might be wlroot specific as in KDE Plasma Wayland and Gnome Wayland, Flameshot is already functional. I think this needs some investigation. I have Hyprland on my machine but few others can be tested either by the community or through VM. What do you think?

Can be used as an option to enable the flag when the user turns on

Issue #3007, opened 2022-12-09
A text marker with high contrast - c.f. Windows snippet tool


FuryBart:
### Feature Description

I would just like to suggest that you take a look at the text marker tool from the Windows "snipping tool", because I think it works different than the marker in flameshot and shows great results (imho). Here is a comparison:
![flameshot_marker](https://user-images.githubusercontent.com/18568915/206711010-8c7e77c0-8925-4951-9bee-dfbfc250df81.png)
1.	Snippet Tool marker
2.	Flameshot Pen	with opacity value 'ff'
3.	Flameshot Pen	with opacity value '80'
4.	Flameshot Pen	with opacity value '30'
5.	Flameshot Marker
Flameshot adds a color layer with opacity over the text. The text is affected and becomes lighter. The Snippet Tool paints the marker color without opacity around the text. So text and marker has more contrast and are very good highlighted.

I have no idea exactly how this works, but I suspect they first analyze the underlying pixel colors before applying the marker as a function of them. Anyway, I would like that kind of text marker in flameshot too.

SamuraiNinjaGuy:
I've confirmed that modifying markertool.cpp gets the results I think you are looking for.

![image](https://user-images.githubusercontent.com/63204645/209303311-6ac1b54f-2789-47bd-89b3-de1fbfe4acc2.png)
6. Modified Flameshot Marker.

Specifically MarkerTool::process and MarkerTool::paintMousePreview contain the line "painter.setOpacity(0.35)" which, if set to 1 instead of 0.35, results in the above changes.

I do wonder why it wasn't just set to 1. I am sure there was a reason for it. Either way, it would be nice if it could be changed. Even if only as an override in the config file.

I've only just looked at flameshot's code though, so I will need to look into where it reads from the ini file before I would submit a pull request.

PR #3033, opened 2022-12-23
Allows users to modify the opacity value of the marker tool.

SamuraiNinjaGuy:
Code was run through the code formatter.

Related to this Enhancement request  #3007

I couldn't figure out why it was changed from 0.5 to 0.35 in 2017 [commit in question](https://github.com/flameshot-org/flameshot/commit/d849e9dbf5f1cb759f1ed2c226e380cc2d24fac9), nor why it was 0.5 originally. It seems to function fine though. I made 0.35 the default, so there should be no differences detected by users.

I am rather rusty with C++ and very unfamiliar with this project, so this should be reviewed.

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -59,6 +59,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initUploadClientSecret();
     initPredefinedColorPaletteLarge();
     initShowSelectionGeometry();
+    initMarkerOpacity();
 
     m_layout->addStretch();
 
@@ -717,6 +718,27 @@ void GeneralConf::initSquareMagnifier()
     });
 }
 
+void GeneralConf::initMarkerOpacity()
+{
+    auto* tobox = new QHBoxLayout();
+
+    int opacity =
+      ConfigHandler().value("markerOpacity").toInt();
+    m_xywhOpacity = new QSpinBox();
+    m_xywhOpacity->setRange(0, 100);
+    m_xywhOpacity->setToolTip(
+      tr("Percentage value of marker opacity, higher is darker"));
+    m_xywhOpacity->setValue(opacity);
+    tobox->addWidget(m_xywhOpacity);
+    tobox->addWidget(new QLabel(tr("Set marker opacity (%)")));
+
+    m_scrollAreaLayout->addLayout(tobox);
+    connect(m_xywhOpacity,
+            SIGNAL(valueChanged(int)),
+            this,
+            SLOT(setMarkOpacity(int)));
+}
+
 void GeneralConf::initShowSelectionGeometry()
 {
     auto* tobox = new QHBoxLayout();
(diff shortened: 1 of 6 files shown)


mmahmoudian:
Thanks for the PR, but in general, it is a good idea to discuss the feature before spending time to implement it.

There is already an issue about discussing this ( #249 ) and a RFC that has very good points:

https://github.com/flameshot-org/flameshot/blob/01c8f095e28c0b8d6d018e703c2557f507dadc20/docs/RFC/0000-Add-Opacity-slider.md

Your PR is a good start‌, but it is too config-centric without are GUI elements to modify that value on the fly (i.e qhen user is in the middle of annotating a screenshot). Are you interested to modify your PR to adopt some of the features discussed there? For instance adding a slider for this value in the side-bar?

SamuraiNinjaGuy:
I'd definitely be interested in making the changes suggested.

I admit, this is only the second project I have attempted to contribute to, and the issues were not yet linked. I will start looking into linking issues before I jump into writing code. It was a "make the change you want to see" moment.

I can see why someone would want an opacity setting for the other tools, and to be able to make changes on the fly (noticed that last night when experimenting with other colors). I also like the layout you suggested.

However, I see a potential issue with workflow when the opacity setting is applied globally. I noticed, with my changes, that opacity set to 1.0 on any default color except yellow (and maybe light green) makes black text on a white background essentially unreadable. By the same token, you may want a red selection box around an object to be opaque, while highlighting some text in red, but realize you need a couple opaque arrows. If opacity is global, you can accomplish this by changing the opacity setting multiple times within the GUI side-bar, but I went years without discovering the GUI side-bar and would like to maintain the same level of simplicity so the GUI side-bar isn't required to use any tool. I also don't want to make changes to the default UIX.

I think I would want to be able to modify default opacity settings by tool and/or allow the option to tie it to a global setting. I'd also like to allow the user to be able to change that without starting over on their screenshot, which means it shouldn't be in the config area. Maybe add a "use global opacity setting" checkbox under the actual opacity slider? Grayed out unless a tool is selected to keep it consistent? Currently the only tool specific options show up for text, and they appear on top, but it seems clunky to but a checkbox so far away from the slider. Default would be checked for all but marker? I worry it would negatively alter the cleanness of the GUI side-bar.

Sorry this is formatted poorly, I wanted to reply right away, and I don't have much time for the next couple days.

SamuraiNinjaGuy:
@mmahmoudian @veracioux First off, I think we should move this discussion to ([ #249 ](https://github.com/flameshot-org/flameshot/issues/249)).

I like the idea of a keyboard shortcut. The main concern I have is I don't want to modify the default UIX. I literally used flameshot for years before I knew of any customization beyond the I could do within flameshot. OOB it worked for essentially all of my needs without any changes. (I only learned of some of them once I started researching the code/UI for the pull request).

The reason why I feel that opacity needs to be able to be modified by tool is because they already don't all share the same opacity settings (specifically the marker). If I don't make a special case for the marker, I will have to change the default UIX, or cap the marker at 35% max opacity (being able to increase this is actually part of the main change of this original pull request). As long as I need to make a case for the marker, I may as well make them all customizable.

I'd rather let the user specify if they want the tools to share a global opacity or not. I think the sidepanelwidget is the right place because they can change it on the fly and if it is in the config, it would be confusing why the settings were inconsistent.

I will this post in ([ #249 ](https://github.com/flameshot-org/flameshot/issues/249)) with these ideas. As this warrants a new pull request, I will also close this one.

Issue #2997, opened 2022-11-29
'flameshot screen' not working on some screens


Massimo-B:
### Flameshot Version

Flameshot v12.1.0 ()

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Gentoo

### Description

I have 2 keyboard shortcuts defined on Xfce on all my machines.

<Print>:  flameshot gui
<CTRL+Print>:  flameshot screen -p /home/massimo/screenshots

Using the send with "flameshot screen" on my 2 screen machine, this only works when focussing the left screen. Focussing the right screen results in a black screenshot.

I have seen other issues on my 3 screen machine, there only one of the screens works, while the others even result in distorted screenshots with a lot of artifacts on the picture.

### System Information

Compiled with Qt 5.15.5
linux: 5.15.75-gentoo
gentoo: 2.9

mmahmoudian:
Please provide all the requested information in the issue reporting form. For instance you didn't provide the monitor configuration (which one is the primary? left or right? do you have fractional scaling? what are the orientations? what are the relative positions, and etc. (the output of `xrandr` helps clarifying most of these)).

Your main issue looks like a duplicate of #2815 and the other issue (the one with distortion seems like a duplicate of #2910

Massimo-B:
2 identical monitor models, both connected via onboard DisplayPort of a desktop machine with i7-4790, integrated Intel graphics only. Usual landscape orientation.
Right monitor is the primary.
flameshot only works on the left Monitor which is not primary.
I'm using native resolution on both.
```
$ xrandr
Screen 0: minimum 320 x 200, current 5120 x 1440, maximum 16384 x 16384
VGA-1 connected (normal left inverted right x axis y axis)
   1024x768      60.00  
   800x600       60.32    56.25  
   848x480       60.00  
   640x480       59.94  
DP-1 connected 2560x1440+0+0 (normal left inverted right x axis y axis) 553mm x 311mm
   2560x1440     59.95*+
   2048x1080     60.00    24.00  
   1920x1080     60.00    50.00    59.94  
   1920x1080i    60.00    50.00    59.94  
   1600x1200     60.00  
   1280x1024     75.02    60.02  
   1152x864      75.00  
   1280x720      60.00    50.00    59.94  
   1024x768      75.03    60.00  
   800x600       75.00    60.32  
   720x576       50.00  
   720x480       60.00    59.94  
[...]

PR #3035, opened 2022-12-25
Fixed issue with black screen when use 'screen' argument

danillucky1234:
When trying to take a screenshot of the whole screen, we try to take pixels from the wrong window, which results in a black screen.
Hotfix to this issue: #2997

### DIFF
--- a/src/core/capturerequest.cpp
+++ b/src/core/capturerequest.cpp
@@ -3,10 +3,7 @@
 
 #include "capturerequest.h"
 #include "confighandler.h"
-#include "imgupload/imguploadermanager.h"
-#include "pinwidget.h"
 #include "src/config/cacheutils.h"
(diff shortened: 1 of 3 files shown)


mmahmoudian:
Thanks for the PR. Regarding 284f3e524c1eee5296e3ce07689aeef255ac2481 and changes to `capturerequest.cpp`, can you clarify why you have remove those includes.

The rest of the PR looks clean and ready.

danillucky1234:
Thank you for the quick response! I removed those headers because none of their contents are used in `capturerequest.cpp` - neither methods from `screenshotsaver.(h|cpp)`, nor classes from `imguploadermanager.h`, `pinwidget.h`

veracioux (approved):

PR #3059, opened 2023-01-15
size indicator - fix disabling via config and performance improvements

lbatalha:
### PR Info

This PR fixes the geometry indicator still causing severe performance issues when disabled.
It also refactors the code to get rid of unnecessary logic and improves performance significantly.

I managed to also get a massive performance boost by using `update()` instead of `repaint()`, now its just barely usable at my total display area on wayland (11kx4k pixels), though not actually usable for me specifically. People with smaller display areas might now be able to use this feature.

I removed some code that seemed to never get executed, namely the guarded `#if(0)` in the `paintEvent` and the `m_sizeIndButton` stuff in `updateSizeIndicator()`, it is not clear to me what that code was related to but I could not find any way to trigger it, or any straight answers from the previous PR. There was also a stray `updateSizeIndicator()` call in `initSelection()` which was not necessary and potentially doubled the performance overhead.

While doing this, I uncovered a bug in the timer that hides the geometry indicator, even after the timer runs out and the indicator is removed, the timer keeps executing (as per Qt docs). This causes the `xywhTick()` function to execute which causes a `repaint()` (this was changed to `update()` in this PR). This caused a very noticeable mouse stutter every time the timer triggered. So, with a 1000ms timer, as you move the mouse around freely, every 1 second you would get a stutter.
Changing to `update()` improved it, but it is still a very annoying bug, so I fixed it by changing the timer to a `singleShot` type.

In the `paintEvent`, part of the refactor included the display Grid code, as it seems when it was implemented the QPainter::save and restore combined flow was not used, and per the comment on line 534, all code that needs to save and restore should do so via this flow.
The display Grid code was not updated to use `update()` instead of `repaint()`, because I saw no performance change between them (the repaint is only triggered when toggling the grid on/off or changing the grid size, and with both functions I still got a tiny stutter on doing so. Not really important imo)

----------
### Performance investigation

In the original PR (#2766) it was noted that there was quite the performance penalty, though not much investigation was done, presumably because the users testing it had a small enough display area to where it was still usable and/or their software/hardware setup was not affected as much (X11 vs Wayland)
Interestingly it was noted that even if the config disables the code in the `paintEvent` function, the performance penalty is mostly still there.

I investigated this and shockingly, commenting out **all** the contents of that function still causes this problem.
The offending line is the `repaint()` line in `showxywh()`, regardless of the `paintEvent` having any code to execute.
I do not understand how this is possible. There is nothing shown on screen besides a black/grey background, not even the screenshot or a cursor or anything.
This definitely feels like the way to go to keep investigating this issue.

**Notably,** the magnifier feature, which uses `update()`, suffers from about the same level of stuttering when enabled.

I did some CPU tracing using `pprof`, `callgrind`, and others, but it just seemed like normal Qt functions being executed, repeated repaints, qtwayland, alpha and geometry calls, etc..

During selection at least one CPU core is maxed out and there is a lot of GPU usage (~70% on my GTX1070) (I am using `sway` as my compositor)

----

Side note: I ran the clang formater and tons of files popped up, and weird stuff like line 28 in `confighandler.h` popped up, is this working as intended?


mmahmoudian:
Thank you for keeping this up to date

ardishco-the-great:
I have a feeling that This PR isn't going to get merged any time soon...

ardishco-the-great:
I have tested this PR, compiled with CMake, and it has fixed #3111, It doesn't take a long time to select the region on Sway now on my system.

ardishco-the-great:
Okay, I have sat down and given this PR a longer review in terms of function and while the region selection is fixed, copying to clipboard does not work, I get the "Copied to clipboard" message but not . I don't know If this issue is a Wayland or Xorg problem, but it is a problem nonetheless.

lbatalha:
Seems to work the same as the current master for me, select an area, click the copy to clipboard button, paste somewhere. (tested on wayland/sway with `QT_QPA_PLATFORM=wayland`)

ardishco-the-great:
> 

Setting the environment variable myself, nothing seems to have changed for me.
If you know If there are any problematic variables in my launch script, let me know 
```
#!/bin/bash

# Wayland Fix
export _JAVA_AWT_WM_NONREPARENTING=1
export QT_QPA_PLATFORM=wayland
export XDG_CURRENT_DESKTOP=sway
export XDG_SESSION_DESKTOP=sway
export WLR_NO_HARDWARE_CURSORS=1
export XCURSOR_THEME=Posy_Cursor_Black
export XCURSOR_SIZE=24
# onagre
export RUST_BACKTRACE=1
# Theming

export QT_QPA_PLATFORMTHEME=qt6ct;qt5ct
#export QT_QPA_PLATFORMTHEME=qt6ct
export GTK_THEME=Kanagawa-Border
#export QT_STYLE_OVERRIDE=qt6ct;qt5ct

# mozilla wayland
export MOZ_ENABLE_WAYLAND=1
# sway itself
exec sway
```
I don't think anything I've set here would cause problems, but do try it on your machine aswell..

ElonVampire:
收到！

lbatalha:
> > 
> 
> Setting the environment variable myself, nothing seems to have changed for me. If you know If there are any problematic variables in my launch script, let me know

> I don't think anything I've set here would cause problems, but do try it on your machine aswell..

According to the wiki, you should also import specific variables into your systemd user session (`systemctl --user show-environment`)

Regardless, have you tested with the current git master? This PR doesn't seem to change any code related to the image itself.

ardishco-the-great:
> 
> According to the wiki, you should also import specific variables into your systemd user session (`systemctl --user show-environment`)
> 
> Regardless, have you tested with the current git master? This PR doesn't seem to change any code related to the image itself.

After you mentioned it, I compiled it from master and the same issue is present.
However, when I use the `flameshot-git` package from the AUR, the selection delay is present, yes, but the copying is not.
the `flameshot` package refuses to do anything, so that's completely broken for me.

I have a few variable that I use, which are all in my Sway config, yeah. 
The output of that command you gave me is:
```
HOME=/home/vaporsnake
LANG=en_US.UTF-8
LOGNAME=vaporsnake
MAIL=/var/spool/mail/vaporsnake
PATH=/usr/local/bin:/usr/bin
SHELL=/bin/bash
SYSTEMD_EXEC_PID=1417
USER=vaporsnake
XDG_DATA_DIRS=/home/vaporsnake/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local>
XDG_RUNTIME_DIR=/run/user/1000
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
DISPLAY=:0
SWAYSOCK=/run/user/1000/sway-ipc.1000.1452.sock
WAYLAND_DISPLAY=wayland-1
XDG_CURRENT_DESKTOP=sway

```

I got the commands from the wiki, I'm pretty sure these are correct. Here are the commands I run on sway startup
```
exec --no-startup-id systemctl --user import-environment DISPLAY WAYLAND_DISPLAY SWAYSOCK
exec --no-startup-id dbus-update-activation-environment --systemd DISPLAY WAYLAND_DISPLAY SWAYSOCK
```
I will open an issue regarding this, thanks.

ardishco-the-great:
After compiling flameshot from master, flameshot-git also has the same issue, nevermind. I guess it's a me-issue, will include it in the new issue.

ardishco-the-great:
I have been playing around with this PR and the flameshot-git package from the AUR recently and it looks like that still has delay issues, but the copying issue was fixed, however, after deleting the files of this PR and recompiling several times, I still seem to have the same copying problem, specifically with this PR now. If I can provide logs of some kind to help, I would be happy to do so but as of now I'm not sure how to help better this PR.

zjeffer:
I just compiled this PR for the first time, and for me the delay issue is fixed but the copying issue is not.

Here's the output (compiled with USE_WAYLAND_GRIM=true`)

```
./src/flameshot gui      
flameshot: warning: grim's screenshot component is implemented based on wlroots, it may not be used in GNOME or similar desktop environments
Unable to get current screen, starting to use primary screen. It may be a cause of logical error and working with a wrong screen.
QLayout: Attempting to add QLayout "" to SidePanelWidget "", which already has a layout
Unable to get current screen, starting to use primary screen. It may be a cause of logical error and working with a wrong screen.
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
Unable to get current screen, starting to use primary screen. It may be a cause of logical error and working with a wrong screen.
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
```

Saving works fine.

lbatalha:
This PR is not related to copying, that's likely some other separate issue as there is no code related to that here.

ardishco-the-great:
> This PR is not related to copying, that's likely some other separate issue as there is no code related to that here.

You are correct, this code has nothing to do with copying, however, this behaviour is observed 100% of the time in your PR and on mainline it's more like 60% (at least in my testing) and wanted to mention it in case anyone happened to know anything about it

mmahmoudian:
@januszmk 
> this fixes the issue for me and doesn't cause new issues

Thanks for letting us know. It is a valuable information

> any idea if and when this will be merged into master?

The code review process needs to be done. All I can say is "soon". As a general rule, I don't want to make a promise that I cannot keep, therefore I will avoid giving direct answer to "when" :)

Gigas002:
Can confirm this PR working for me too, been using my fork with this and one other PR merged for a month without issues

emrakyz:
> > This PR is not related to copying, that's likely some other separate issue as there is no code related to that here.
> 
> You are correct, this code has nothing to do with copying, however, this behaviour is observed 100% of the time in your PR and on mainline it's more like 60% (at least in my testing) and wanted to mention it in case anyone happened to know anything about it

@ardishco-the-great  Can you try it with two cmake configs. You also need to make sure you have wl-clipboard:
USE_WAYLAND_CLIPBOARD
USE_WAYLAND_GRIM

Clone the repo, apply the patch, enable the above configuration and compile it. Hope it helps.

zeyadtamimi:
Hey, also reporting that this PR seems to fix #3274

veracioux (approved):

Issue #1344, opened 2021-02-17
Easy OCR


leoneivaw:
I would like to suggest OCR feature. 
It is very useful for a lot of people.
It demands no more than 8 lines code and one dependence.
I use Flameshot with OCR and is great. It as native feature is a lot better!

ocr.sh file

trap "rm ~/Imagens/ocr*" EXIT
flameshot gui -p ~/Imagens 
while [ ! -f ~/Imagens/ocr.png  ]; do sleep 1; done
tesseract -l por ~/Imagens/ocr.png ~/Imagens/ocr &> /dev/null 
while [ ! -f ~/Imagens/ocr.txt ]; do sleep 1; done
cat ~/Imagens/ocr.txt | xsel -bi
sleep 2
exit

borgmanJeremy:
Thanks for posting this, but it is a duplicate of #702 and we actually had someone submit a PR to add an OCR feature. It requires some clean up so it wont make it into 0.9 but its on the way.

mmahmoudian:
@leoneivaw I just faced a PDF that needed OCR and reading your message got me to think if there is a better way (not to write to disk and read and while loop and etc., and I found it:

```sh
flameshot gui --raw | tesseract -l fin stdin stdout | xclip -selection clipboard
```

I have used the Finnish language here, but of course that should be changes according to your needs.

The following are the software versions I have:

> tesseract 4.1.1
> Flameshot v0.10.0
> xclip version 0.13

leoneivaw:
@mmahmoudian 

Works great!
Simple and fast.

Thanks

stefanocudini:
it would be very useful a button OCR which transforms the image in the selection into text.
That would be really cool

mmahmoudian:
@stefanocudini you should wait until the plugins are implemented. For the time being you can use the script I suggested above and bind it to a keybinding

ashuvssut:
@mmahmoudian  your command works fine for large enough texts
`flameshot gui --raw | tesseract -l fin stdin stdout | xclip -selection clipboard`

BUT, for small fonts, it doesnt work so well...

is there some way use imagemagick to magnify and then OCR it?
I tried something like the command below after reffering [this StackOverflow ANs](https://stackoverflow.com/a/4945131/12872199)
`flameshot gui --raw | convert -resize 400% stdin stdout | tesseract -l eng stdin stdout | xclip -selection clipboard`
(The command writted above is wrong)

can you please help me in this

[...]

Issue #720, opened 2020-04-14
Choose custom colors of color wheel


TCastelein:
I would like to customize the colors in the color wheel activated when right mouse clicking. I see in the code that there is a setting called `userColors` however i fail to add it correctly to the `ini` file so i don't have it working.

Can anybody help me with an example of how to edit the `userColors` or the feature request of setting the colors through the CLI just like main and contrast color.

socketbox:
Yes, this would be nice. The code seems to indicate that a list of strings will be read in from the INI file, but none of the following work.
```
userColors={"black","white",...
userColors=@Variant(\0\0\0\x43\xff\xff\xffx\x86\x00\x00\x00\0\0)     
userColors={"#ffffff", "#000000
```
 `syslog` shows `qt5ct: custom style sheet is disabled`, apparently after the application fails to parse the file.

nst0022:
The syntax is:

`userColors=#800000, #ff0000, #ffff00, #00ff00, #008000, #00ffff, #0000ff, #ff00ff, #800080`

= default colors = darkRed, red, yellow, green,    darkGreen, cyan, blue, magenta, darkMagenta

socketbox:
@nst0022 I'm sorry, I must be dense, but putting that exact line in my `~/.config/Dharkael/flameshot.ini` file prevents the app from opening.

That is, if I add `userColors` as you've typed...
```
[General]                                                                                                                                                      
buttons=@Variant(\0\0\0\x7f\0\0\0\vQList<int>\0\0\0\0\xf\0\0\0\x1\0\0\0\x2\0\0\0\x3\0\0\0\x4\0\0\0\x5\0\0\0\x6\0\0\0\xf\0\0\0\a\0\0\0\b\0\0\0\t\0\0\0\n\0\0\0\v\0\0\0\f\0\0\0\r\0\0\0\xe)                                                                                                                                     
contastUiColor=@Variant(\0\0\0\x43\x2\xff\xff\x39\b\xeb\xf9\xe9\xce\0\0)                                                                                       
disabledTrayIcon=false                                                                                                                                         
drawColor=@Variant(\0\0\0\x43\x1\xff\xff\xff\xff\0\0\0\0\0\0)                                                                                                  
drawThickness=0                                                                                                                                                
filenamePattern=flameshot-%F-%T                                                                                                                                
savePath=/home/socketbox/Desktop/Screenshots                                                                                                                         
showHelp=true                                                                                                                                                  
[...]

Issue #702, opened 2020-03-19
OCR to clipboard hook for selections


kaihendry:
Sometimes I'm working with screenshots and would be awesome to make a selection around some text and send it off to say https://aws.amazon.com/rekognition/ for it analyse and return the text to set my clipboard.

MyriaCore:
OCR would be great honestly!

nsa:
There is a powerful OCR engine from Google called https://github.com/tesseract-ocr/tesseract and it's open-sourced. It would be super cool if the maintainers could integrate tesseract to flameshot as the author suggested, by doing so there would no need for external API connections such as Amazon's, because tesseract works offline.

JScriber:
There's a project called [dpscreenocr](https://github.com/danpla/dpscreenocr) that does the job, yet the selection is not as convenient and user-friendly as flameshot's.
OCR integrated to flameshot with a dpscreenocr inspired configuration would be really great.

enoy19:
I implemented that feature and created a pull request: #1239

kaihendry:
https://news.ycombinator.com/item?id=27242392 discusses the topic in a MacOS context

Pheggas:
Anything new around this feature request? It would be super cool to me tho. Oftentimes i'm working with text copying and this 
feature would be absolute killer!

xeaon:
@Pheggas If you are comfortable with compiling yourself, check my comment above.

I'm using this feature since then and it works great.

borgmanJeremy:
I am okay to merge this feature if we can:
- Fix the merge request to it passes CI
- Explain how to package / distribute the language packs on all platforms.

Pheggas:
> @Pheggas If you are comfortable with compiling yourself, check my comment above.
> 
> I'm using this feature since then and it works great.

I'd like to. But it would be my first compilation. I'll try to look at it. In case of failure, I'll write you 🙂

leoneivaw:
I am anxious waiting for this feature!
Thanks!

irfan798:
For anybody waiting for the functionality I found this blog post:
https://slint.github.io/blog/ocr-screenshot.html

If you install _tesseract_, it is basically a one liner
```bash
flameshot gui --raw | tesseract stdin stdout | xclip -in -selection clipboard
```

ek1ng:
It's an amazing solution! Thanks!

ek1ng:
[...]

PR #3074, opened 2023-01-27
Add OCR acction with tessercat

rsrdesarrollo:
This PR adds an OCR action using tesseract.

It needs tesseract to be in path or configure full path

image

image

### DIFF
--- a/data/graphics.qrc
+++ b/data/graphics.qrc
@@ -96,8 +96,10 @@
         <file>img/material/white/move_up.svg</file>
         <file>img/material/white/delete.svg</file>
         <file>img/material/black/apps.svg</file>
+        <file>img/material/black/ocr.svg</file>
         <file>img/material/black/image.svg</file>
         <file>img/material/white/apps.svg</file>
+        <file>img/material/white/ocr.svg</file>
         <file>img/material/white/image.svg</file>
     </qresource>
 </RCC>
--- a/data/img/material/black/ocr.svg
+++ b/data/img/material/black/ocr.svg
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg id="svg284" version="1.1" viewBox="0 0 1e3 1e3" xmlns="http://www.w3.org/2000/svg">
+ <path id="path282" d="m696.4 10h-563.9v980h735v-809zm97.2 183.8h-109.9v-109.9zm12.7 735h-612.6v-857.6h428.8v183.8h183.8z"/>
+ <rect id="rect596" x="288.67" y="358.05" width="395.13" height="75.212"/>
+ <rect id="rect596-8" x="288.67" y="493.82" width="395.13" height="75.212"/>
+ <rect id="rect596-1" x="288.67" y="629.59" width="395.13" height="75.212"/>
+ <rect id="rect596-5" x="288.67" y="765.36" width="395.13" height="75.212"/>
+</svg>
--- a/data/img/material/white/ocr.svg
+++ b/data/img/material/white/ocr.svg
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg id="svg284" version="1.1" viewBox="0 0 1e3 1e3" xmlns="http://www.w3.org/2000/svg">
+ <path id="path282" d="m696.4 10h-563.9v980h735v-809zm97.2 183.8h-109.9v-109.9zm12.7 735h-612.6v-857.6h428.8v183.8h183.8z" fill="#fff"/>
+ <rect id="rect596" x="288.67" y="358.05" width="395.13" height="75.212" fill="#fff"/>
+ <rect id="rect596-8" x="288.67" y="493.82" width="395.13" height="75.212" fill="#fff"/>
+ <rect id="rect596-1" x="288.67" y="629.59" width="395.13" height="75.212" fill="#fff"/>
+ <rect id="rect596-5" x="288.67" y="765.36" width="395.13" height="75.212" fill="#fff"/>
+</svg>
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -56,6 +56,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initAntialiasingPinZoom();
     initUploadHistoryMax();
     initUndoLimit();
+    initTesseractPath();
     initUploadClientSecret();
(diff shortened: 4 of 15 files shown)


mmahmoudian:
Thanks for the PR, but please always discuss the new features before spending time implementing it. Also it would be faster for you if you search among the open and closed issues to see if a topic is discussed.

In my opinion this PR is unacceptable for few reasons:
1. OCR is out of the scope of a screenshot tool, but it is already possible to do it as explained in #1344
2. We will only allow such features via our plugin system which is explained in details in #2529

I suggest we keep your PR (at least partially) for when the plugin system is implemented, as some of the parts (e.g icons) can be used in the PR.

I would also like to know the opinion of other devs (e.g @veracioux @hosiet @ZetaoYang ) as the above does not necessarily reflect the stance of the project as a team, but rather just my opinion.

rsrdesarrollo:
To be completely honest, the implementing time was nearly none, I needed the feature, and I'm currently using it in a more effective and usable way than piping commands. 

Also, this project, despite being an amazing tool, lacks of any OCR capability and, for the past nearly 3 years, since the first request to implement this came (#720) AFAIK no one has done really any code to implement the feature in a usable way for any kind of user. AFAIK plugins system is only a concept, nothing implemented.

> OCR is out of the scope of a screenshot tool,

Let me, greenshot, sharex and snagit disagree, OCR is indeed a great and wanted feature for screenshot tools.

In my humble opinion, here is a PR to allow 90% of your users to do what they need today. You can easily think about moving it to plugins when the plugin system is implemented.

That being said, I have no need for this PR to be merged, I can live with my fork, just wanted to give something back to the community, I guess anyone who needs it can take it from my copy of the project.

Regards!

debuglevel:
Sad it's not merged. I'm always starting https://github.com/ShareX/ShareX to do a quick OCR. It just works.

erfanium:
> OCR is out of the scope of a screenshot tool

I think this argument can also be applied to the currently implemented post-processing (draw shape, pen, pixelate) capabilities. ‌I use flameshot not because it's the only software that takes photos of my screen. I use it because it's feature rich and I really miss the OCR feature.

mmahmoudian:
@erfanium 
> I think this argument can also be applied to the currently implemented post-processing (draw shape, pen, pixelate, ...) 

Flameshot is a screenshot annotation tool. All those features you mentioned are necessary for annotating screenshots. OCR has nothing to do with annotation.

> I use it because it's feature rich and I really miss the OCR feature.

Then please go and support the plugin system with which this feature will be available.

This was a unanimous decision by all devs to only allow such features as plugin. One of the reason for this is extremely high maintenance burden of such features on all supported operating systems.

-------

@debuglevel 
> I'm always starting https://github.com/ShareX/ShareX to do a quick OCR. It just works.

They only support Windows. We support Linux, Mac, Windows and even BSD.

-------

@rsrdesarrollo 
> Let me, greenshot, sharex and snagit disagree, OCR is indeed a great and wanted feature for screenshot tools.

For now, the policy of maintainers is to keep it screenshot and annotation specific. Any other gizmo (watermark, printing, GIF, OCR, barcode reader,...) should only and only be allowed through a plugin system. This is to help us keep this project maintainable and prevent it from bloat.

But that said, Flameshot is a FLOSS tool (emphasis on "free" and "libre"), therefore, anyone is legally allowed to maintain their own fork based on the restrictions imposed by the license of Flameshot.

Also they are all Windows-only tools and are not cross-platform, or commercial in case od Snagit.

> In my humble opinion, here is a PR to allow 90% of your users to do what they need today

You are right, it will allow some some users do OCR today, but maintaining it will be an everlasting burden on devs. You kindly provided a PR but you are not taking the responsibility to take care of this part of the code for ever. Do you?

I also don't think 90% of Flameshot users are in need of OCR. We don't have telemetry, so we don't have statistics on our user count, but I bet the number you should aim for is less than 1%.

> AFAIK plugins system is only a concept, nothing implemented.

True, it is not implemented yet (at least not yet landed on the main branch). But:
1. If there is enough demand, someone (our dev or someone from community) would start working on it 
2. We are in negotiation with a company. They need certain features that we would like to provide only as plugin. They agreed to commission this. If things goes well and if we find a dev who would like to do this big task, then we will have the plugin system.

I thank you again for your PR. But I think you can also agree that in general:
1. Policy is policy, especially when it is democratically decided upon
2. Attacking/pressuring devs and volunteers is not constructive (emphasis on "in general")

debuglevel:
> @debuglevel
> 
> > I'm always starting https://github.com/ShareX/ShareX to do a quick OCR. It just works.
> 
> They only support Windows. We support Linux, Mac, Windows and even BSD.

Yes...? So does tesseract.

GrabbenD:
I agree with @mmahmoudian

To keep bloat to minimum, ensure low maintenance burden and provide high cross-platform compatibility, it just doesn't make sense to have advanced tools like OCR as part of the core project. Furthermore, alternative projects such as ShareX are clogged with unnecessary tools.. I hope Flameshot never gets to that state.

**Plugin system** is a very reasonable solution. I've come to realize even a highly maintained project like Tesseract are prone to providing bad results and [needs tuning](https://github.com/flameshot-org/flameshot/issues/702#issuecomment-2068046288). Having plug-and-play plugins would be easier to tweak rather than having to re-compile the entire project to try platform specific tweaks.

This really sums up the reason why I enjoyed this PR in the first place:

> It checks for Tesseract in PATH meaning **the feature is opt-in** and uses upstream updates rather than a statically linked library (which would inevitably become outdated)

mmahmoudian:
@GrabbenD 
> This really sums up the reason why I enjoyed this PR in the first place

The only reason I've kept this PR open is due it's simplicity and straightforward implementation. I am hoping that we can somehow incorporate this PR as an add-on so that @rsrdesarrollo can maintain this add-on instead a a full fork. At the end of the day, I'm one person in a team and I should not and cannot act alone.

-------

@debuglevel This is a fully volunteer-based FLOSS project. We are not getting paid to serve you, nor we owe you anything! This is our project and we maintain and manage it as we see fit. Lastly, be civil; we have zero tolerance policy for trolls.

debuglevel:
Yeah, therefore just close the issue so everybody knows this is a "won't fix". Keeping it open does not help anybody but just attracts more people who hope this will get implemented.

GrabbenD:
@debuglevel

> I am hoping that we can somehow incorporate this PR as an add-on

PR #3097, opened 2023-02-12
Add developer docs

veracioux:
Adds developer docs, including documentation generated from code comments using doxygen.

The documentation itself is not complete, but please do read it if you find the time and make suggestions what can be improved or added.

@mmahmoudian Especially read the 'Maintaining the documentation' section so you know how the documentation is deployed and maintained.

The docs are live here: https://flameshot.org/docs/dev


mmahmoudian:
I finally had some time to go through it. I just have one comment. Considering that the `DOC.md` is in the `/docs/`, doesn't it pollute the `/docs/` folder? My point is that doesn't it be more clean and clear to move all these files under a dedicated folder (e.g `/docs/dev-docs/` or something like that) to have all the files related to this developer documentations contained within the same folder?

veracioux:
@mmahmoudian Ah yes. I used this file as a temporary file where I would write documentation brainstorms. I accidentally committed it - and I will remove it now. (There are still some things from that file that need to be added to the proper documentation, but I'll do that another time).

PR #3103, opened 2023-02-17
Update and rename Internationalization_sv_SE.ts to Internationalizati…

eson57:
…on_sv.ts

Ok, I fixed  and completed Swedish translation. I also renamed .ts file because sv_SE didn't work. Flameshot did read only sv.


mmahmoudian:
Thanks, but we use Weblate for translation.  Merging this way for a language that we don't speak is not easy. But since you changed the file name, let's see.

Thanks for your contribution and spending time in making Flameshot better 👍

eson57:
Ok, I can do Weblate but I don't seem to be able to upload my finished file. Project seem to be locked. [I've been translating](https://hosted.weblate.org/user/eson/) many projects on Weblate since 2015.

PR #3145, opened 2023-04-02
Add generic wlroots DesktopInfo for compositors such as labwc

joshuataylor:
Compositors such as [labwc](https://github.com/labwc/labwc) use "wlroots" as their `XDG_CURRENT_DESKTOP`.

I like that they do that, and hope others do it as well, as it means Flameshot doesn't need to add an ∞ amount of case statements.

### DIFF
--- a/docs/Sway and wlroots support.md
+++ b/docs/Sway and wlroots support.md
@@ -15,6 +15,8 @@ export XDG_SESSION_DESKTOP=sway
 exec sway
 ```
 
+Other compositors, such as labwc, use `wlroots` as the XDG_CURRENT_DESKTOP value.
+
 You will also need to ensure that systemd/dbus is aware of these environment variables; this should be done **in your sway config** so that the DISPLAY and WAYLAND_DISPLAY variables are defined.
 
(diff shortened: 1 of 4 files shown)


mmahmoudian:
isn't Sway a wlroot? I'm also not sure about hyprland but I think that is also wlroot. If these are correct, then perhaps having them in `src/utils/desktopinfo.h` and `src/utils/screengrabber.cpp` is redundant.

xvkdev:
> isn't Sway a wlroot? I'm also not sure about hyprland but I think that is also wlroot. If these are correct, then perhaps having them in `src/utils/desktopinfo.h` and `src/utils/screengrabber.cpp` is redundant.

sway self reports as sway
```
✦2 ❯ echo $XDG_CURRENT_DESKTOP
sway
```

RichAyotte:
Hyprland reports itself at `Hyprland`

```sh
✦ ❯  echo $XDG_CURRENT_DESKTOP
Hyprland
```

borgmanJeremy:
Is there utility in this given sway reports as "sway" and hyrland reports as "hyprland"?

borgmanJeremy:
This is not needed anymore.

Issue #3164, opened 2023-04-15
screen shot got half resolution on Plasma Wayland


microcai:
### Flameshot Version

v0.10.1

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Gentoo

### Description

on plasma wayland session, flameshot can only toke screen shot at half the resolution.

### Screenshots or screen recordings

![Screenshot_20230415_182705](https://user-images.githubusercontent.com/172208/232208431-a26775b7-28cc-4d48-825e-47375e7778fb.png)

### System Information

KDE plasma 5.27 on wayland.

SebTM:
Same issue on SwayWM/Wayland

mmahmoudian:
This should have been fixed with #4498 
If you can, try compiling from the master branch and let us know is this issue still persists. I'm going to close this for now, but if the issue persists, we will re-open it.

PR #3165, opened 2023-04-15
fix kde wayland screen shot half resolution problem

microcai:
this will address issue #3164 

using org.kde.KWin.ScreenShot2 interface can solve the problem.

see screencast below

Screenshot_20230415_184419


mmahmoudian:
People in #3164 have said that #3166 have solved their problem. This PR is much more massive compared to #3166, so we should be careful about merging.

PR #3215, opened 2023-06-02
Add gnome-based wayland universal screenshot adapter

jack9603301:
By default, flameshot uses the program's default dbus protocol interface to communicate with gnome's screenshot protocol, but this is now broken on some gnome installations, so we'll use gnome-screenshot as gnome's wayland screenshot adapter and then communicate directly with the gnome screenshot component!

Designed to address: https://github.com/flameshot-org/flameshot/issues/3213

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -216,11 +216,18 @@ target_link_libraries(
 )
 
 if (USE_WAYLAND_CLIPBOARD)
+  if(!USE_WAYLAND_GNOME)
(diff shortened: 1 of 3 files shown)


jack9603301:
@mmahmoudian This solution is pending review, and it seems to work fine in GNOME

xarpen:
Great work, this finally makes flameshot work under arch/gnome/wayland!
Would be nice if this can be merged into mainstream.

borgmanJeremy:
I think this is not needed based on testing on #4055 . Screenshotting on gnome generally works, and this didnt resolve an issue with snapcraft.

PR #3266, opened 2023-07-10
feat: Add configuration setting to allow users to choose catbox.moe instead of Imgur

rlperez:
Add configuration setting to allow users to choose catbox.moe instead of Imgur

This is a PR to add a configuration setting to allow users to choose [catbox.moe](https://catbox.moe/) instead of Imgur. This is useful for users who want to use a different image host, or who are in a country where Imgur is blocked. This was done following the efforts on the imgur uploader code. This is the first time I have written C++ since I was working on my undergrad over a decade ago so any feedback on the code would be welcomed.

Configuration dialog changes

uploaded image: https://files.catbox.moe/c39iz0.png

capture of upload and viewing the image remotely


mmahmoudian:
@rlperez Thanks for the PR. I wish you were discussing this with the dev team before putting the efforts into implementing this. As we have mentioned in [the contribution guideline](https://github.com/flameshot-org/flameshot/blob/master/docs/CONTRIBUTING.md) "For larger changes submit an [RFC](https://github.com/flameshot-org/flameshot/blob/master/docs/RFC.md)" .

The reason is that we have decided long ago that we only allow uploads to other platforms via our plugin system (#2529). You can see the list of all such issues here:
https://github.com/flameshot-org/flameshot/issues?q=is%3Aissue+is%3Aopen+label%3Aplugin

I will change this to draft as you already invested some time and hopefully we can at some point port it into a plugin.

rlperez:
@mmahmoudian I wasn't sure if this was big enough for an RFC but now I have my answer. Beforehand I did see mentions of plugins but had not seen anything that was an implementation to work around yet. This is my favorite screen cap software so I don't mind helping out with some things. 

What would be the best next steps?
* Write an RFC for this service?
* Rewrite to the plugin spec?
* Help finish whatever work is there for the plugin and plugin manager?

mmahmoudian:
@rlperez we definitely can use some extra hand on the deck. Perhaps it would be easier to coordinate with other devs if you joing our matrix space and the Developers channel:

Join my space Flameshot https://matrix.to/#/#flameshot-org:matrix.org

The current high priority are:
1. Updating Flatpak dependencies to the latest
2. Improving Wayland support (labeled by "Wayland")
3. Address fractional scaling issue
4. Imgur bugs: Investigate why sometimes the uploaded image does not end up in the list of uploaded images in Flameshot)
5. Windows CLI support

PR #3276, opened 2023-07-24
Wayland daemon fix & hyprland multi window implementation

FelixJochems:
This PR is based on PR #3059 because it's unusable for me to debug otherwise.

This PR implements a Hyprland specific screenshot function, although some of this could easily be extended to other wayland DE's. The currently active screen is obtained from ```hyprctl activewindow``` which is the only hyprland specific part right now.

I have some questions before furthering development on this:

- I've Added some of the USE_WAYLAND_GRIM conditions, but there's also a waylandDetected() function. Does flameshot aim to dynamically detect wayland and run based on that or should we use the preprocessor to eliminate as much of these as possible. For example: the activateWindow() isn't implemented in wayland and we could solve this with either solutions.

- The clipboard, currently we keep the daemon running as long as necessary and watch it until flameshot no longer owns it to then end the program. Wayland also has the ```wl-copy,  wl-paste & wl-clipboard-persist``` programs. So the question is, do we want to daemon to offer the clipboard information or do we want to transfer these responsibilities to programs such as those mentioned before?

All in all, this PR is meant to bring flameshot to a more usable level for hyprland. I know there's currently a bug where the daemon doesn't always end when the clipboard data changes, but I want to know what direction flameshot heads in on the subject of clipboard information before digging deeper into that bug. Let me know if you need any extra information.

P.S. For hyprland users: ```windowrule = float, ^(flameshot)$``` will make the pinned screenshot immediately float ;)

### DIFF
--- a/src/core/flameshotdaemon.cpp
+++ b/src/core/flameshotdaemon.cpp
@@ -108,11 +108,15 @@ void FlameshotDaemon::start()
 
 void FlameshotDaemon::createPin(const QPixmap& capture, QRect geometry)
 {
+    #ifdef USE_WAYLAND_GRIM
+    start();
+    #endif
     if (instance()) {
         instance()->attachPin(capture, geometry);
         return;
     }
-
+    
+    #ifndef USE_WAYLAND_GRIM
     QByteArray data;
     QDataStream stream(&data, QIODevice::WriteOnly);
     stream << capture;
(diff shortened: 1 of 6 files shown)


alba4k:
only takes a screenshot of the first monitor from the left (usually the one at 0x0)

FelixJochems:
Looks like I made a mistake where it only copies the current dimensions of the current active screen instead of all screens. Will fix this soon

FelixJochems:
This can be fixed by always forcing the left most display and a Hyprland rule to always set the application to 0,0. While this is a fix, I'm no longer pursuing this because it's superjanky and I prefer a wayland focused fork (which requires a lot more effort) with native wayland elements that also avoids DBUS.

Closing

PR #3331, opened 2023-09-10
applauncherwidget: enter key to launch current/first shown app

d-a-s:
Currently with the open with program dialog, you have to click the application with the mouse. This can be inconvenient, especially when using keyboard shortcuts such as `ctrl+O`, and/or typing a filter.

This code allows for pressing the enter key to open.

If there is a "Currently selected" item, (e.g. one uses `shift+tab` to focus the control, and arrow keys to select), it will open that selected item.

If there is no current selection, it will use the first visible item.


mmahmoudian:
sound like a good addition to me, and the code looks clean and good to me too. let's wait for approval of @veracioux but in the meanwhile I will run the Actions

veracioux (approved):

PR #3334, opened 2023-09-14
added qt5-image-formats-plugins as optional dependency

HardenedSteel:
Added qt5-image-formats-plugins as optional dependency to README.md

It's required for to able save in WebP format.

I only added for Debian, missing packages needs to be added for other distros.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,7 @@ Also you can open and build/debug the project in a C++ IDE. For example, in Qt C
 - Git
 - OpenSSL
 - CA Certificates
+- QT Image Formats
 
 #### Debian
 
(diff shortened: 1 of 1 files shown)


jobukkit:
On Arch the package is called `qt5-imageformats`. This indeed needs to be documented, I was confused where the WebP option was.

ElTh0r0:
This PR is deprecated and could be closed, because the code was fully ported to Qt 6 in the meantime.
Updating the README (including 'Qt Image Formats') for Qt 6 is in progress: #4029

mmahmoudian:
@HardenedSteel thanks for the PR and sorry for getting to this PR very late. Hopefully we will have your support and contribution in the future too :)

PR #3354, opened 2023-10-01
Fix missed escape sequences at /src/CmakeLists.txt

MakarenkoPavel:
Hello

I found out that when the project repository is cloned to a path containing spaces, it leads to a Cmake configuration error.

The problem occurs only on MacOS


mmahmoudian (approved):

mmahmoudian:
Thanks for the PR. Looks good to me. Let's wait for another Dev's approval

MakarenkoPavel:
Hi @mmahmoudian 

I also noticed that [build instruction](https://github.com/flameshot-org/flameshot#build) is a little bit outdated and may seem to be confusing in some way.

Since root [CMakeLists](https://github.com/flameshot-org/flameshot/blob/9557cd68ae0cd113b2e617a6f5e1af7288aeec05/CMakeLists.txt#L56) has implicitly specified Qt5_DIR variable on its own we don't need to provide this variable as command line parameter after `brew install qt5`

veracioux (approved):

PR #3364, opened 2023-10-05
Fix full/screen capture use last region

### LINKED ISSUE #3086: flameshot full captures only part of the screen
### Flameshot Version

Flameshot v12.1.0 ()
Compiled with Qt 5.15.3

### Installation Type

Linux Mint 21.1 Cinnamon

### Description

For example, this is a screenshot of the whole screen

And this is a screenshot after running `flameshot full`

I use the "Use last region" option, but I don't think it should work for `flameshot full`

I don't use fractional scaling or Wayland and have only one display.

### Steps to reproduce

1. check "Use last region" in flameshot configuraton
2. run `flameshot gui`
3. select a region on the primary monitor
4. save to file
5. run `flameshot full`
6. you should get the same screenshot as in the previous step

_No response_

### System Information

1. Linux Mint 21.1 Cinnamon

2. 

3.
System:
  Host: mercury Kernel: 5.15.0-58-generic x86_64 bits: 64
    Desktop: Cinnamon 5.6.7 Distro: Linux Mint 21.1 Vera
Graphics:
  Device-1: NVIDIA GK104 [GeForce GTX 760] driver: nvidia v: 470.161.03
  Display: x11 server: X.Org v: 1.21.1.3 driver: X: loaded: nvidia
    gpu: nvidia resolution: 1920x1080~60Hz
  OpenGL: renderer: NVIDIA GeForce GTX 760/PCIe/SSE2
    v: 4.6.0 NVIDIA 470.161.03

Kit-p:
Ignore "Use last region" option when running `flameshot full` or `flameshot screen`.

Closes #3086.

### DIFF
--- a/data/translations/Internationalization_bg.ts
+++ b/data/translations/Internationalization_bg.ts
@@ -1054,13 +1054,13 @@ Please solve them manually in the configuration file.</translation>
     </message>
     <message>
         <location filename="../../src/config/generalconf.cpp" line="247"/>
-        <source>Use last region</source>
-        <translation>Use last region</translation>
+        <source>Use last region for GUI mode</source>
+        <translation type="unfinished">Use last region for GUI mode</translation>
     </message>
     <message>
         <location filename="../../src/config/generalconf.cpp" line="248"/>
-        <source>Uses the last region as the default selection for the next screenshot</source>
-        <translation>Uses the last region as the default selection for the next screenshot</translation>
+        <source>Use the last region as the default selection for the next screenshot in GUI mode</source>
+        <translation type="unfinished">Use the last region as the default selection for the next screenshot in GUI mode</translation>
     </message>
     <message>
         <location filename="../../src/config/generalconf.cpp" line="262"/>
--- a/data/translations/Internationalization_ca.ts
+++ b/data/translations/Internationalization_ca.ts
@@ -1062,13 +1062,13 @@ Please solve them manually in the configuration file.</translation>
     </message>
     <message>
         <location filename="../../src/config/generalconf.cpp" line="247"/>
-        <source>Use last region</source>
-        <translation>Use last region</translation>
+        <source>Use last region for GUI mode</source>
+        <translation type="unfinished">Use last region for GUI mode</translation>
     </message>
     <message>
         <location filename="../../src/config/generalconf.cpp" line="248"/>
-        <source>Uses the last region as the default selection for the next screenshot</source>
-        <translation>Uses the last region as the default selection for the next screenshot</translation>
+        <source>Use the last region as the default selection for the next screenshot in GUI mode</source>
+        <translation type="unfinished">Use the last region as the default selection for the next screenshot in GUI mode</translation>
     </message>
     <message>
         <location filename="../../src/config/generalconf.cpp" line="262"/>
--- a/data/translations/Internationalization_cs.ts
+++ b/data/translations/Internationalization_cs.ts
@@ -1121,13 +1121,13 @@ Vyřešte je, prosím, ručně v souboru s nastavením.</translation>
     </message>
     <message>
         <location filename="../../src/config/generalconf.cpp" line="247"/>
-        <source>Use last region</source>
-        <translation>Use last region</translation>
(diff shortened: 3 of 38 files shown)


mmahmoudian:
Thanks for the PR. I believe instead of "exclusion" criteria we should only limit this to `flameshot gui`. This is because the `flameshot screen` should also ignore this as far as I can understand the use-case.

Kit-p:
Thanks for the feedback. That makes a lot of sense.
In this case, it feels like we should also clarify it in the wordings.
Something like "Use last region with GUI capture"?

mmahmoudian (approved):

mmahmoudian:
> In this case, it feels like we should also clarify it in the wordings.
> Something like "Use last region for GUI capture"?

Yes, you right, the current wording is sub-optimal. Would you like to also include those changes in this PR (considering that it kinda fits the context)? What needs to be changed is the config window text and it's tooltip.

Kit-p:
I am not sure how the translation system works, will changing the keys invalidate all current translations?

I can include it in this PR, but the translations should be a separate issue I suppose.

Kit-p:
Okay, after some digging I believe here is what I need to do.
Please correct me if I misunderstand anything.

1. Update the translation keys (English) in code.
2. Replace the translation keys in all translation files so current translations are kept.
3. Notify translators to help update the translations by setting type="unfinished". (?)

mmahmoudian:
As far as I understand, you don't need to care about translations. All you need to do is to change the code for English keys. Then weblate will pull changes from the master branch and automatically roll it as a string to be translated in all the supported languages.

This is also my understanding. I'm relatively new to handling translation conflicts (already a messy one has happened and I'm in the process of merging the translations manually between the master and the Weblate fork). For now, I have locked the Weblate from the admin dashboard, so it is safe change anything on our side.

Kit-p:
Just tried changing English in code and compile with the `GENERATETS` option.
It set the current translation to have `type=vanished` and added a new string to be translated.
I am just wondering if it would be better to keep the current ones, because the core meaning of the wordings hasn't deviated.

Btw, thanks a lot for the translation merging work, I definitely also noticed this messy situation.

veracioux (approved):

Kit-p:
Looks like this will be merged in the current state.
So attached is a patch for updating the translation files, feel free to ignore if it isn't helpful.
[fix-flameshot-use-last-region-translation.patch](https://github.com/flameshot-org/flameshot/files/12849185/fix-flameshot-use-last-region-translation.patch)

veracioux:
@Kit-p I've reviewed the patch file and it looks good. I think you can apply the changes in a commit here, and we'll merge them all together. But let's wait for @mmahmoudian to confirm.

@mmahmoudian You mentioned a messy merge, are you talking about the Internationalization_ta.ts file between current weblate/master and master? If yes, it looks like that file effectively has no changes, but the whole file shows up as changed in the diff viewer. This is probably due to a change in windows/linux line endings. There are also some other conflicts, but not many and not especially tricky, unless I'm looking at the wrong thing.

mmahmoudian:
> You mentioned a messy merge, are you talking about the Internationalization_ta.ts file between current weblate/master and master

I think the file in question was `*_fi.ts` that was causing issues. This is what Weblate is reporting itself:

> Rebasing (1/44)
> Auto-merging data/translations/Internationalization_fi.ts
> CONFLICT (content): Merge conflict in data/translations/Internationalization_fi.ts
> error: could not apply 5aa5a5c... Translated using Weblate (Finnish)
> hint: Resolve all conflicts manually, mark them as resolved with
> hint: "git add/rm <conflicted_files>", then run "git rebase --continue".
> hint: You can instead skip this commit: run "git rebase --skip".
> hint: To abort and get back to the state before "git rebase", run "git rebase --abort".
> Could not apply 5aa5a5c... Translated using Weblate (Finnish)

If you have the setup ready for merging please go ahead and merge. I haven't had time this week to work on it. If not, I will try to merge tomorrow.

Kit-p:
@veracioux Ready for review, thanks!

veracioux:
@Kit-p Merged. Thank you for your contribution!

Issue #3323, opened 2023-09-03
Blur is not sufficient


madduck:
### Flameshot Version

Flameshot v12.1.0 (Debian 12.1.0-2)
Compiled with Qt 5.15.8

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Debian unstable

### Description

When using the blur tool to hide e.g. email addresses or other sensitive information, one has to drag across the area twice, or else the original information is not sufficiently obscured.

Case in point, below is a screenshot of the version info above, where I tried to hide the Debian version in use (just for demonstration).

Please increase the blur amount.

### Steps to reproduce

1. Run GUI
2. Select portion of screen
3. Use blur tool to obscure some text
4. Observe that the text can still be discerned.

### Screenshots or screen recordings

![image](https://github.com/flameshot-org/flameshot/assets/195073/55730006-a422-4335-be56-8f8a7af2332e)

### System Information

Not that I think this is relevant, but:
Debian unstable, Xorg, dual-monitor using XRandR, AwesomeWM

mmahmoudian:
This is duplicate of multiple other issues. Usually searching among open and closed issues can save your time. 😊


Like any other tool in Flameshot you can"increase the tool size" by scrolling or kust simply typing the number you want. Regarding bluring security issues and how to mitigate it, I have answered this is great details here:

https://github.com/flameshot-org/flameshot/issues/3242#issuecomment-1609122285

Issue #3242, opened 2023-06-21
Increase the strength of the blur tool


tetku:
### Flameshot Version

Flameshot v12.1.0 (96c2c82e)
Compiled with Qt 5.15.2
winnt: 10.0.19045
windows: 10

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Windows 10 22H2

### Description

Flameshot's blur tool doesn't seem to work stably compared to similar tools. It's too weak. If you apply the blur tool to the text once, the text will still be readable. To completely hide the text, you must apply the blur tool on the text a minimum of three times. This looks like a bug.

Please see the results of other editing tools below for comparison. The built-in blur tool of other editing tools hides all the text the first time, making it unreadable. To solve the problem, the intensity of the existing blur tool needs to be increased.

Flameshot Windows
![flameshot-](https://github.com/flameshot-org/flameshot/assets/135356201/6f6e9342-7e0d-45f3-91d1-362decf17962)

[Annotely web editor](https://annotely.com/editor)
![annotely_image](https://github.com/flameshot-org/flameshot/assets/135356201/72817514-dc3b-452e-a7b1-1dd554b907b3)

[Snipaste Pro](https://apps.microsoft.com/store/detail/snipaste/9P1WXPKB68KX)
![Snipaste_2023-06-21_03-27](https://github.com/flameshot-org/flameshot/assets/135356201/52689da0-22a6-4a84-87d4-a87b3ef2abea)

[ShareX](https://getsharex.com/)
![sharex](https://github.com/flameshot-org/flameshot/assets/135356201/eebc5380-ac36-460e-9f90-6c347fcc4d3f)

### Steps to reproduce

Use the blur tool on any screenshot. (not pixelate tool)

### System Information

Windows 10 22H2 19045.3086

2560x1440

Issue #2439, opened 2022-02-17
Pixelation as blur can be reversed, time to use another blur / redaction algorithm?


R8s6:
### Feature Description

I came across [a blog](https://bishopfox.com/blog/unredacter-tool-never-pixelation) claiming pixelation can now be "hacked" or reversed to reveal the original english text and numbers.

![](https://github.com/BishopFox/unredacter/raw/main/img/wow_such_secrets.gif)

The code is available on github too:
https://github.com/bishopfox/unredacter

Nothing too urgent, but perhaps the current blur algorithms can be changed to something else?

borgmanJeremy:
I saw that too. You can simply draw a black box if you need more security. You can also increase the kernel size in flameshot to make more letters to into a single block.

R8s6:
black blocks will do, thanks!

mmahmoudian:
@R8s6 if you read their explanation you'll get to know that they are very much relying on the information they get from adjacent pixels' contrast and formation. Therefore one easy way is ro increase the pixel size of the pixelisation tool (scroll) to have the pixels larger than your line height. Also this way when you use normal fonts, one pixel will cover more than one character and therefore much harder, if not impossible, to guess the text.

@borgmanJeremy there is an algorithmically easy fix, although I don't know how the pixelation is implemented. If the pixels are shuffled around randomly, then this can make the guessing also very hard as massive noise is induced

PR #3368, opened 2023-10-09
added randomize pixels function to pixelate tool

### LINKED ISSUE #3289: Option to obfuscate mosaics to avoid retrieving redacted text from mosaics
### Feature Description

Considering that [Unredacter](https://github.com/bishopfox/unredacter) can currently retrieve redacted text from mosaics, I propose a feature which obfuscates mosaics to allow using the mosaic tool without Unredacter being able to retrieve it.

How to obfuscate the mosaic is an implementation detail, but I have a couple of suggestions for what to do:

- Change the colors of the mosaic pixels to a random nearby color for each color channel within a percentage range, like 20% of the total range of colors.
  - For example, if one of the mosaic pixels is RGB (144, 144, 144), it could be randomly changed to RGB (153, 134, 146).
    - Applying this randomization to each large mosaic pixel would throw off Unredacter's guessing.
- Split the mosaic into sections and randomly mirror, flip, and shuffle the sections.

kurayami07734:
## Concerns
Pixelate tool's look changes completely. It now looks like white noise

## Changes 
- Added a randomize pixels function to pixelate tool
- Utilized `random_shuffle` from `<algorithm>` and `memcpy` from `<cstring>`

### before
lorem before

### after
lorem after

### DIFF
--- a/src/tools/pixelate/pixelatetool.cpp
+++ b/src/tools/pixelate/pixelatetool.cpp
@@ -8,6 +8,8 @@
 #include <QGraphicsScene>
 #include <QImage>
 #include <QPainter>
+#include <algorithm>
+#include <cstring>
 
 PixelateTool::PixelateTool(QObject* parent)
   : AbstractTwoPointTool(parent)
@@ -78,10 +80,21 @@ void PixelateTool::process(QPainter& painter, const QPixmap& pixmap)
         QPixmap t = pixmap.copy(selectionScaled);
         t = t.scaled(size, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
         t = t.scaled(selection.width(), selection.height());
-        painter.drawImage(selection, t.toImage());
+        QImage image = t.toImage();
+        image = randomizePixels(image);
+        painter.drawImage(selection, image);
     }
 }
 
+QImage PixelateTool::randomizePixels(const QImage& image)
+{
+    QVector<QRgb> pixels(image.height() * image.width());
+    std::memcpy(pixels.data(), image.bits(), image.byteCount());
+    std::random_shuffle(pixels.begin(), pixels.end());
+    QImage shuffledImage(image.width(), image.height(), image.format());
+    std::memcpy(shuffledImage.bits(), pixels.data(), image.byteCount());
+    return shuffledImage;
+}
 void PixelateTool::drawSearchArea(QPainter& painter, const QPixmap& pixmap)
 {
     Q_UNUSED(pixmap)
--- a/src/tools/pixelate/pixelatetool.h
+++ b/src/tools/pixelate/pixelatetool.h
@@ -24,6 +24,7 @@ public:
 
 protected:
     CaptureTool::Type type() const override;
+    QImage randomizePixels(const QImage& image);
 
 public slots:
     void pressed(CaptureContext& context) override;


mmahmoudian:
@kurayami07734 Thanks for the great work. Is it possible to:
1. change the pixel size in accordance to the tool size that the user is using?
2. considering that its a drastic change, can you add an item in the settings that allow the user to toggle this on/off? This is good for privacy, but it now looks like an extra opaque layer.

kurayami07734:
> change the pixel size in accordance to the tool size that the user is using?

I did not find where exactly the pixel size is being set

> considering that its a drastic change, can you add an item in the settings that allow the user to toggle this on/off? This is good for privacy, but it now looks like an extra opaque layer. This may cause backlash from users.

I tried to add a setting for this in my latest commit. I am some how getting a segfault now

kurayami07734:
Hey guys, you probably missed this PR

veracioux:
@kurayami07734 I'm reviewing your PR now. Sorry about the wait.

src/config/generalconf.cpp
```diff
@@ -264,6 +270,17 @@ void GeneralConf::initSaveLastRegion()
             this,
             &GeneralConf::saveLastRegion);
 }
+void GeneralConf::initRandomizePixelation()
```
veracioux, line 273: @kurayami07734 You never call this method.

veracioux (changes requested):

src/utils/confighandler.cpp
```diff
@@ -98,6 +98,7 @@ static QMap<class QString, QSharedPointer<ValueHandler>>
     OPTION("saveAfterCopy"               ,Bool               ( false         )),
     OPTION("savePath"                    ,ExistingDir        (                   )),
     OPTION("savePathFixed"               ,Bool               ( false         )),
+    OPTION("randomizePixelation"         ,Bool               ( false         )),
```
veracioux, line 101: @kurayami07734 Please add a corresponding entry to `flameshot.example.ini`.

src/config/generalconf.cpp
berkentekin: all init methods are called once in the constructor method at the beginning of this file

nxmndr:
Is it not the same issue as #2439, #3323, #3242 etc ?

The current blur tool can be set at max size by scrolling and I think the effect is very satisfying. I'd rather have it than these little dots (no offense). It may not be clear enough that the scroll can be used for size though.

kurayami07734:
> Is it not the same issue as #2439, #3323, #3242 etc ?
> 
> The current blur tool can be set at max size by scrolling and I think the effect is very satisfying. I'd rather have it than these little dots (no offense). It may not be clear enough that the scroll can be used for size though.

I've randomized the pixel using `std::random_shuffle`, so it looks like white noise a bit.

AlexFolland:
What's the status of reviewing this?  The only conversation in the latest review looks resolved.  Is this ready for merging?

jrpie:
The pRNG is not initialized using a suitably secure seed (time can be brute forced, or - in the case of images - simply taken from the metadata). Hence using this implementation does not improve security over pixelation (it might even be worse).

More generally I'd argue that the contents of the area to be hidden should not be used as an input *at all*, see my [comment](https://github.com/flameshot-org/flameshot/issues/3289#issuecomment-2429210951) on #3289.

AlexFolland:
> The pRNG is not initialized using a suitably secure seed (time can be brute forced, or - in the case of images - simply taken from the metadata). Hence using this implementation does not improve security over pixelation (it might even be worse).

That's a good point.  Using something like time since local boot instead of time since epoch in the seed assignment (currently `unsigned seed = std::chrono::system_clock::now().time_since_epoch().count();`) would be a useful change.

jrpie:
Using uptime is not a good idea either. The uptime of most machines is less that a month. And a month has only about `2.6 * 10^6` seconds which can be easily brute-forced. Using time to generate secure random numbers is not going to work. What would be needed is a [CPRNG](https://en.wikipedia.org/wiki/Cryptographically_secure_pseudorandom_number_generator).

A better solution would be to sort the pixels by value and then shuffle using an insecure PRNG for the same visual effect. That way reversing the insecure shuffling would only reveal the sorted pixels.

However as I said I'm not convinced that the approach of shuffling works at all. All values of all pixels are kept, which is quite a lot of information. There might be situations where e.g. the precise number of oddly colored pixels from [subpixel rendering](https://en.wikipedia.org/wiki/Subpixel_rendering) might reveal substantial information about the secret. The point is that I have absolutely no idea how to prove that no such attack exists. I'm also quite pessimistic about other similar approaches. There are simply way too many footguns.
However when not using the input at all, proving security is trivial.

borgmanJeremy:
I appreciate your contribution on this, but it's directly conflicting with #3765 and we are going to merge that one. Thanks again for the work.

PR #3313, opened 2023-08-21
Make loading of languages consistent across invocation types

veracioux:
Currently, languages are loaded only when invoking flameshot as a daemon, so localization doesn't work for the other invocation types. This fixes that.

PR #3370, opened 2023-10-09
Pass argc as reference to reinitializeAsQApplication (fixes #3367)

### LINKED ISSUE #3367: [Debian testing crashed] exec 'flameshot gui' in debian testing distro, crashed by Segmentation Fault
### Flameshot Version

Flameshot 12.1.0 and master branch

### Installation Type

Debian testing

### Description

Installed the Flameshot 12.1.0 in testing version, when I exec 'flameshot gui', output:
**_QLayout: Attempting to add QLayout "" to SidePanelWidget "", which already has a layout
Segmentation fault_**

then I pull the master branch and compile in locally and meet the same issues.
_m75t@debian:~/Downloads$ flameshot gui
QLayout: Attempting to add QLayout "" to SidePanelWidget "", which already has a layout
Segmentation fault
m75t@debian:~/Downloads$_

### Steps to reproduce

1. install or compile from source code, v12.1.0 and master branch
2. run 'flameshot gui'
3. crashed -> QLayout: Attempting to add QLayout "" to SidePanelWidget "", which already has a layout
Segmentation fault

flameshot-gui

### System Information

m75t@debian:~/Downloads$ uname -a
Linux debian 6.4.0-3-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.4.11-1 (2023-08-17) x86_64 GNU/Linux

Monitor:

monitor

marceldev89:
Fixes crash caused by https://github.com/flameshot-org/flameshot/pull/3313

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -117,7 +117,7 @@ void configureApp(bool gui)
 
 // TODO find a way so we don't have to do this
 /// Recreate the application as a QApplication
-void reinitializeAsQApplication(int &argc, char* argv[])
+void reinitializeAsQApplication(int& argc, char* argv[])
 {
     delete QCoreApplication::instance();
     new QApplication(argc, argv);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -117,7 +117,7 @@ void configureApp(bool gui)
 
 // TODO find a way so we don't have to do this
 /// Recreate the application as a QApplication
-void reinitializeAsQApplication(int argc, char* argv[])
+void reinitializeAsQApplication(int &argc, char* argv[])
 {
     delete QCoreApplication::instance();
     new QApplication(argc, argv);


src/main.cpp
```diff
@@ -117,7 +117,7 @@ void configureApp(bool gui)
 
 // TODO find a way so we don't have to do this
 /// Recreate the application as a QApplication
-void reinitializeAsQApplication(int argc, char* argv[])
+void reinitializeAsQApplication(int &argc, char* argv[])
```
veracioux, line 120: @marceldev89 Nice catch. Though it looks to me that `argv` could fail in the same way. Can you replace it with `char *(&argv)[]`?

marceldev89: @veracioux it shouldn't. The problem was that the `QApplication` constructor signature which is `QApplication(int &argc, char **argv)`. So it expects a reference to `argc` instead of the value. `argv` should be fine as `char* argv[]` is basically the same as `char **argv` I think :)

veracioux: @marceldev89 Yup, very subtle and annoying. Please do a [clang-format](https://github.com/flameshot-org/flameshot/blob/master/docs/CONTRIBUTING.md#code) on the code, so our pipeline can pass before merge.

marceldev89: Done :)

veracioux (approved):

src/main.cpp
veracioux: Thank you! Merging.

PR #3379, opened 2023-10-13
fix appimage build

borgmanJeremy:
Tested this on my fork and its building. Let's see if I messed up the patch when I back ported it :)


mmahmoudian (approved):

mmahmoudian:
I'm so puzzled. Why the `Linux-pack.yml` github action does not run for this PR. I don't see any reason why this is happening.

PR #3408, opened 2023-11-12
Add screenshot capture method for D-Bus

Troplo:
I've added a new D-Bus method allowing the screen to be captured via `dbus-send --session --dest=org.flameshot.Flameshot --type=method_call / org.flameshot.Flameshot.captureScreen string:1` for example.

Having a way to invoke a screen capture via an existing process improves performance drastically when running Flameshot via a shortcut for example, as the official recommended method of executing `flameshot gui` every time you want to take a screenshot spawns a new instance each time whereas this method allows it to be executed with the instance of Flameshot running in the system tray.

Please let me know if this is an improper use of D-Bus of course, I just believe it's a nice addition to improve performance.

### DIFF
--- a/data/dbus/org.flameshot.Flameshot.xml
+++ b/data/dbus/org.flameshot.Flameshot.xml
@@ -36,5 +36,15 @@
       <arg name="notification" type="s" direction="in"/>
     </method>
 
+    <!--
+       captureScreen:
+       @captureMode: Desired mode of capture.
+
+       Allow the user to capture a screenshot over DBus.
+    -->
+    <method name="captureScreen">
+      <arg name="captureMode" type="q" direction="in"/>
+    </method>
+
   </interface>
 </node>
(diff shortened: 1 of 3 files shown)


veracioux:
@Troplo Thank you for the PR. We usually insist on creating an issue and discussing potential features. We used to have a few dbus methods for this purpose (see https://github.com/flameshot-org/flameshot/blob/6d29e133d3182b9162c7dee15f12afc510cf8c4b/data/dbus/org.flameshot.Flameshot.xml), because previously `flameshot gui` and the other commands actually did rely on the daemon to do the work. But when I made the commands standalone, I removed the DBus methods as everything could be performed via CLI.

Can you describe the performance issues you are having, maybe open an issue as well?

Troplo:
Made an issue #3419 describing the performance issues.

ElTh0r0:
I think this PR can be closed in regard of merged #4605

PR #3415, opened 2023-11-20
Change Ubuntu 20.04 label for better accuracy with the OS interface

rbatinov:
Change the Keyboard label for Ubuntu 20.04 as it is in the OS for better understanding. 
image


veracioux:
@rbatinov Thanks for your contribution. Merging.

rbatinov:
You are welcome!

Issue #3416, opened 2023-11-21
Find surrounding post, e.g., in a forum; line edge detection


libertyteeth:
### Feature Description

I see several other issues that describe selecting an entire window:
#5 #334 #1816 #1814 
My use case is to select the entire post, in a forum.  I am perhaps halfway through coding the solution.
My idea of usage:
1. Bind (in my OS) a keystroke to "flameshot gui --post".
2. Before hitting that keystroke, put the mouse over a post, preferably over the background.
3. When invoked with that option, it will start from the mouse cursor and go out 1 pixel at a time in all four directions, checking whether "that edge" is a line of the same color and different from the initial starting point.  Bonus points for "different from the predominant color" inside the box, which would be effectively auto-detecting the background color.
4. If it finds a line, stop in that direction.  If on the next iteration the ends of the line are no longer the same color as the line, start moving again.
5. If all four edges are stopped, we've found the post; return that region.
6. If we reach the edge of the screen in any direction, we can stop and return a null region.
One enhancement to the above I can see while typing this is, some forums may have edges with rounded corners, so could add an option and a "percentage that's corner" or direct pixel count, e.g. "flameshot gui --post --corner=5%" or "--corner=50" for pixels.  Then in step 4 above, it wouldn't start moving again until the ends of the line exceeded e.g. 50 pixels, for the "--corner=50" example.
An example from linuxquestions.org:
![Linuxquestions org example of Flameshot auto-selecting an area based on edge detection - Screenshot from 2023-11-20 21-42-53](https://github.com/flameshot-org/flameshot/assets/32841567/62187261-3c7c-4180-b7c5-8f70a91713e2)
Note that this would only select a portion of the post, as shown in the image.  Which leads to another enhancement: "flameshot gui --post --border=2" so it would have to find two pixels worth of edges on all four sides in order to stop and return that region.  Maybe make this the default, once implemented.
Here's that, updated:
![Linuxquestions org example of Flameshot selecting the entire post with --border=2 - Screenshot from 2023-11-20 21-54-00](https://github.com/flameshot-org/flameshot/assets/32841567/e71f208a-7ee1-4878-840f-956bd213d83d)

libertyteeth:
I have implemented basic functionality: detect a single-line box around the mouse cursor.

This addresses Pull Request #3418 (I am new at this, please guide me if I’m making mistakes).

Invoke with "flameshot gui --box", while the mouse cursor is inside a box, preferably over a background color.

[...]

PR #3418, opened 2023-11-25
Implement basic functionality: detect a single-line box around mouse cursor.

libertyteeth:
Implement basic functionality: detect a single-line box around mouse cursor.

This addresses Issue #3416 (I am new at this, please guide me if I’m making mistakes).

Invoke with "flameshot gui --box", while the mouse cursor is inside a box, preferably over a background color.

I changed it to "box" instead of "post" as that's more general.

This is my first "large" contribution so please let me know if there's anything I missed?  Thanks!  (I had wanted to comment on the above issue, but it appears if I comment it will close it, which wasn’t what I wanted.  So, making the Pull Request first, instead.)

ISSUES:

1. Note that I have not translated any strings that I added into other languages.

2. I have not updated documentation.

3. I have not tested on multiple monitors.

FUTURE:

1. Want to add the two features I mentioned in Issue #3416: allow for rounded corners by ignoring a percentage, or a certain number of pixels, of the ends of the lines being detected; and, an option to keep detecting until a border of a certain size has been found.

2. Want to add the "background color detection" mentioned in Issue #3416: basically, count the pixels of each different color inside the box being checked, and the highest count is the background.  (For efficiency: save the result and then only add the "box/lines just inside the four edges" to the calculation, similar to how the line-detection code only checks for the first and last pixels of an already-detected line as it grows in place.)

3. Want to add another icon/button, so the user can click that, then click anywhere on the screen, and have it perform the box detection and auto-select a region.  In other words, freeze the screen first, then select.  One use case is: hover over a link which causes a popup to appear, and the user wants to capture that popup.  Trying to move the mouse inside the box to be captured, though, moves the mouse off the link, and the popup goes away.  So to capture those, will need this additional feature.

GENERAL ISSUES:

1. I'd like to merge the strings in the code with the strings in flameshot.fish; there's duplication there, and the risk is that someone may update one, and not the other.  Can they pull from a common source?  (Probably I should create an "Issue" about this?)

### DIFF
--- a/data/shell-completion/flameshot.fish
+++ b/data/shell-completion/flameshot.fish
@@ -83,6 +83,7 @@ __flameshot_complete gui -l "print-geometry"    -s "g"  -f   -d "Print geometry 
 __flameshot_complete gui -l "upload"            -s "u"  -f   -d "Upload the screenshot"
 __flameshot_complete gui -l "pin"                       -f   -d "Pin the screenshot to the screen"
 __flameshot_complete gui -l "accept-on-select"  -s "s"  -f   -d "Accept capture as soon as a selection is made"
+__flameshot_complete gui -l "post"              -s "q"  -f   -d "Capture the post the mouse is hovering over"
 
 # SCREEN subcommand
 __flameshot_complete screen                             -f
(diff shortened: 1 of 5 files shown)


mmahmoudian:
@libertyteeth thanks for the PR. We are all learning everyday 🤓 thanks for being a responsible person and transparently discussing the issues you have found. When your changes are ready, you can create a PR as a draft and get some feedback and have some constructive discussion 👍👍

Issue #3277, opened 2023-07-24
Would it be possible to add a config setting that removes the glowing border around a pinned screen capture?


rickwillcox:
### Feature Description

I love this program, I use it every day. My only issue with it is that is makes hacking together a quick mock up kinda of awkward because there is the glowing border around each capture. When I put 5-6 together then I have a bit of a mess of borders everywhere. Id love to have an option to turn that border off when if possible. Cheers

PR #3427, opened 2023-12-02
Add new config option to hide drop shadow around pinned images

schwubdiwub:
Added a new general config option "pinHideDropShadow" which hides the drop shadow around the PinWidget.

Based on this issue #3277


borgmanJeremy:
@mmahmoudian  whats your thought on this change?

mmahmoudian:
@schwubdiwub thank you for showing interest in Flameshot and contributing to it. Also, sorry that it took us so long to get to this PR.

My only comment is that in the configuration text and the hint and etc. We should be consistent with the wording (i.e., 'border' -> 'shadow')

@borgmanJeremy i have heard this IRL from friends that they want the glow/shadow to be toggleable. Initially i was somewhat against it if we don't add at least some border, but now that I'm thinking about it, I don't see why not. It is a 12 lines of code and it makes the user experience more customizable. What is your opinion? 

The part that we are gradually falling behind is a better organization of options in the configuration window. I should look into this and come up with a proposal on how to make the UI feel less cluttered.

As a general question: how hard is it in Qt/QML to add a fuzzy finder search box at the top of the configuration window, so that typing some keyword would in realtime filter the options?

schwubdiwub:
@mmahmoudian I made the wording consistent as requested.

Another thought I now have about this is, that this is a global option and can't be changed on the fly. In case someone wants to hide/show the drop shadow of a single pin spontaneously, the user has the change the option and recreate the pin. So maybe it would be a good idea to additionally add a "drop shadow toggle" option to the context menu of the pins. 

And right now a change in the option does not change the drop shadow visibility of existing pins. So that might be something to consider, too.

PR #3430, opened 2023-12-05
2023-12-05 Add new drawing tool 'LockPick'

hugovonhorse:
Hi, this is only my 2nd ever gitHub submission (after practising with the 'first-contributions' project), so apologies for any rubbish. 

I use Flameshot on a daily basis in my professional life (mainly for creating images of custom software for documentation, bug reports, etc) and I wanted what im describing as a "LockPick" but is really just a type of 'arrow', in the style that I've now created.

I checked the code for how the arrow worked and the graphics component seemed pretty simple, so i forked, copied the entire arrowTool but renamed it and gave it unique function names etc, tested, committed, pushed and am now asking for a 'Pull'. 

Note: ive added the new icon to the far-right of the captureToolButton widget because, although its not with the other drawing tools (which would be the logical place to put it), I didnt want to interfere with other peoples muscle-memory of where to find the various buttons.  Im happy for this to be moved around though if some underlying design principle requires it :+1: 

Thanks muchly,
Hugo
linuxgamingy@gmail.com


borgmanJeremy:
Is this style of arrow common in other places? I've never seen something like it before.

@mmahmoudian  do you have thoughts on this one? I can help clean up the code a bit if we want to add the feature.

hugovonhorse:
No I don't believe it is.  I needed a way to document low and high powered
connections in an electrical project but without being able to use colour.

More than happy for it be cleaned up 👍

mmahmoudian:
@hugovonhorse First things first, thank you for the PR and the time you have put into this. Also, hopefully we can see more contribution to FLOSS world from you :)

Interesting PR. The shape is odd to me, and I found nothing similar when I search for "half arrow types" or "lockpick arrow". The closest I have found is from this post:

https://mathematica.stackexchange.com/a/229834



@hugovonhorse Is this something technical in your professional field?

My understanding is that it is alternating half of the arrow from each part.

@borgmanJeremy To be honest I don't see any harm in adding this, but I think to avoid cluttering the UI I would by default hide it. I'm interested to know the opinion of few others:
@FelixJochems @jack9603301 @panpuchkov @veracioux 

If we end up merging this, I would suggest some adjustments to the SVGs. For example the angles are not perpendicular:

image

I spent some time on the shape and I came up with some variations. The red one is direct copy of the `black/lockpick.svg` from this PR which is rotated to align with vertical line in Inkscape:

bitmap

I believe the middle black one is what the red one is supposed to be, just without any overlap and with perpendicular angles. The one on the left is the most conventional one that came up in many of my online searches, and the right black one is just to make it symmetric and solid fill on one side, and hollow fill on the other. I also like the arrows from that StackExchange that I initially mentioned. it is more minimalistic and clean. But these are just me thinking out loud.

hugovonhorse:
@mmahmouor thanks for spending the time you have looking into this - really
wasn't expecting it to get any attention honestly :)  Also i wish i did
have more time for contributing to FLOSS stuff but alas i just rarely do
(unless like in this case, i really have a motive to do so).

Whilst I use it in a professional capacity, the arrow design isn't from any
ISO or otherwise standard that I'm aware of for representing a certain type
of event/connection etc. I cant remember what name I committed the 'new
arrow shape' as (if it was "Lockpick" that surprises me because it seems a
little bit too intelligent a word for me and its not really what I was
using it for lol) - but , and this will contribute to my feelings on the
shape-design in a moment, i was aiming for something a bit like a
thunderbolt / lightning-strike.

With that in mind (although I fully appreciate the research you carried out
and the effort on the design mock-ups) I personally still prefer the red
arrow in the pictures provided as the slightly overlapping 'halves' is
closer looking to that lightning-strike idea than the proposed "middle of
the black ones".. On the other hand, your proposed shape is actually more
generally useful as it looks more professional AND gives people an
acceptable and *smarter* option for an additional type of arrow-head.

I completely agree with your suggesting that even if anything resembling it
is added, it should probably be hidden by default - I use FlameShot
literally every day, not hundreds of times every day, but certainly every
day - thank you to @everyone who contributed towards this! - and the UI is
certainly close to bursting point. I also convert as many people as I can
to using it.

Anyway, if this does get merged, seeing as someone else will be tidying up
the code I don't have any problem with how it really looks. I think to have
the choice of multiple arrow heads is a useful feature in FlameShot
nevertheless and it's a pretty simple change that, if hidden by default,
shouldn't hurt anyone 👍

Thanks again for everyone's time on this!
Regards
Hugo

PR #3439, opened 2023-12-16
Fix for aspect ratio losing ratio 

### LINKED ISSUE #3436: Using `CTRL` to resize and maintain aspect ratio doesn't work if the ratio is not 1:1
### Flameshot Version

commit `96541b2c6c02ee36f8f660a6c094b449b7a4d936`

### Installation Type

Compiled from source

linux XUbuntu

### Description

`CTRL`+ drag  

### Steps to reproduce

1. Open flameshot in any mode
2. Make sure you have flameshot in the size ratio you want
3. Resize the window in any way repeatedly (for wide rectangles, it's more easily seen horizontally)
4. Ratio is now changed until reaching `1:1` ratio at which it stays.

All resizing was done while keeping `CTRL` pressed continuously.
[cannotMaintainRatio.webm](https://github.com/flameshot-org/flameshot/assets/639467/45d97598-da9b-439f-97a9-0d4332290c5a)

### System Information

(I believe this doesn't make a difference but ok)
1. xubuntu 22.04
2. 
image

brunoais:
Fix for aspect ratio losing ratio (bad code version)

The reason why it's bad code is because I'm using a global variable for this purpose and that is a really bad idea. Unfortunately, I don't know where can this information be stored.

**Please help:** The part I'm missing is somewhere to store the float variable. Once I have a suitable place, this PR is complete.

Note: I'm not used to C++ code, this is my first time working with C++ code without prior help, please be patient.

Fixes #3436


src/widgets/capture/selectionwidget.cpp
```diff
@@ -215,9 +218,11 @@ void SelectionWidget::parentMouseMoveEvent(QMouseEvent* e)
     }
 
     auto geom = geometry();
-    float aspectRatio = (float)geom.width() / (float)geom.height();
     bool symmetryMod = qApp->keyboardModifiers() & Qt::ShiftModifier;
     bool preserveAspect = qApp->keyboardModifiers() & Qt::ControlModifier;
+    if(aspectRatio < 0 || ! preserveAspect){
```
borgmanJeremy, line 223: whats the reason for checking if the aspectRatio is < 0?

mmahmoudian: I think because of line 194, this if statement means "if the user have not forced certain aspect ratio (and if it is not marked to be preserved), then ..."

But based on what I can see, aspect ratio will always be `-1.0` before this `if`. Unless in C++ the variables are not bound to the function scope and line 224 in changes the global value of the variable.

Issue #3022, opened 2022-12-14
MacOS: Pinned image jumps from original spot


bsavant:
### Flameshot Version

Flameshot v12.1.0 (96c2c82e)
Compiled with Qt 5.15.5
darwin: 22.1.0
osx: 13.0

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

MacOS

### Description

After selecting a region and pinning the image, the pinned image moves from the original selection.  This does not happen with the same version of flameshot on Linux Mint.

### Steps to reproduce

1. launch flameshot app
2. Use hotkey to capture region
3. select a region to capture an image
4. pin image

### MacOS
![MacOS](https://user-images.githubusercontent.com/209364/207498898-ccd2a98e-63ce-4174-b72b-8fe557a8d4a4.gif)

### Linux Mint
![Linux Mint](https://user-images.githubusercontent.com/209364/207498974-da3b6424-a0e3-4c0a-a6da-f603ed1b5cdc.GIF)

### System Information

MacOS 13.0.1 Ventura (Intel)

mmahmoudian:
This is how macOS works. You can make Finder full screen and then back to windowed and see the same animation. There is nothing we can do about this until and unless Qt implements a feature that let us bypass this behavior (very unlikely)

bsavant:
I'm not referring to the animation, I understand that's a feature of MacOS.  What I'm referring to is the fact that the pinned screenshot that remains is offset from the original selection region.

mmahmoudian:
@bsavant Thanks for the explanation. I re-opened this. I don't have access to a mac to test/reproduce this. Let's see if other devs have.

jorpilo:
I believe this is similar to #2612.

I can verify the issue happens on my M1 and window is shifted downwards regardless if it's full screen or not.

@mmahmoudian how can I help to solve this issue? this is being a very frustrating problem for me

mmahmoudian:
@jorpilo I have a hunch that this downward shift is due to the top menu bar in mac. Would you please move your menu bar once to left and once to bottom and see if the shift is always downwards or it is dependent on the bar position

jorpilo:
I can confirm that the issue still happens regardless if menu bar is on the left, right or bottom position and the screen always shifts a few pixels down losing the bottom part of the screen.

yzq-del:
> You

Why do other QT screenshot tools not have this animation effect？such as snipaste

yzq-del:
[...]

PR #3445, opened 2023-12-25
fix for QRect keep width and height

lededev:
setTopLeft() call here will make QRect width and height changed, should use moveTo().
eg. when widgetOffset(-100, -100), geometry(x:0,y:0,w:100, h:100).
after setTopLeft(), geometry(x:-100,y:-100,w:200,h:200)
use moveTo, geometry(x:-100,y:-100,w:100,h:100)
fix i[ssue#3022](https://github.com/flameshot-org/flameshot/issues/3022)


borgmanJeremy:
Thank you! Can you rebase this on master to kick off the now fixed CI?

borgmanJeremy (approved):

borgmanJeremy:
I meant to post my comment here not the issue: https://github.com/flameshot-org/flameshot/issues/3022#issuecomment-2917864530

But this PR does not resolve the issue when I test it.

PR #3451, opened 2023-12-29
Include showStartupLaunchMessage in example ini file

hekazu:
I found out about this option from Home Manager setting examples and was surprised to see it was not included in the official ini example. Adding it there should help discoverability of the setting.


hekazu:
Forgot to use a non-master branch, reopening in a bit

PR #3477, opened 2024-01-22
Add new option where arrow can be drawn in a reveresed way. Fixed #2144

### LINKED ISSUE #2144: Setting to draw arrows pointer first
### Feature Description

It'd be awesome if there was a setting to that let the arrows draw using the first click to set the pointer of the arrow, then being able to drag the tail where ever (reversing the current functionality). 

The current tail first approach makes the arrow very difficult to use for more complex pictures (especially when doing screenshots of logs). Right now, I need to drop the tail end of the arrow off some place else, then re-find the content I want to draw attention to. There is more planning that needs to go into making sure that the arrow doesn't overlap any needed content while doing it's job.

lvhungdev:
This PR attempts to fix #2144 where arrow can be drawn in a reversed way.
The head will be drawn first, the tail later.
This behavior can be toggled via configuration.

TODO Add showcase videos and images.

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -65,6 +65,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initShowMagnifier();
     initSquareMagnifier();
     initJpegQuality();
+    initReverseArrow();
     // this has to be at the end
     initConfigButtons();
     updateComponents();
@@ -93,6 +94,7 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
     m_showMagnifier->setChecked(config.showMagnifier());
     m_squareMagnifier->setChecked(config.squareMagnifier());
     m_saveLastRegion->setChecked(config.saveLastRegion());
+    m_reverseArrow->setChecked(config.reverseArrow());
 
 #if !defined(Q_OS_WIN)
     m_autoCloseIdleDaemon->setChecked(config.autoCloseIdleDaemon());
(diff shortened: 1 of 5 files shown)


lvhungdev:
Hi. This PR has been sitting for a while without any updates. Am I missing something to get this PR reviewed?

UnkwUsr:
Throwing random proposal for discussion:

Maybe it can toggle between default state by holding some key modifier, like `alt` or `ctrl` (instead of checkbox) (but yes, then it is harder to find out about this feature even exists)?

mmahmoudian:
@lvhungdev Thanks for the PR. We are waiting for this PR to be approved. I generally support the idea and the PR but conventionally I will not attempt to merge code changes without approval of our C++ devs.

While we are here, would it be possible for you to add the <kbd>Alt</kbd> to toggle this on the fly? i mean whatever the setting/config is, the <kbd>Alt</kbd> flips the direction as long as <kbd>Alt</kbd> is pressed?

@UnkwUsr Thank you for the suggestion. I appreciate your input. I just want to point out that <kbd>Ctrl</kbd> is already is use. while drawing an arrow or line, holding <kbd>Ctrl</kbd> would snap the arrow to 45 degree angles. I believe the <kbd>Alt</kbd> is a better choice as also it "alters" the behavior :)

eepp:
> I agree with @mmahmoudian  that this should be toggled on the fly with a hotkey rather than a config option. 

I say: why not both?

Let's make the hot key reverse the configured default direction.

I don't see a reason why I would ever want to draw from the tail (the current behavior), therefore I don't need to press the hot key every time.

jack9603301:
I totally agree with this PR, it's really cool, I watched the video, thanks for your contribution

borgmanJeremy (approved):

borgmanJeremy:
Okay seems consensus is people want this one. We can add alt in a different PR. Please rebase on master to kick of the CI, then when that passes I'll merge.

borgmanJeremy:
@lvhungdev are you able to rebase this? If not I can manually take this patch into a new PR

lvhungdev:
Hi @borgmanJeremy, I accidentally merged the master to my feature branch. Is this acceptable or do we explicitly need to rebase?

borgmanJeremy:
Thanks fine too, I'll squash it before merging. Thank you!

Issue #3522, opened 2024-03-06
Tool settings is sized with the primary monitor, goes out of screen if the left screen is smaller than that


aoreskovic:
### Flameshot Version

Flameshot V12.1.0 (96c2c82e)

![image](https://github.com/flameshot-org/flameshot/assets/30411098/14f2a4a4-d8e7-42ba-a231-842fd4e421f3)

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...), Using the ready-made package from Github Releases

### Operating System type and version

Windows 11

### Description

My theory is that Tool settings is sized with the primary monitor, goes out of screen if the left screen is smaller than that.

My monitor configuration is like this:
![Screenshot 2024-03-06 110715](https://github.com/flameshot-org/flameshot/assets/30411098/dc2c5394-a5e8-40a1-8771-ffcc17e7bbe8)

SO when I start annotation and start placing arrows/lines and such, I can't see first few tools. Example:
Leftmost monitor
![Screenshot 2024-03-06 110601](https://github.com/flameshot-org/flameshot/assets/30411098/76ecd62e-68c5-4680-9bcb-4e149259e70a)
Middle Monitor: 
![Screenshot 2024-03-06 110611](https://github.com/flameshot-org/flameshot/assets/30411098/51d13928-d891-46da-bcb7-4ebcdf346bb7)

My theory is that rest of the menu is here, out of screen:

![Screenshot 2024-03-06 110715 - Copy](https://github.com/flameshot-org/flameshot/assets/30411098/6e828259-7654-4c67-82cd-51f5b33c2c4c)

### System Information

![image](https://github.com/flameshot-org/flameshot/assets/30411098/54b08ba3-2c75-4a41-acbf-ebcd906d5b0f)

![Screenshot 2024-03-06 110715](https://github.com/flameshot-org/flameshot/assets/30411098/7d3ad566-9cc1-445c-a1f3-aaabb0b56b43)

theafolabijo:
I think the issue still exists.

<img width="193" height="302" alt="Image" src="https://github.com/user-attachments/assets/ce4fc691-dab0-4529-89fd-fb81ec69d808" />
<img width="1026" height="624" alt="Image" src="https://github.com/user-attachments/assets/4f298b08-4477-46a8-b330-65f3834494b2" />

![Image](https://github.com/user-attachments/assets/f317a7d5-5b9f-4fbb-b853-19cafe60c223)

[...]

PR #3550, opened 2024-03-31
Fix for side panel widget out of screen #3522

petrugrd:
This fixes the issue #3522

The problem was that the panel was moved relative to (0, 0) coordinates, which is not an issue in case of single display, or multiple displays in the same row.
Since capture widget is placed on all screens, the point (0, 0) will be the top left corner of virtual screen instead of top left corner of the current screen concluding from QDesktopWidget documentation (*). This caused the panel to be shifted up if the height of the primary screen was smaller than the virtual one.

 * https://doc.qt.io/qt-5/qdesktopwidget.html#screen-geometry

flameshot-issue-3522


mmahmoudian:
@petrugrd Thanks. Looking forward for your PR to get out of draft and merge ready. Also thanks for using the Windows CI.

petrugrd:
Hi @mmahmoudian,
Would be nice if someone tested this first on a Mac with the same screen configuration as pointed in the screenshot. I have doubts that more changes have to be done, which may or may not break other stuff, especially removing the next `#if (defined(Q_OS_MACOS))` blocks:

``` patch
diff --git a/src/widgets/panel/utilitypanel.cpp b/src/widgets/panel/utilitypanel.cpp
index 5fc1889e2f..4fd9310fbd 100644
--- a/src/widgets/panel/utilitypanel.cpp
+++ b/src/widgets/panel/utilitypanel.cpp
@@ -42,7 +42,7 @@ UtilityPanel::UtilityPanel(CaptureWidget* captureWidget)
             m_internalPanel,
             &QWidget::hide);
 
-#if (defined(Q_OS_WIN) || defined(Q_OS_MACOS))
+#if (defined(Q_OS_MACOS))
     move(0, 0);
 #endif
     hide();
@@ -87,7 +87,7 @@ void UtilityPanel::show()
     m_showAnimation->setEndValue(QRect(0, 0, width(), height()));
     m_internalPanel->show();
     m_showAnimation->start();
-#if (defined(Q_OS_WIN) || defined(Q_OS_MACOS))
+#if (defined(Q_OS_MACOS))
     move(0, 0);
 #endif
     QWidget::show();
```

mmahmoudian:
@petrugrd I have no access to MacOS. Maybe @panpuchkov has some access to a mylti-monitor Mac setup.

petrugrd:
Hi @mmahmoudian , @panpuchkov. Is there any chance to test this PR on a Mac setup, or it can be reviewed and merged as it is?

src/widgets/panel/sidepanelwidget.cpp
```diff
@@ -81,7 +81,7 @@ SidePanelWidget::SidePanelWidget(QPixmap* p, QWidget* parent)
     m_layout->addWidget(m_colorWheel);
     m_layout->addWidget(m_colorHex);
 
-    QHBoxLayout* gridHBoxLayout = new QHBoxLayout(this);
+    QHBoxLayout* gridHBoxLayout = new QHBoxLayout();
```
borgmanJeremy, line 84: Why did you drop (this) from the constructor?

borgmanJeremy:
I will test this on Mac as well.

fishtopher:
I don't have a Mac to test and confirm it works, but wondering if you managed to do it?
I'd love if this could be merged - I can't see the tools because my right screen is 4k and the left screen is not 😅 

(I'm not a contributor, but I really appreciate all the work everyone is doing on Flameshot!)

PR #3562, opened 2024-04-09
fix content-type header in imgur uploader request

3dprogramin:
The request made to imgur has an incorrect content-type header.

Original (incorrect):
*application/x-www-form-urlencoded*

Modified (correct):
**application/x-www-form-urlencoded**


borgmanJeremy:
since imgur is deprecated i am going to close this

PR #3586, opened 2024-04-28
Update Windows-pack.yml: update dependencies

Saibamen:
Should speedup artifacts upload: https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/

Fix deprecated warnings

PR #3587, opened 2024-04-28
Fix update cis

mmahmoudian:
I've updated everything to the latest minor or major version depending on if they have their major version tag following their latest stable minor version tag or not.

I have kept the `actions/upload-artifact` back in v3 because:
1. It seems there are breaking changes between `v3` -> `v4` ([more info](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md))
2. there is a PR (#3586) that have tried to update this for the Windows CI, so I would not change it to avoid duplicate and confusion

### DIFF
--- a/.github/workflows/Linux-pack.yml
+++ b/.github/workflows/Linux-pack.yml
@@ -563,7 +563,7 @@ jobs:
           echo "================================"
           echo "VERSION=${last_committed_tag:1}" >> $GITHUB_ENV
       - name: Packaging snap
-        uses: snapcore/action-build@v1
+        uses: snapcore/action-build@v1.2.0
         id: snapcraft
         with: 
           snapcraft-args: --enable-experimental-extensions
--- a/.github/workflows/Linux-pack.yml
+++ b/.github/workflows/Linux-pack.yml
@@ -2,16 +2,24 @@ name: Packaging(Linux)
 
 on:
   push:
-    branches:
-      - master
+    branches: [ master ]
     paths-ignore:
       - 'README.md'
       - 'LICENSE'
+      - 'CODE_OF_CONDUCT.md'
+      - 'shell.nix'
+      - 'docs/**'
 
   pull_request:
+    branches: [ master ]
     paths-ignore:
       - 'README.md'
       - 'LICENSE'
+      - 'CODE_OF_CONDUCT.md'
+      - 'shell.nix'
+      - 'docs/**'
+
+  workflow_dispatch:
 
 env:
   PRODUCT: flameshot
--- a/.github/workflows/Windows-pack.yml
+++ b/.github/workflows/Windows-pack.yml
@@ -90,7 +90,7 @@ jobs:
           echo "VERSION=$(cat CMakeLists.txt |grep 'set.*(.*FLAMESHOT_VERSION' | sed 's/[^0-9.]*//' |sed 's/)//g')" >> $GITHUB_ENV
 
       - name: Restore from cache and run vcpkg
-        uses: lukka/run-vcpkg@v11
+        uses: lukka/run-vcpkg@v11.5
         with:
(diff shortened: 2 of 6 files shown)


.github/workflows/Linux-pack.yml
```diff
@@ -555,7 +563,7 @@ jobs:
           echo "================================"
           echo "VERSION=${last_committed_tag:1}" >> $GITHUB_ENV
       - name: Packaging snap
-        uses: snapcore/action-build@v1
+        uses: snapcore/action-build@v1.2.0
```
Saibamen, line 566:
```suggestion
        uses: snapcore/action-build@v1
```

`v1` will install latest v1 version ;)

.github/workflows/Windows-pack.yml
```diff
@@ -84,7 +90,7 @@ jobs:
           echo "VERSION=$(cat CMakeLists.txt |grep 'set.*(.*FLAMESHOT_VERSION' | sed 's/[^0-9.]*//' |sed 's/)//g')" >> $GITHUB_ENV
 
       - name: Restore from cache and run vcpkg
-        uses: lukka/run-vcpkg@v11
+        uses: lukka/run-vcpkg@v11.5
```
Saibamen, line 93:
```suggestion
        uses: lukka/run-vcpkg@v11
```

.github/workflows/Windows-pack.yml
```diff
@@ -94,13 +100,13 @@ jobs:
...
         id: cache-qt
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: ./build/Qt/${{ matrix.qt_ver }}/${{ matrix.config.qt_arch_install }}
           key: ${{ runner.os }}-QtCache/${{ matrix.qt_ver }}/${{ matrix.config.qt_arch }}
 
       - name: Install Qt
-        uses: jurplel/install-qt-action@v2
+        uses: jurplel/install-qt-action@v3.3.0
```
Saibamen, line 109:
```suggestion
        uses: jurplel/install-qt-action@v3
```

.github/workflows/build_cmake.yml
```diff
@@ -101,13 +110,13 @@ jobs:
...
         id: cache-qt
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: ./build/Qt
           key: ${{ runner.os }}-QtCache
 
       - name: Install Qt
-        uses: jurplel/install-qt-action@v2
+        uses: jurplel/install-qt-action@v3.3.0
```
Saibamen, line 119:
```suggestion
        uses: jurplel/install-qt-action@v3
```

Saibamen (changes requested):

mmahmoudian:
@Saibamen It is not clear in [the Github documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses) if there is a difference between `@v4` and `@v4.3.2`. Bascially if `@v4` means `4.0.0` tag or `4` tag, or latest of `4.x.x` tags.

In some workflows, they move their major version tag to the latest minor version (e.g `v1` always updated to the latest of `v1.x.x`: https://github.com/snapcore/action-build/tags ), but this does not happen in all workflows (e.g.  https://github.com/jurplel/install-qt-action/tags )

Do you know how for example the `@v4` is handled by Github Actions?

Saibamen:
You can use https://github.com/saadmk11/github-actions-version-updater

holazt:
`actions/upload-artifact` v4 can't merge artifacts with the same name, v3 can. Nevertheless I think you can update `actions/upload-artifact` to v4, but you'll need to change the artifacts name, e.g. add some architecture or distribution information.

borgmanJeremy:
@mmahmoudian I think this conflicts with other PR's that were merged. Is there still work to do on this or should it be closed?

Issue #1341, opened 2021-02-17
[Feature Request] Remap print-screen activation key on Windows


T3rminat0r:
**I'd love to modify the Shortcut used for a Screencapture on windows (flameshot already running)**

Flameshot-Version: 0.8.4 (couldn't find any newer Windows MSI package)

OS: Windows 10 (20H2)

I've been using flameshot on Linux, for a long time, and never found a suitable alternative for use on Windows, so, I was happy to find out there's windows versions available, now (didn't check in a while...)
On Linux, I just set up a global shortcut (ctrl+y, in my case) to start flameshot gui, and am quite happy with that. On Windows, I can't (easily?) do that (at least I've not found any documentation for it, and plainly calling flameshot.exe gui does nothing)
The Windows Flameshot does come with local Shortcuts, though, while it's running... I'd just love to remap the PrintScreen one to a different shortcut (either in the .conf file or via a UI?) 
Is this still a thing, and I just missed the documentation for it? Is it maybe already planned for a comming release? Or am I just thinking in the wrong direction and the solution to my problem is something else? (please don't say "just press printscreen" ;) )

Thanks for developing this great tool, there literally is none with as great a UX feeling as flameshot.

borgmanJeremy:
just press printscreen :) (just kidding)

Right now there is not a way to override the printscreen activation on windows. I've tagged this as a future feature request.

borgmanJeremy:
Believe this is the relevant section of code: https://github.com/flameshot-org/flameshot/blob/b5ca48dde0102a6a6256c65762dffe582d05996e/src/core/controller.cpp#L85

Quite busy with the 0.9 release so I will come back to this later or someone else might tackle it.

dontworryseoul:
> **I'd love to modify the Shortcut used for a Screencapture on windows (flameshot already running)**

I need the feature too.

WeridoLm:
I need the feature too.thx

WeridoLm:
I need the feature too.thx

WeridoLm:
I need the feature too.thx

TroyWilliams3687:
[...]

PR #3593, opened 2024-05-04
WIP: Add example of custom global shortcut on Windows

dsatyanov:
Example of custom global shortcut on Windows - hardcoded Alt+Shift+3.
It's just prove of the concept

Feature discussed here:
https://github.com/flameshot-org/flameshot/issues/1341
https://github.com/flameshot-org/flameshot/pull/2168
and in other places - please link/comment them if they're needed here

@mmahmoudian It looks like we actually don't need [CLI on Windows](https://github.com/flameshot-org/flameshot/issues/1341#issuecomment-1126314750) to have custom global shortcuts on this OS, is it correct?
Please tell me if it's worth to finish this solution to have fully customizable shortcuts on Windows, so I'll try to do it.

P.S. I'm not a c++ dev, I've found the solution from ChatGPT, so maybe it's a really bad code but it seems to work for me for now

### DIFF
--- a/src/core/flameshot.cpp
+++ b/src/core/flameshot.cpp
@@ -37,6 +37,41 @@
 #include <QScreen>
 #endif
 
+#if defined(Q_OS_WIN)
+#include <Windows.h>
+
+HHOOK hook;
+bool isAltPressed = false;
+bool isShiftPressed = false;
+
+// Just an example of global shortcuts on Windows.
+// This is hardcoded Alt+Shift+3 shortcut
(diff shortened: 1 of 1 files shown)


mmahmoudian:
>  It looks like we actually don't need CLI on Windows to have custom global shortcuts on this OS, is it correct?

I haven't used or touched Windows in years. My suggestion is 


> Please tell me if it's worth to finish this solution to have fully customizable shortcuts on Windows, so I'll try to do it.

I bet it does. But I mainly do triage in this repo. C++ devs are @veracioux and @panpuchkov . Perhaps @jack9603301 can also give us his opinion

Issue #3529, opened 2024-03-11
Hyprland Wayland Copy to Clipboard doesn't work


enk-it:
### Flameshot Version

Flameshot v12.1.0 (3d21e496)

### Installation Type

User repository (AUR)

### Operating System type and version

ArcoLinux

### Description

I use Hyprland and copy to clipboard works only if i'm currently focused on some application.
For example i have 3 applications opened: yandex-browser-stable, discord and xfce4-terminal. If i'm focused on discord, i can take screenshot and copy it to CB (i can take any region by the way not only discord), but if i'm, focused on yandex or terminal i'm unable to copy screenshot to CB. Saving screenshot in file works fine in any scenarios. 
I start flameshot this way:
`flameshot gui`
When focused on  gnome-calculator or Insomnia Flameshot behaves fine. 
focused on Galculator doesn't copy as well

### Steps to reproduce

1) Focus on xfce4-terminal
2) start flameshot with 
`flameshot gui`
3) press ctrl+c or double click on region (if activated option in settings)

### System Information

System:
  Host: silencer-a320mh Kernel: 6.7.9-arch1-1 arch: x86_64 bits: 64
  Desktop: Hyprland v: 0.36.0-77-g717d5b3c Distro: ArcoLinux
Graphics:
  Device-1: NVIDIA TU117 [GeForce GTX 1650] driver: nvidia v: 550.54.14
  Display: wayland server: X.org v: 1.21.1.11 with: Xwayland v: 23.2.4
    compositor: Hyprland v: 0.36.0-77-g717d5b3c driver: X: loaded: nvidia
    unloaded: modesetting gpu: nvidia,nvidia-nvswitch
    resolution: 1920x1080~144Hz
  API: EGL Message: EGL data requires eglinfo. Check --recommends.

MrMasrozYTLIVE:
Can confirm on:

User repository (AUR) - flameshot-git
ArchLinux 6.8.1-arch1-1
Hypr (Hyprland 0.35.0-2)

gwstorm:
fwiw, I was cruising through here for a different issue and figured I'd mention that I can't duplicate this one. 

hyprland 37.1
flameshot-git
arch 6.8.1

MrMasrozYTLIVE:
> fwiw, I was cruising through here for a different issue and figured I'd mention that I can't duplicate this one.
> 
> hyprland 37.1 flameshot-git arch 6.8.1

Got an update. I cant replicate it neither now

[...]

Issue #3479, opened 2024-01-24
Screenshots aren't copied to the clipboard


OnkelTem:
### Flameshot Version

Flameshot v12.1.0 (96c2c82e)
Compiled with Qt 5.12.8

### Installation Type

AppImage

### Operating System type and version

Linux Kubuntu 22.04

### Description

I can't copy screenshot to the clipboard. When I click this button, flameshot silently exists.
I activate Flameshot via Print Screen, which is assigned to `flameshot gui --delay 2500`

Currently the only option left working is saving screenshot to file :(

### Steps to reproduce

1. Import KDE shortcuts
2. Assign Print Screen to "Take screenshot with delay" which is the second option with command `flameshot gui --delay 2500`
3. Press Print Screen, make selection
4. Press copy. 

### System Information

```
$ inxi --width 80 --system --graphics
System:
  Host: fractal Kernel: 6.5.0-14-generic x86_64 bits: 64
    Desktop: KDE Plasma 5.24.7 Distro: Ubuntu 22.04.3 LTS (Jammy Jellyfish)
Graphics:
  Device-1: NVIDIA GA104 [GeForce RTX 3060 Ti Lite Hash Rate] driver: nvidia
    v: 525.147.05
  Device-2: Logitech BRIO Ultra HD Webcam type: USB
    driver: hid-generic,snd-usb-audio,usbhid,uvcvideo
  Display: x11 server: X.Org v: 1.21.1.4 driver: X: loaded: nvidia
    unloaded: fbdev,modesetting,nouveau,vesa gpu: nvidia resolution:
    1: 2560x1440~144Hz 2: 1920x1080~60Hz 3: 1920x1080~60Hz
  OpenGL: renderer: NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
    v: 4.6.0 NVIDIA 525.147.05
```

mmahmoudian:
Flameshot definitely works flawlessly on X11.

Based on your text I think you are not experienced enough with Linux. So here is what you should do:
1. If you are using AppImage, you cannot use `flameshot`. You have to run that Flameshot AppImage (this is how AppImage works). We have clearly explained it in our documentation ( https://flameshot.org/docs/installation/installation-linux/#appimage )
2. remove the `--delay 500`, that is not needed, it will just add half a second delay
[...]

PR #3598, opened 2024-05-07
Don't initialize Notifications dbus if disabled

insunaa:
This PR fixes an issue where Flameshot hangs while it's waiting for the notifications QDBusInterface to initialize.

Some systems don't come with a notifications daemon by default, which causes the QDBusInterface constructor to block for 5-20 seconds, during which flameshot hangs and cannot copy the screenshot into the clipboard buffer.

I tested this on Linux on KDE Plasma 6 with Wayland enabled. CMake flags are `cmake .. -GNinja -DUSE_WAYLAND_CLIPBOARD=1`

- ~~might fix https://github.com/flameshot-org/flameshot/issues/3529~~
- ~~might fix https://github.com/flameshot-org/flameshot/issues/3479~~

~~Only *might* fix, because the assumption is that the issue names are misnomers, and that the true issue is just that the screenshot is copied to clipboard after 5-20 seconds without any hint (except systemd logs)~~

Edit: Ignore the potentially fixed issues, there are more things afoot with those.

### DIFF
--- a/src/utils/systemnotification.cpp
+++ b/src/utils/systemnotification.cpp
@@ -24,6 +24,9 @@ SystemNotification::SystemNotification(QObject* parent)
   , m_interface(nullptr)
 {
 #if !(defined(Q_OS_MACOS) || defined(Q_OS_WIN))
+    if (!ConfigHandler().showDesktopNotification()) {
+        return;
+    }
     m_interface =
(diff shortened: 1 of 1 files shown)


untainsYD:
I do vote for this PR, it's essential to fix it!

borgmanJeremy (approved):

borgmanJeremy:
Thank you!

PR #3604, opened 2024-05-12
Set device pixel ratio for grim screenshot

tobiolo:
Grim returns a bitmap with the physical size of the screen. This needs to be scaled back to be displayed as underlay for Flameshot on screens where logical size differs from the physical size (i.e. scale is active).

In order to do this, set the device pixel ratio, see <https://doc.qt.io/qt-5/qpixmap.html#setDevicePixelRatio>.

Please note that this works only properly with integer scale. QScreen::devicePixelRatio rounds when fractional scale is active on Wayland.


tobiolo:
Note: It should be this or #3606, but not both.

borgmanJeremy:
What happens if fractional scaling is set?

borgmanJeremy:
All backends have moved to freedesktop portal

PR #3606, opened 2024-05-12
Set grim output image scale factor to 1.0

tobiolo:
In thie case, the pixmap size matches
the logical size of the screen.

Downside of this approach is the visible image degradation in case of high scaling.


tobiolo:
Note: It should be either this or #3604

Issue #3582, opened 2024-04-24
Add an option to hide cursor on screenshots


Tiagoquix:
### Feature Description

I think it would be great if there was an option to hide cursor on screenshots.

I'm using Fedora Linux 40 KDE on Wayland.

Thanks in advance.

berkentekin:
Hello, it seems that for KDE and GNOME screenshots are requested from an external interface with no options to set whether a cursor should be hidden. I am working on providing the option for other wayland-based WMs though.

sbot50:
A solution to this on kde is to wiggle your mouse to make it bigger and THEN take a screenshot, it wont appear if you do it like that for some reason.

schworak:
That is so weird that the cursor doesn't appear in the screenshot if you wiggle it first. I suspect the OS is hiding the real cursor and adding a special larger cursor that Flameshot can't (or won't) capture. But it would be very handy to have a configuration option to hide the cursor under Wayland. I am training myself to move the cursor to the far edge of the screen, or at least out of the capture zone before trying to make a screenshot. But I don't always remember to do that. So if possible, please add this option.

tracure1337:
This option is still needed yes!

Gotoro:
Please fix this problem, it is really annoying, worked perfectly on X11

PaprusTeuton:
Still needed 2 years later, please!

PR #3607, opened 2024-05-13
New option to hide cursor for wlroots-based grim

berkentekin:
- This PR adds an option to hide cursor for screenshots taken in DEs other than GNOME and KDE. The checkbox is therefore disabled in GNOME and KDE.
- In `screenGrabber.cpp`, the application decides to use either a DBus interface or the wlroots-based `grim` program depending on the desktop environment. Unfortunately, from what I understand, KDE and GNOME do not support wlr protocols, and the DBus interface does not provide an option to show or hide the cursor. Therefore, my changes (should) only apply to DEs that can utilize grim.
- `grim` includes the cursor in screenshots with `-c` argument. This PR basically just allows passing this option.
- The cursor is hidden by default. 

Partially resolves  flameshot-org/flameshot#3582

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -79,6 +79,7 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
     m_sidePanelButton->setChecked(config.showSidePanelButton());
     m_sysNotifications->setChecked(config.showDesktopNotification());
     m_autostart->setChecked(config.startupLaunch());
+    m_hideCursor->setChecked(config.hideCursor());
     m_copyURLAfterUpload->setChecked(config.copyURLAfterUpload());
     m_saveAfterCopy->setChecked(config.saveAfterCopy());
     m_copyPathAfterSave->setChecked(config.copyPathAfterSave());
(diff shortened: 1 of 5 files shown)


mmahmoudian:
Thanks for the PR. I believe the correct course UX is to have only and only one checkbox that works universally on all DEs and OSs. What I mean is that the user should not care which feature they should toggle based on what software stack they are using. So if we cannot turn off cursor in Gnome Wayland or KDE or Windows, we should somehow show a message to the user and inform them why the checkbox is disabled.

In this implementation the Windows, Mac, and BSD users will just get a disabled checkbox without any explanation, and the Gnome and KDE users will get different tooltip. 

I believe it is better if
- we can show a red text above this checkbox to notify the user
- include Mac and BSD in the mix by reversing the if statement condition to `if linux -> if wayland -> if not gnome or not kde -> ...`

berkentekin:
@mmahmoudian Would it be acceptable if I made the entire checkbox red instead? I think it looks a bit tidier compared to adding text above it
Screenshot_20240514_182027
Screenshot_20240514_182620

mmahmoudian:
Aesthetically I like it, but the point is that users rarely check the tooltip for more info. How about adding a ℹ️ or something similar at the end of such items that kind of invite/attract the user to click/hover to get the reason on why this is red

berkentekin:
@mmahmoudian Thank you for your kind words. 

On second thought I think adding a text is a better idea precisely because of the reason you've mentioned. I just deleted my comment because it's such a minor issue... Besides, whether an unsupported environment actually grabs the cursor or not is outside our control, so adding a text directly resolves the ambiguity of a disabled-but-checked box.

src/config/generalconf.cpp
```diff
@@ -399,6 +408,57 @@ void GeneralConf::initAutostart()
...
+                &QCheckBox::clicked,
+                this,
+                &GeneralConf::hideCursorChanged);
+    };
+
+#if defined(Q_OS_LINUX)
+    if (!m_info.waylandDetected()) {
+        initDisabledHideCursor();
+    } else {
+        switch (m_info.windowManager()) {
```
borgmanJeremy, line 447: It might better to "opt in" desktops where we KNOW this feature will work, than just to opt out KDE and GNOME. We have people using basically every esoteric DE there is.

borgmanJeremy (changes requested):

oficsu:
@berkentekin 

> Unfortunately, from what I understand, KDE and GNOME do not support wlr protocols, and the DBus interface does not provide an option to show or hide the cursor

Currently, Spectacle (default KDE screenshot tool) has the checkbox to include/exclude cursor. It uses the following D-Bus API:
```
method call ... -> destination=org.kde.KWin.ScreenShot2 path=/org/kde/KWin/ScreenShot2; interface=org.kde.KWin.ScreenShot2; member=CaptureScreen
   string "eDP-1"
   array [
      dict entry(
         string "include-cursor"
         variant             boolean true
      )
      dict entry(
         string "include-decoration"
         variant             boolean true
      )
      dict entry(
         string "include-shadow"
         variant             boolean true
      )
      dict entry(
         string "native-resolution"
         variant             boolean true
      )
   ]
   ...
```

PR #3608, opened 2024-05-13
feat: ubuntu 22.04.06 instructions

jonathantito:
I added an image to be more clear and updated about configure flameshot in ubuntu 22.04.6


borgmanJeremy:
Thank you for working on this. However, I think the graphic could be improved. The numbers are added inconsistently (sometimes to the side, sometimes below) and the sequence of the numbers does not reflect the order in which the user should click the buttons. 

Could you restructure the graphic to look neater and show the user what order to press the buttons?

mmahmoudian:
@jonathantito Thank you for your contribution. This can indeed help some Ubuntu users. I agree with @borgmanJeremy that the image can be improved. Here are my suggestions in addition to what @borgmanJeremy mentioned:
1. Considering that this is a screenshot project, it would be nice if we can at the same-time showcase the power of this software. So I suggest using the "counter bubble"
2. It would be nice if the picture can be cleanly cropped so that it does contain the full window, but does not contain the content of other windows. For instance in the provided image on the left margin, the background window is visible

Perhaps you can get more inspirations from the images in this part of our documentation:

https://flameshot.org/docs/guide/imgur-help/

Issue #3287, opened 2023-08-01
Option to disable the "screenshot aborted" notification


AlexFolland:
### Feature Description

Here is my rationale for this feature request.

- Since the Flameshot overlay disappears when I abort a screenshot, I know when the screenshot was aborted, so the notification is not useful for me to see.
- While the notification is on-screen and I decide to start a new screenshot, the notification can be included in the screenshot, which I'd like to avoid.
- I find the other notifications from Flameshot useful, so I do not wish to disable all Flameshot notifications; only the "screenshot aborted" notification.

Together, these cause me to wish the "screenshot aborted" notification specifically was either removed or had an option to disable it.

If there is an option to disable it, I suggest disabling the "screenshot aborted" notification by default, since it seems unlikely to me that someone might need to see such a notification.

For context, this is the notification to which I'm referring.

![image](https://github.com/flameshot-org/flameshot/assets/601517/c573793b-4456-4104-977d-e097212195d8)

mmahmoudian:
Thanks for the suggestion. I myself also find it a little bit annoying but it disappears fast enough that I never thought of it as an issue. I personally completely agree with making it configurable.

At the moment user can completely turn off all notifications (even the very useful ones):
![image](https://github.com/flameshot-org/flameshot/assets/390889/1faf8500-7ec2-4f3d-ac16-b2f1785bd94e)


> If there is an option to disable it, I suggest disabling the "screenshot aborted" notification by default, since it seems unlikely to me that someone might need to see such a notification.

Well, this is the part that I disagree for tow reasons:
1. the notification can help the new user to understand the dynamic of the tool. This is extra important when something cancels Flameshot automatically.
2. disabling it by default also means that the current default behavior of Flameshot would change. And we all know how certain people react when something visual changes.

I think it should be Enabled by default, at least for the next few versions, and then we can ask the community how they feel about it. This way the backlash would be minimal.

AlexP11223:
How do you abort? I have never seen it.

AlexFolland:
> How do you abort? I have never seen it.

This is how:
![image](https://github.com/flameshot-org/flameshot/assets/601517/0be2fb73-258f-4e12-bfcb-2bc584ce4908)

AlexP11223:
hm, I tried Esc, no notification. Other notifications appear fine, e.g. after saving.

KDE, Arch Linux.

AlexFolland:
[...]

PR #3610, opened 2024-05-14
Add option to disable abort notification

j4k0xb:
Closes https://github.com/flameshot-org/flameshot/issues/3287

It's my first PR here so not too familiar with the codebase, hope the way of excluding the logger target is how it's intended to be used (still logs to files/stderr).
I ran clang-format and tested it on Linux.

> I think it should be Enabled by default, at least for the next few versions, and then we can ask the community how they feel about it. This way the backlash would be minimal.
(https://github.com/flameshot-org/flameshot/issues/3287#issuecomment-1664139462)

image

### DIFF
--- a/flameshot.example.ini
+++ b/flameshot.example.ini
@@ -42,6 +42,9 @@
 ;; Show desktop notifications (bool)
 ;showDesktopNotification=true
 ;
+;; Show abort notifications (bool)
+;showAbortNotification=true
+;
 ;; Filename pattern using C++ strftime formatting
 ;filenamePattern=%F_%H-%M
 ;
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -38,6 +38,7 @@ GeneralConf::GeneralConf(QWidget* parent)
 #endif
     initShowTrayIcon();
     initShowDesktopNotification();
+    initShowAbortNotification();
 #if !defined(DISABLE_UPDATE_CHECKER)
     initCheckForUpdates();
 #endif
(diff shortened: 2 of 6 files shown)


AlexFolland:
I'm not a maintainer, but I am a programmer and this looks great to me.  I read the code changes and they make sense.  The logging is preserved and only the notification is toggled by this option.

Is there any reason not to merge this at this point?  It makes the software better and from my review of the code, I don't see how it could introduce any regressions.

mmahmoudian:
@j4k0xb thanks for implementation, and @AlexFolland thank for reviewing the code. I'm a maintainer but I mostly do triage. We should wait for other devs to review and merge this. I typically only merge documentation or minor code changes, and this PR contains 33 lines of code across 6 files.

@j4k0xb and @AlexFolland it would be a great help if you can give us your opinion on other PRs. The more eyes we have on the codebase, the safer the users will be and the better code quality we would have. I appreciate every bit of help.

Thank you again for the PR and for the review.

AlexFolland:
I'm not deeply familiar with the code base, but even I can tell that this particular PR is just a bunch of settings-related boilerplate and documentation, plus the one change that actually interprets and uses the setting's value.  It's a very simple change, so that's why I could easily claim to have reviewed it.

AlexFolland:
Please review and merge this soon.  It looks ready and the abort notification is disruptive to regular screenshot workflows, especially when capturing particular moments while gaming.  I would prefer to never see an abort notification in my next attempt at the right screenshot after starting a capture at the wrong moment.

borgmanJeremy:
I'm reviewing now. Can you help me understand why this notification needs to be treated differently than all other notifications?

j4k0xb:
I would have preferred if it didn't show this notification and there wasn't a configuration, but another maintainer wanted to keep it backwards compatible: https://github.com/flameshot-org/flameshot/issues/3287#issuecomment-1664139462

> I myself also find it a little bit annoying but it disappears fast enough that I never thought of it as an issue. I personally completely agree with making it configurable.

> disabling it by default also means that the current default behavior of Flameshot would change. And we all know how certain people react when something visual changes.
I think it should be Enabled by default, at least for the next few versions, and then we can ask the community how they feel about it. This way the backlash would be minimal.

and other types of notifications are indeed useful (for copying/sharing)

borgmanJeremy (approved):

borgmanJeremy:
Ok thanks I'm on board. Would you mind rebasing on master to get the CI to run? I think when this was first submitted the CI was broken but its working now. If not I can merge it and fix issues later.

AlexFolland:
I've tested this by installing `flameshot-git` from the AUR and enabling the new setting for it. It worked! Thank you very much!

PR #3612, opened 2024-05-18
Disable grim's png compression

MaeIsBad:
Grim by default used level 6 for compressing PDFs, which could take as long as a second depending on the hardware. This caused a big delay when taking a screenshot with grim, leading to a bad user experience. This commit sets  the compression level to 0(uncompressed) radically speeding up the process of taking a screenshot(down to just a few ms on my machine).

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -34,6 +34,9 @@ void ScreenGrabber::generalGrimScreenshot(bool& ok, QPixmap& res)
     QProcess Process;
     QString program = "grim";
     QStringList arguments;
+    // Disable PNG compression
+    arguments << "-l"
+              << "0";
     arguments << "-";
     Process.start(program, arguments);
     if (Process.waitForFinished()) {


mmahmoudian:
I'm not sure how I feel about this. On one hand you are right that it speeds up the process, on the other hand it will result in much larger PNGs (benchmarking is needed to assess how large). Perhaps a better way is to allow the user to adjust this value, potentially via a slider.

Also, before you wonder, I generally don't merge code changes unless is approved my one of the devs. As for this PR, perhaps we should wait for opinion of other devs too.

Thanks again for this PR.

MaeIsBad:
> I'm not sure how I feel about this. On one hand you are right that it speeds up the process, on the other hand it will result in much larger PNGs (benchmarking is needed to assess how large). Perhaps a better way is to allow the user to adjust this value, potentially via a slider.

This is not an issue. Flameshot recompresses the image on it's own when saving the image regardless, meaning this change doesn't impact the final file size

mmahmoudian:
After some more thoughts I'm ok with this change. Thanks for the PR.

@veracioux what do you think.

mmahmoudian (approved):

borgmanJeremy:
I think this is supersceded by #3859. If not LMK and I can take a second look

Issue #3505, opened 2024-02-20
External screens not available (until restart of Flameshot)


tpcgold:
### Flameshot Version

v12.1.0 (96c2c82e)

### Installation Type

`brew install --cask flameshot`

### Operating System type and version

Soonoma 14.3.1

### Description

Flameshot area doesn't extend to external screens when plugging them in. 
So taking a screenshot or "selection" only involves the main display of the M2 Mac. 

Tried to reinstalled etc. but the problem still occures (this problem exists since a few months)

Current workaround found:  close flameshot and reopen -> but this is really annoying over time

### Steps to reproduce

1. start Mac
2. add external Screen via usb-c/thunderbolt
3. problem occurs

### System Information

1. macOS Soonoma 14.3.1
2. Mac M2 with two Dell external Screens or Mac M2 with Samsung external Screen

lateparty:
Second this on Windows. Flameshot snapshots your multi-screen configuration on startup and then keeps that regardless of plugging or unplugging and moving to different setups with the device. Would love a fix for this.

tpcgold:
seems the problem could be fixed by using:
https://doc.qt.io/qt-6/qguiapplication.html#screenAdded

(but I'm not a contributor atm. and don't have time to read into the code. just this trigger doesn't occur in the code anywhere , although it seems it should)
Also this trigger should occur:
https://doc.qt.io/qt-6/qguiapplication.html#screenRemoved

man0s:
+1

johannesE:
I can confirm that restarting flameshot on OSX fixed the issue for me.

Saibamen:
>fixed the issue for me

This is a workaround, not fixing

leehambley:
I'm trying today to apply the advice from https://github.com/flameshot-org/flameshot/issues/3505#issuecomment-1965879269 and see if that works in general, if it does I'll link my pull request here. First impressions are good, but I'm not fluent in C++ or QT so I can't promise anything.

leehambley:
PR [here](https://github.com/flameshot-org/flameshot/pull/3618) but I don't know if it really really helps. You can try cloning my branch and building with 

```
brew install qt@5 cmake
cmake -S . -B build -DQt5_DIR=$(brew --prefix qt5)/lib/cmake/Qt5 && (cd build && make)
open build/src/flameshot.app
```

RichDCs:
[...]

PR #3618, opened 2024-05-27
[#3505] use QT APIs screenAdded and screenRemoved to reiniitalize Fla…

leehambley:
…meshot when the display configuration changes


src/main.cpp
```diff
@@ -148,6 +149,17 @@ int main(int argc, char* argv[])
         auto c = Flameshot::instance();
         FlameshotDaemon::start();
 
+        /* Register handlers for when the screen configuration changes to reinit the app */
```
Saibamen, line 152:
```suggestion
        /* Register handlers for when the screen configuration changes to reinit the app to fix #3505 */
```

Saibamen:
https://github.com/flameshot-org/flameshot/issues/3505#issuecomment-2167569019:

> @leehambley Tried it, it does detect the external monitor (even after unplugging and plugging again) 👍
> 
> However, it introduces a couple of regressions:
> 
> * The screenshot capture on the external monitor does not show any open windows. You can only screenshot an empty screen.
> * It adds an additional window on the main monitor - to now dismiss the screenshot mode, you have to press esc -> click on the main monitor to focus -> press esc again.
> 
> Tested on MacBook Air M2 with 1 external monitor.

Issue #240, opened 2018-06-01
Option to load existing image and edit it.


deragon:
Flameshot does not allow loading of existing files like Shutter does.  It should allow to load an existing image so annotations can be added later, easily.  Please add a File / Snapshot - Open  - Save - Save as menu.

mbirth:
Please also let us load from clipboard in case we forgot to mark/blur something.

Migliutin:
Please add corresponding **terminal commands** as well, thus allowing to integrate Flameshot within **external tool** (e.g. **clipboard managers**).

lupoDharkael:
I've been secretly working on this feature but it still needs some work.
As you can see the CaptureWidget receives a bool "fullscreen" and when you set that value to false it creates an instance in window mode as it is required by this feature request.
https://github.com/lupoDharkael/flameshot/blob/415b059fb248eec4b4ee3be85a1594ce99bfce67/src/widgets/capture/capturewidget.cpp#L52

I'm sure the next version of Flameshot will have a mode to use it as an image editor.

RB3rg:
I can't wait on this;
Shutter was removed from Ubuntu 19.04, yeah, I know we can install from 3rd repo but it will be dead soon. 
Need a replacement urgently.
Great tool. Thanks.

RB3rg:
I can't wait on this;
Shutter was removed from Ubuntu 19.04, yeah, I know we can install from 3rd repo but it will be dead soon.
Need a replacement urgently.
Great tool. Thanks.

kramer65:
May I ask whether this feature is still in the works? I'm finding myself needing this almost every week. As a workaround I currently open the image and take a screenshot of the image to annotate it. It works, but it would be great if flameshot could do this.. :-)

infroger:
Really missing this feature. Shutter still unbeatable because you can easily edit images later.

zaidrpm:
Any Updates? I am eagerly waiting for this feature. It will be very useful for students making notes on their pc's as flameshot has got best annotation tools on linux. It would be great if you can set option to "darken only" for brush as it is very useful to highlight text.

pierluigiferrari:
Any updates? I'm also waiting on this feature :)

GrantEdwards:
I just installed flameshot hoping to use it to annotate some existing image files. I guess I'll have to try shutter, even though it's going to require 50+ other packages be installed as dependencies.

lkrms:
[...]

PR #3619, opened 2024-05-28
New Feature: now flameshot has ability to show screenshot history

Michaelzhouisnotwhite:
# Usage

use ', ' and '. ' to get history of screenshots:

- ',' for showing history backward
- '.' for showing history forward

# Details:

## My contribution
add a singleton class: CaptureHistoryUtils

history screenshots that user export are saved in `getCachePath() + /cap.his.<selection size> + <datetime>`

## ScreenShots

https://github.com/flameshot-org/flameshot/assets/52443370/453dc9c6-a1b9-4d1e-ba5a-d607a1a8ce05


src/widgets/capture/capturewidget.cpp
```diff
@@ -279,6 +279,9 @@ CaptureWidget::~CaptureWidget()
         geometry.setTopLeft(geometry.topLeft() + m_context.widgetOffset);
         Flameshot::instance()->exportCapture(
           pixmap(), geometry, m_context.request);
+
```
Michaelzhouisnotwhite, line 282: save screenshots to cache

src/widgets/capture/capturewidget.cpp
```diff
@@ -1575,6 +1581,9 @@ void CaptureWidget::initShortcuts()
                 SLOT(selectAll()));
 
     newShortcut(Qt::Key_Escape, this, SLOT(deleteToolWidgetOrClose()));
+
```
Michaelzhouisnotwhite, line 1584: add shotcuts

src/widgets/capture/capturewidget.h
```diff
@@ -157,6 +161,10 @@ private slots:
     // Context information
     CaptureContext m_context;
 
+    CaptureContext m_currentContext;
```
Michaelzhouisnotwhite, line 164: fields that save the current state

mmahmoudian:
Thanks for the PR. I truely appreciate it.

In general we would like to have a PR after the issue is already discussed in an issues or in our Matrix channel. This would help the maintainers to see if they think the feature is in-line with the perspective of the project, and also would help during the review process.

Apart from that, I'm extremely puzzled on what this feature does. I quickly checked the code and that didn't solve much of my confusion. Would you please elaborate on:
1. What is the use-case of this feature
2. How a user should use this

To me it seems you clearly have a need for such feature and you were kind enough to also push the feature to the main repo. I just don't understand what that need was.

mmahmoudian:
Thanks for the explanation. I can see that it can be useful for some users. I also like the name you chose for it (back-tracking).

In the light of [Microsoft Windows Recall](https://support.microsoft.com/en-us/windows/retrace-your-steps-with-recall-aa03f8a0-a78b-4b3e-b0a1-2eb8ac48701c) and the similarities to this back-tracking feature, I have a few concerns:

1. User should be able to toggle it on/off, and by default it should be off (such features with potential privacy implications should be Opt-in)
2. User should be able to define how much cache they would like to keep, and the rest should be cleaned up (I think number of screenshots is better than a Megabyte cutoff, so the default I guess should be a relatively small number like 5)
3. Would be super nice (end even essential) that the screenshots in the cache folder can be encrypted somehow and only decrypted when needed. This is something that I don't have a clear solution for it. For example:
    - Flameshot create a password-protected ed25519 pgp key pair for asymmetric encryption. The keys can be saved in the config folder of Flameshot
    - when screenshot is taken and this back-tracking feature is enabled, the public key is used to encrypt the screenshot and then saved to the cache folder. For this part, no input from user is needed as the public key does not need password to encrypt
    - when user wants to use the back-tracking to bring the previous screenshots from the cache, we can ask the user once for the password to decrypt the screenshots on the fly and show them
    - when user is done with back-tracking and closes the window of flameshot, then flameshot "forgets" the password

You can create a new tab in the config window (`flameshot config`) for back-tracking settings. There can be:
1. a checkbox to toggle this feature
2. a part for the user to define cache folder path
3. a button to generate the pgp key pairs for the user (if the key already exists, the generate button should be disabled and a text should be be added to indicate the existence of a key)

About the encryption part we need to discuss this with some more folks and get their opinion on the best practices, but for now if you can add the tab in the config to address my first two concerns, that would be great.

In the meantime, we can also wait for the following folks to weigh in and express their opinions:
@veracioux @panpuchkov  @jack9603301 @hosiet @holazt @lupoDharkael

Michaelzhouisnotwhite:
> Thanks for the explanation. I can see that it can be useful for some users. I also like the name you chose for it (back-tracking).
> 
> In the light of [Microsoft Windows Recall](https://support.microsoft.com/en-us/windows/retrace-your-steps-with-recall-aa03f8a0-a78b-4b3e-b0a1-2eb8ac48701c) and the similarities to this back-tracking feature, I have a few concerns:
> 
> 1. User should be able to toggle it on/off, and by default it should be off (such features with potential privacy implications should be Opt-in)
> 2. User should be able to define how much cache they would like to keep, and the rest should be cleaned up (I think number of screenshots is better than a Megabyte cutoff, so the default I guess should be a relatively small number like 5)
> 3. Would be super nice (end even essential) that the screenshots in the cache folder can be encrypted somehow and only decrypted when needed. This is something that I don't have a clear solution for it. For example:
>    
>    * Flameshot create a password-protected ed25519 pgp key pair for asymmetric encryption. The keys can be saved in the config folder of Flameshot
>    * when screenshot is taken and this back-tracking feature is enabled, the public key is used to encrypt the screenshot and then saved to the cache folder. For this part, no input from user is needed as the public key does not need password to encrypt
>    * when user wants to use the back-tracking to bring the previous screenshots from the cache, we can ask the user once for the password to decrypt the screenshots on the fly and show them
>    * when user is done with back-tracking and closes the window of flameshot, then flameshot "forgets" the password
> 
> You can create a new tab in the config window (`flameshot config`) for back-tracking settings. There can be:
> 
> 1. a checkbox to toggle this feature
> 2. a part for the user to define cache folder path
> 3. a button to generate the pgp key pairs for the user (if the key already exists, the generate button should be disabled and a text should be be added to indicate the existence of a key)
> 
> About the encryption part we need to discuss this with some more folks and get their opinion on the best practices, but for now if you can add the tab in the config to address my first two concerns, that would be great.
> 
> In the meantime, we can also wait for the following folks to weigh in and express their opinions: @veracioux @panpuchkov @jack9603301 @hosiet @holazt @lupoDharkael

Thanks for your suggestions. In the near future，I will complete the following needs：
1. a checkbox to toggle this feature
2. a part for the user to define cache folder path
3. a spinbox to define a number of screenshot need to keep in cache.

**footnote：** this 'back-tracking' idea comes from anothor closed-source screenshot application named `Snipaste`. However, the `Snipaste` only avaliable on windows. `Flameshot` is a great app and opensource, so I want the `flameshot` has this feature as well. So this pr came up.

src/utils/capturehistoryutils.cpp
```diff
@@ -0,0 +1,266 @@
...
+                }
+                selParseState = PS_SUCCESS;
+                break;
+            }
+            default:
+                break;
+        }
+    }
+    if (selParseState == PS_ERROR) {
+        AbstractLogger::error() << "cache pic selection parse errror";
```
mmahmoudian, line 192: `errror` -> `error` ? (unless it is intentional)

mmahmoudian:
@Michaelzhouisnotwhite Thank you for your hard work. I highly appreciate it.

May I suggest going through the code and use a unified naming scheme for this feature? I put [a small comment](https://github.com/flameshot-org/flameshot/commit/53bdb9798f21d23078faee0f9bebbf2a69e2a7c0#r142799272) on one of your commits, but now that I went through the whole PR again, I realized that this is potentially a wider concern.

In the code we have these diverse terminologies which are more or less pointing to the same concept
- "backtrack[er]" (sometimes "backTrack")
    - E.g. the file name `src/config/backtrackingimpl.cpp`
- "rollback" (sometimes "rollBack")
    - E.g. inside the file `src/config/backtrackingimpl.cpp`
- "btk"
    - E.g.  inside the file `src/config/backtrackingimpl.cpp`
    - E.g.  inside `src/config/backtrackingimpl.h`
- "screen history"
    - E.g.  inside `src/widgets/capture/capturewidget.cpp`
- "History"
    - E.g.  inside `src/widgets/capture/capturewidget.h` lines 78-81
- "capturehistory"
    - E.g.  file name of `src/utils/capturehistoryutils.cpp`
        - inside `src/utils/CMakeLists.txt`

I Understand that they are not all identical, but I suggest trying to unifying the terminology to make the maintenance of the project easier in the future. I believe the "backtrack" is a very fitting name and less confusing comparing to "rollback" or "history", because these pictures (as far as I understand them) are raw PNGs (no annotation, no cropping...).

Anyways, I feel this is a nice PR. Thank you again.

stuart938503:
While this PR wouldn't meet the original request in #240, it is related (IMO) and helps address the essence of the problem in that issue - editing screenshots after the capture window is closed.

Issue #2612, opened 2022-05-19
Flameshot zooms in & can't handle multiple screens on mac m1.


nohehf:
### Flameshot Version

Flameshot v11.0.0 (ad1bf28)
Compiled with Qt 5.15.2
darwin: 21.1.0
osx: 12.0

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

macOS Monterey v12.0.1, Apple M1 Pro

### Description

When trying to take a screenshot, flameshot insanely zooms in, leaving only the top-right part of the screen accessible to screen shot. I'm on google chrome, and have multiple screens. Also I can only screen shot the main screen.
 

### Steps to reproduce

1. Start flameshot screenshot onto google chrome

### Screenshots or screen recordings

https://user-images.githubusercontent.com/64989428/169247415-e2fffc09-8233-47d2-93a6-627cde732efc.mov

### System Information

<img width="635" alt="Screenshot 2022-05-19 at 10 19 47" src="https://user-images.githubusercontent.com/64989428/169247322-1780520f-575a-4dd4-becc-14059e119245.png">
<img width="447" alt="Screenshot 2022-05-19 at 10 20 26" src="https://user-images.githubusercontent.com/64989428/169247329-13766410-bd08-4918-8cc2-9ffa36d9979d.png">

nohehf:
You're right, thought I had already done it my bad. Tho I still have always a little zoom in, which is pretty annoying (see video, how the graph at the bottom gets cropped).

https://user-images.githubusercontent.com/64989428/169377997-37609191-9750-42ba-9c2a-531ef478fbb0.mov

mmahmoudian:
@nohehf  As far as I can see, this is not a zoom effect, but there is downwards shift of the image. Note the additional black region on top of the image. How does you monitor setup look like when you face this issue?

nohehf:
Right, it's shifting, and I'm only on my laptop screen.

borgmanJeremy:
The other issue specifically pointed out that this is on a screen with a notch. I wonder if its related to that and Qt is using the region below the notch to draw full screen instead of the region above the notch. I have no way to test / work on this since I have an air.

jasonycw:
I have been experiencing the exact same issue on the macbook with notch. I think it's due to the way MacOS handling the notch with a "full screen application"

[...]

PR #3625, opened 2024-06-05
fix: workaround macos fullscreen animation

### LINKED ISSUE #2886: Active desktop switches to Desktop 1 after saving/copying a screenshot on MacOS
### Flameshot Version

v12.1.0 (96c2c82e)

### Installation Type

macOS Monterey (Version 12.5.1)

### Description

The screenshot function works perfectly fine. All the modifications (edits) I make to the screenshot are all there but immediately when I save/copy it I find myself on `Desktop1` instead of the fullscreen window I took the screenshot from. I have to swipe my way back to desktop I was on first.

### Steps to reproduce

[ only tested this on macOS Monterey 12.5.1 ]

1. Open some app and make it fullscreen
2. Invoke `flameshot` through the keyboard shortcut or from the icon on the menubar
3. Select the area for a screenshot by dragging the box or resizing it (make sure you're on the desktop that has the fullscreen app)
4. Click on the save or copy icon

At this point, you should have the screenshot saved to disk or stored in the clipboard but the active desktop switched to `Desktop1`.

bug

### System Information

**Operating system and version**
macOS Monterey 12.5.1

image

image

paperclover:
**edit**: do not think this is viable due to https://github.com/flameshot-org/flameshot/issues/2886#issuecomment-2205197790

This would resolve #2886 at the cost of the system menu bar not being able to be visible as part of the capture window.

image

If there is something available in QT to put the window above the menu bar or disable the animation, we should use that instead. I couldn't find an API to do that though -- I feel like i0t has to be possible *somehow*, since the built-in screenshot command does not have this issue, and System Settings is able to draw overlays that are placed on top of the menu bar (if you have multiple monitors and re-arrange them, a red border surrounds the screen); Maybe this API is reserved for the system apps?

For now, I find this trade off reasonable as the fullscreen exit animation as shown in the issue is incredibly annoying, and most screenshots do not need to annotate the menu bar (you can still screenshot it, but the system bar is in the way for annotations).

### DIFF
--- a/src/core/flameshot.cpp
+++ b/src/core/flameshot.cpp
@@ -120,8 +120,9 @@ CaptureWidget* Flameshot::gui(const CaptureRequest& req)
 #ifdef Q_OS_WIN
         m_captureWindow->show();
 #elif defined(Q_OS_MACOS)
-        // In "Emulate fullscreen mode"
-        m_captureWindow->showFullScreen();
+        // Cannot use showFullScreen as macOS plays an animation on open and
+        // close which does not seem to be removable.
+        m_captureWindow->show();
         m_captureWindow->activateWindow();
         m_captureWindow->raise();
 #else
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -39,7 +39,6 @@
 #include <QPainter>
 #include <QScreen>
 #include <QShortcut>
-#include <QtCore/qnamespace.h>
 #include <draggablewidgetmaker.h>
 
 #if !defined(DISABLE_UPDATE_CHECKER)
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -39,6 +39,7 @@
 #include <QPainter>
 #include <QScreen>
 #include <QShortcut>
+#include <QtCore/qnamespace.h>
 #include <draggablewidgetmaker.h>
 
 #if !defined(DISABLE_UPDATE_CHECKER)
@@ -55,7 +56,11 @@
 CaptureWidget::CaptureWidget(const CaptureRequest& req,
                              bool fullScreen,
                              QWidget* parent)
-  : QWidget(parent)
+  : QWidget(parent
+#if defined(Q_OS_MACOS)
+  , Qt::FramelessWindowHint | Qt::WindowStaysOnTopHint
+#endif
+  )
   , m_toolSizeByKeyboard(0)
   , m_mouseIsClicked(false)
   , m_captureDone(false)


mmahmoudian:
Do I understand it correctly that this PR makes Flameshot to be a "maximized" window but not a full-screen one?

Untersander:
I believe this would also fix the issue from #2612 as not capturing the menu bar, for me it would be worth to lose the feature to capture the menu bar but if anybody has a solution how this could be fixed without losing this feature would be nice.

ElijahLynn:
Thanks for the initiative!

I would like to suggestions from the community on what other OSS screenshot tooling is there that does this correctly? Maybe then we could see what API or techniques they use.

paperdave:
from https://github.com/flameshot-org/flameshot/issues/2886#issuecomment-2205197790, it seems this approach is not going to be viable at all. we'll need a different technique. not sure what though.

jackielii:
I tried this patch, it doesn't work in my case. I can't see any window: as soon as I activate capture, the applications disappear.

paperdave:
> I tried this patch, it doesn't work in my case. I can't see any window: as soon as I activate capture, the applications disappear.

flameshot or the program executing flameshot doesn't have screen capture permissions. not a bug with this patch since i ran into this while building and running master.

paperdave:
just tried cleanshot and they are currently the only tool other than the built in one that ive seen that doesnt have this issue.

so it's possible. just no idea how. might have to use more direct system apis to do this

ElijahLynn:
@chrismessina is the creator of Cleanshot, not sure he would be willing to help us since but @ing him just in case!

chrismessina:
> @chrismessina is the creator of Cleanshot, not sure he would be willing to help us since but mentioning him just in case he is! Chris, would you be up for sharing how you implemented screenshotting a fullscreen desktop so that it keeps you on the desktop? Was it just a standard API call, or did you have to do something hacky?

Hey @ElijahLynn — I'm an avid user of Cleanshot, but certainly not its creator! 

You'll want to contact [Luke](https://t.me/lukaszmtw) or Paul from MakeTheWeb. 

Here's their [URL scheme docs](https://cleanshot.com/docs-api), in case they're useful.

ElijahLynn:
> Hey @ElijahLynn — I'm an avid user of Cleanshot, but certainly not its creator!

DOH! Wonder how I arrived at that conclusion. Thanks for pointing out who to contact!

dfredell:
Thank you @paperdave I just switched from Linux to Mac. In the GA mac version the desktop flipping after the capture was complete was annoying. This MR solved that for me. Thanks.

paperdave:
> might have to use more direct system apis to do this

random conversation with the creator of ghostty (terminal application) led me to see that its "Quick terminal" feature was able to paint above the menu bar and on multiple screens.

the code is currently private but that feature is implemented in Swift (couple hundred lines of code). so the proper version of this fix would require some swift code to be added to the build system, but in return it would become the best free screenshot tool for the Mac.

mmahmoudian:
Then we need some help from community to help us with swift. As far as I know, none of the devs are mac yser and none of them are swift developrs.

icaroscherma:
I don't mind not being able to screenshot the menubar if I get the Desktop Switching fix.

As it's a compromise-able change and we could find people wanting either of the "sides", I think a toggle switch on the settings, maybe requiring it to restart the App when changed might be a feasible option.

I really like the idea of staying with the same app regardless of the OS as most of the shortcuts are now pretty much muscle memory to me.

PR #3640, opened 2024-06-14
Added a way to install the program in ALT Linux

Krator3:


mmahmoudian:
Thanks for the PR, but I wonder if such addition is necessary. In this README the idea was to provide an example of different package managers, and also cover some of the most commonly used distros (i.e Ubuntu). ALT is using apt as package manage (although being RPM-based). I believe is someone is using ALT, they already know how to install software using apt. I'm not extremely against this addition, I'm just speculating that if it is even necessary.

As a side note, I was surprised to see that ALT is not listen in repology:

https://repology.org/project/flameshot/versions

Perhaps would be nice to have repology list the ALT repo too.

Krator3:
Hello! Thanks for your reply. The site you indicated has ALT systems. In addition, ALT has the peculiarity that it cannot work with apt, but only with apt-get. Besides In addition, this distribution has little popularity in the community, but at the same time it is a very cool system. Why not help make it more meaningful?

mmahmoudian:
> The site you indicated has ALT systems.

You are right, somehow I missed it!

> Why not help make it more meaningful?

Although I don't think adding it to a readme of a screenshot tool would influence popularity of a distro, for sure it doesn't hurt. But before merging are you sure using `su` is the correct way? Don't ALT has any privilege escalation tool like `sudo` or `doas`?

Additionally, would you like adding it to our website as well?

https://github.com/flameshot-org/flameshot-org.github.io/blob/master/content/docs/installation/installation-linux.md

Krator3:
Hello! Yes, I’m sure that you need to use su, since by default ALT has sudo, but it is disabled for everyone, and many users fundamentally do not want to enable it. About the site... I wouldn't mind having information about ALT in as many places as possible, but I don't know how to add information to the site. I would be very glad if you could help me.

Krator3:
I'm sorry, but will you accept my PR or is there any problem?

Krator3:
Hello! Have you forgotten about me?)

mmahmoudian:
Hi, no, but this is not my full-time job either ;)

mmahmoudian (approved):

mmahmoudian:
Thanks for your contribution 👍

Krator3:
Thanks!

PR #3650, opened 2024-06-28
removes Slack from documentation

mmahmoudian:
We used to use Slack for organizing Flameshot, but later we moved to Matrix, and the Slack was abandoned. Last year I put notice on Slack to inform users that the Slack will not be checked and used anymore. Therefore, we need to reflect that in our documentations.


FelixJochems (approved):
changes that dont impact code flow, you can just merge them directly

mmahmoudian:
I can, but based on the policy we had, we configured the repo so that the person who submit the PR cannot merge it themselves. This is to have an extra pair of eyes on changes.

PR #3651, opened 2024-06-28
exclude documentation modifications from CI

mmahmoudian:
This is to prevent CI to run when we only change the documentation of the project. CI's are computationally expensive and time consuming and therefore we can save some time and some electricity by running the CI when they are actually needed.


FelixJochems (approved):
LGTN

PR #3672, opened 2024-07-19
Adds WebP Quality Config

PoneyClairDeLune:
Like #3285, this PR aims at introducing a quality config into Flameshot, allowing users to customize the quality of the WebP image emitted by Flameshot.

However during the testing of this PR, I've noticed that the output of the final WebP image appears to be far inferior than that emitted by `cwebp` with the same quality setting, and `cwebp` can excel at customizing other parameters like effort, increasing the final quality without increasing the final file size. So perhaps there should be a config option, allowing Flameshot to automatically execute a command to optimize the final image from PNG.

### DIFF
--- a/flameshot.example.ini
+++ b/flameshot.example.ini
@@ -94,7 +94,7 @@
 ; jpegQuality=75
 ;
 ;; Set WebP Quality (int in range 0-100)
-; webpQuality=95
+; webpQuality=90
 ;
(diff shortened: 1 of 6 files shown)


mmahmoudian:
@jack9603301 What is your opinion about this question?

borgmanJeremy:
Since we are on Qt6 now, it would be good to see if this is still needed.

PR #3674, opened 2024-07-25
Added river support

cvoges12:
This attempts to generalize wlroots compositors by checking against the proper desktop name instead of only sway. This fixes the problem of only supporting sway. Any additional desktop will just have to be added to the source to check for the correct string.

### DIFF
--- a/docs/Sway and wlroots support.md
+++ b/docs/Sway and wlroots support.md
@@ -4,7 +4,7 @@ Flameshot currently supports Sway and other wlroots based Wayland compositors th
 ## Basic steps
 The following packages need to be installed: `xdg-desktop-portal xdg-desktop-portal-wlr grim`. Please ensure your distro packages these, or install them manually.
 
-Ensure that environment variables are set properly. If your distro does not set them automatically, use a launch script to export `XDG_CURRENT_DESKTOP=sway` **before** Sway is launched.
+Ensure that environment variables are set properly. If your distro does not set them automatically, use a launch script to export `XDG_CURRENT_DESKTOP=sway` or `XDG_CURRENT_DESKTOP=river` **before** Sway or River is launched.
 ```sh
 #!/bin/bash
 export SDL_VIDEODRIVER=wayland
@@ -15,7 +15,20 @@ export XDG_SESSION_DESKTOP=sway
 exec sway
 ```
 
-You will also need to ensure that systemd/dbus is aware of these environment variables; this should be done **in your sway config** so that the DISPLAY and WAYLAND_DISPLAY variables are defined.
+or
+
+
+```sh
+#!/bin/bash
+export SDL_VIDEODRIVER=wayland
+export _JAVA_AWT_WM_NONREPARENTING=1
+export QT_QPA_PLATFORM=wayland
+export XDG_CURRENT_DESKTOP=river
+export XDG_SESSION_DESKTOP=river
+exec river
+```
+
+You will also need to ensure that systemd/dbus is aware of these environment variables; this should be done **in your sway or river config** so that the DISPLAY and WAYLAND_DISPLAY variables are defined.
 
 (taken from [Sway wiki](https://github.com/swaywm/sway/wiki#gtk-applications-take-20-seconds-to-start)):
 ```sh
@@ -29,6 +42,13 @@ To ensure that Flameshot is correctly positioned on multiple outputs (monitors) 
 for_window [app_id="flameshot"] border pixel 0, floating enable, fullscreen disable, move absolute position 0 0
 ```
 
+and add the following on your River config:
+
+```
+riverctl float-filter-add "flameshot"
+```
+
+Otherwise, flameshot will not take all of the screen and tiles its window instead like a normal application. Note however, that some clipboard stuff is broken so it might be good to save your screenshot as a file while having it copied to a clipboard in case if clipboard does some weird stuff like not pasting the overall screenshot.
 
 
 Starting from 0.17.0 xdg-desktop-portal requires a configuration file (e.g. in ~/.config/xdg-desktop-portal/sway-portals.conf):
@@ -54,21 +74,6 @@ Q) Flameshot takes one screenshot, then won't take anymore!
 
 A) There is a bug in xdg-desktop-portal-wlr and Flameshot causing calls with the same token to fail. If you see a sdbus vtable error in the xdpw logs, either used the [patched version](https://github.com/nullobsi/xdg-desktop-portal-wlr/tree/improve-screenshot) or update Flameshot to the latest master.
 
-# River wlroots support
-
-Like mentioned above, flameshot now works on wlroots based Wayland compositors, however, there is a weird problem with river and that is when setting `XDG_CURRENT_DESKTOP=river`, flameshot won't work. The fix is you need to trick flameshot that you are on `sway`. Hence, you need to run river like so:
-
-```sh
-XDG_CURRENT_DESKTOP=sway dbus-run-session river
-```
-
-and add the following on your config such as in `$HOME/.config/river/init`
-
-```
-riverctl float-filter-add "flameshot"
-```
-
-Otherwise, flameshot will not take all of the screen and tiles its window instead like a normal application. Note however, that some clipboard stuff is broken so it might be good to save your screenshot as a file while having it copied to a clipboard in case if clipboard does some weird stuff like not pasting the overall screenshot.
 
 #### For more information, please refer to https://github.com/emersion/xdg-desktop-portal-wlr/wiki/%22It-doesn't-work%22-Troubleshooting-Checklist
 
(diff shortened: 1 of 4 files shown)


docs/Sway and wlroots support.md
```diff
@@ -29,6 +42,13 @@ To ensure that Flameshot is correctly positioned on multiple outputs (monitors)
 for_window [app_id="flameshot"] border pixel 0, floating enable, fullscreen disable, move absolute position 0 0
 ```
 
+and add the following on your River config:
+
+```
+riverctl float-filter-add "flameshot"
```
Sunderland93, line 48: This rule is not support since River 0.3.x. You should use `riverctl rule-add -app-id "flameshot"` float instead

cvoges12: Okay, I'll change this

cvoges12: Changed :+1:

TAforever:
Maybe I don’t understand something, but river is already supported in the master branch and you no longer need to indicate that the current desktop is sway

midirhee12:
> Maybe I don’t understand something, but river is already supported in the master branch and you no longer need to indicate that the current desktop is sway

Changing your XDG vars to sway is not a solution. It is a very hacky work around. It breaks all other software using XDG vars. So explicit support is required instead of the hack which the docs recommend.

TAforever:
You don’t understand, you no longer need to indicate that the current desktop is sway. I have` XDG_CURRENT_DESKTOP=river` set on river and `XDG_CURRENT_DESKTOP=Hyprland` on hyprland

midirhee12:
> You don’t understand, you no longer need to indicate that the current desktop is sway. I have` XDG_CURRENT_DESKTOP=river` set on river and `XDG_CURRENT_DESKTOP=Hyprland` on hyprland

Can you link the commit that solved this?

TAforever:
Maybe https://github.com/flameshot-org/flameshot/commit/3ededae5745761d23907d65bbaebb283f6f8e3f2 not sure but you can build the master branch and make sure everything works

midirhee12:
@TAforever 

1. This only adds Hyprland support which no longer uses the wlroots portals. This issue is for river support and wlroots compositors in general through the future thus completely off topic.
2. I'm starting to seriously doubt that you actually tested this on both Hyprland AND river. It sounds like you only tried Hyprland and assumed it works for all. Don't tell others to build against master on the exact target when they have and you obviously haven't. 
3. The code in its current state (both v12.1.0 release and master) matches for gnome, qtile, sway, hyprland, and KDE [here](https://github.com/flameshot-org/flameshot/blob/master/src/utils/desktopinfo.cpp#L31) and [here](https://github.com/flameshot-org/flameshot/blob/master/src/utils/screengrabber.cpp#L137). So unless your compiler is magic, everyone else's compiler is somehow defunk, the docs are wrong, and/or our complete understanding of how a switch-case statement works is defunk, you're wrong or lying. 

Please find some other person to troll when you're too lazy and lack the time to actually perform a proper system test.

TAforever:
It was very strange to read this. If you don’t believe me (although why should I deceive you), then trust yourself simply by checking. Good luck

midirhee12:
You may have `XDG_CURRENT_DESKTOP=Hyprland` still set for river which because of the xdg desktop hyprland portal being a fork of the wlroots portal may somehow still be backwards compatible.  But this again is another workaround and will probably not work in the foreseeable future. Also, you may not be exposing your XDG vars to dbus.

> then trust yourself simply by checking

I have. I don't make PRs or comment on issues or PRs without checking like you.

TAforever:
https://github.com/user-attachments/assets/62d218f5-500a-4a94-b09b-4c89206cdeb6

midirhee12:
What are you running for `dbus-update-activation-environment`?

TAforever:
In init I have `dbus-update-activation-environment --systemd WAYLAND_DISPLAY`

midirhee12:
Then that's probably why. You haven't properly told dbus your `XDG_CURRENT_DESKTOP` which some but not all programs require.

midirhee12:
Also, how are you launching flameshot? I see no flameshot command executed on the terminal. So this is really deceiving.

TAforever:
I don't know how to comment on this. Goodbye

midirhee12:
Well, you could show how you're running flameshot or even send another video of actually running flameshot from the terminal. But the fact that you're only telling dbus that you're on Wayland and not which compositor already throws a red flag.

mmahmoudian:
Apart from the discussion, and despite of not being a river user myself, the PR's code change looks good to me. let's wait for @veracioux, @panpuchkov, @jack9603301, or other devs review it too before merging.

As a side note, imho this `if(...) || (...) || ... ` situation is not scalable for wlroots. There must be a better way and more universal approach to this.

midirhee12:
> situation is not scalable for wlroots

I 100% agree. But the only other thing I can think of to find if something uses wlroots is to ask if the xdg-desktop-portal-wlr is installed which is not always the case. So to be consistent with the other  code and to be conforming with every wlroots system, this is probably the best choice *for now*.

midirhee12:
I mean, I guess the xdg-desktop-portal is kind of required for a screenshot tool to be fair. So it may be possible to test based on this?

mmahmoudian:
> So it may be possible to test based on this?

Perhaps. I think for now we go with this and keep an open mind on alternative approaches

midirhee12:
I'll play around with the xdg-desktop-portal in the meantime to see if I can figure out something. I think it'll be a lot of research though. So I don't anticipate having a PR ready any time soon. So yeah, I agree. I think merging this in the meantime is the best idea.

midirhee12:
Any further thoughts? Or is there anything keeping from merging this?

0xmayro:
Hello, not sure if it's helpful, 
this PR worked for me on Arch with the river compositor.

midirhee12:
Yeah, I'm not sure what the hold up is. It compiles. It works. I'm willing to test for river in the future. And no is merging it. Absolutely nuts.

borgmanJeremy:
I'm reviewing this now for merging. I usually like to test stuff before merging, but I'll be honest the river config was a bit beyond me so I'll probably tag you for any River questions if thats okay.

Would you be able to rebase this on the master branch to trigger the CI and resolve merge conflicts? I suspect clang-format will fail but after CI is passing I'll merge.

midirhee12:
@borgmanJeremy Ready to merge.

docs/Sway and wlroots support.md
```diff
@@ -29,6 +42,13 @@ To ensure that Flameshot is correctly positioned on multiple outputs (monitors)
 for_window [app_id="flameshot"] border pixel 0, floating enable, fullscreen disable, move absolute position 0 0
 ```
 
+and add the following on your River config:
+
+```
+riverctl rule-add -app-id "flameshot" float
+```
+
+Otherwise, flameshot will not take all of the screen and tiles its window instead like a normal application. Note however, that some clipboard stuff is broken so it might be good to save your screenshot as a file while having it copied to a clipboard in case if clipboard does some weird stuff like not pasting the overall screenshot.
```
borgmanJeremy, line 51: Did you build flameshot with WAYLAND_CLIPBOARD=true and still have clipboard issues?

src/utils/desktopinfo.cpp
```diff
@@ -34,8 +34,8 @@ DesktopInfo::WM DesktopInfo::windowManager()
         if (desktop.contains(QLatin1String("qtile"), Qt::CaseInsensitive)) {
             return DesktopInfo::QTILE;
         }
-        if (desktop.contains(QLatin1String("sway"), Qt::CaseInsensitive)) {
-            return DesktopInfo::SWAY;
+        if (desktop.contains(QLatin1String("sway"), Qt::CaseInsensitive) || desktop.contains(QLatin1String("river"), Qt::CaseInsensitive)) {
```
borgmanJeremy, line 37: the wrapping on this needs to change to satisfy clang-format

borgmanJeremy (changes requested):

docs/Sway and wlroots support.md
midirhee12: Oh, I just realized that part is old and no longer relevant. Let me fix that.

midirhee12:
@borgmanJeremy I think I resolved all requested changes. Is there anything else blocking?

borgmanJeremy:
clang-format is still unhappy but I know it can be a PITA. If its too annoying I can fix it after merging. I usually run this to automatically format the code: 
```
find src/ -iname '*.h' -o -iname '*.cpp' | xargs clang-format -i
```

midirhee12:
@borgmanJeremy Okay, thanks for the tip. I managed to run `clang-format` from the `clang-tools` package. It should all be good now.

borgmanJeremy:
Thank you!

PR #3683, opened 2024-07-31
Remove QT+GNOME+Wayland 'xcb' hack on fixed Qt versions

danudey:
We implement code on GNOME desktops to force `QT_QPA_PLATFORM` to `xcb`; this works around a clipboard-related bug on GNOME+Wayland+Qt.

This bug was fixed (or worked around) in Qt 5.15.2, so we implement a version check; if the runtime Qt version is < 5.15.2, still force the workaround; otherwise, we don't need the workaround so we skip it.

We also add logging output so that users attempting to force the app not to use xcb won't get confused (as I did).

Note that the original discussion on this issue was in ksnip/ksnip#416, which indicates when this issue was fixed in Qt.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -37,10 +37,23 @@
 // source: https://github.com/ksnip/ksnip/issues/416
 void wayland_hacks()
 {
-    // Workaround to https://github.com/ksnip/ksnip/issues/416
+    int suffixIndex;
     DesktopInfo info;
-    if (info.windowManager() == DesktopInfo::GNOME) {
-        qputenv("QT_QPA_PLATFORM", "xcb");
+
+    const char *qt_version = qVersion();
+
+    QVersionNumber targetVersion(5, 15, 2);
+    QString string(qt_version);
+    QVersionNumber currentVersion = QVersionNumber::fromString(string, &suffixIndex);
+
+    if (currentVersion < targetVersion) {
+        if (info.windowManager() == DesktopInfo::GNOME) {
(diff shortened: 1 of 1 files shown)


danudey:
Is @mmahmoudian or anyone else available to review this PR/provide feedback on it?

mmahmoudian:
Thanks, the code looks clean and make sense to me and I believe it is small enough that I can merge it without waiting for other devs.

I just had a small question about a defined constant string, that I'll be grateful if you can provide some insight.

Thanks.

mmahmoudian (approved):

mmahmoudian:
I'm so confused on why GithubActions are not running for this PR! I'm also merging #3725 and that one is running all the CIs. I need to make sure all CIs pass before merging.

mmahmoudian (approved):

danudey:
@mmahmoudian I added another commit (with clang-format updates), and it says "This workflow requires approval from a maintainer", and links to [this document](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks). Perhaps that's what's missing?

mmahmoudian (approved):

danudey:
Thanks for taking a look! Hope this PR can get merged soon and we can have one more Wayland-native app.

jack9603301:
Looks good

borgmanJeremy:
Since this is stale from us I'm going to merge this and if it fails CI I'll post a fix for it. I think to trigger the now fixed CI you would need to rebase.

PR #3699, opened 2024-08-26
Adds Support for Command-Line on Windows

### LINKED ISSUE #2118: Implement the same CLI args in Windows build

niwamo:
## Description

Adds support for command-line usage / command-line arguments on Windows.

Closes #2118

## Approach

I assume this wasn't done previously because Windows apps can be console apps or windows apps, but not both. That leaves two options with a single flameshot executable:

1. A GUI app that respects CLI arguments but never returns any text to console
2. A console app that always pops up a console window, even when started with a double-click

There are some workarounds to mitigate the inconveniences of these approaches, but ultimately, the best solution is two executables.

*Good discussions on this topic can be found on Stack Overflow [here](https://stackoverflow.com/questions/15952892/using-the-console-in-a-gui-app-in-windows-only-if-its-run-from-a-console) and [here](https://stackoverflow.com/questions/493536/can-one-executable-be-both-a-console-and-gui-application).*

This commit adds the windows-cli.cpp source, which compiles into flameshot-cli.exe when built on Windows. It is a minimal wrapper around flameshot.exe but ensures that all stdout is captured and output to the console.

## Code Changes

- The preprocessor macro that prevented arg parsing on Windows has been removed from main.cpp
- windows-cli.cpp has been added as the source for flameshot-cli.exe
- The flameshot-cli target has been added to cmake (only when building on Windows)
- README updates

## README Updates

> ### Usage on Windows
>
> On Windows, `flameshot.exe` will behave as expected for all supported command-line arguments, but it will not output any text to the console. This is problematic if, for example, you are running `flameshot.exe -h`.
> 
> If you require console output, run `flameshot-cli.exe` instead.


niwamo:
@mmahmoudian hi, looks like you're the most active maintainer... would you mind giving this a review?

RivenSkaye:
Did you test if this plays nice in unicode-enabled environments such as Windows Terminal? I have no idea _what_ character sets it passes around, but considering it allows the use of e.g. emoji on the CLI (2+ characters in UTF-8, 1+ on UTF-16) that might be worth checking. I'm asking because CLI invocations can do a lot of stuff (like naming files with disallowed characters) that are caught in graphical environments

mmahmoudian:
@niwamo sorry, this last comment of yours fell into the cracks and I missed it.

My role in this project is mainly community management, triage, and documentation as I'm not as experienced as other devs in C++. I will review to the best of my abilities, but based on an old agreement, I refrain to merge large codes that I might not fully understand the caveats.

Nonetheless, I will review it when I get behind a computer. 🙂

Regardless of the results of the review, I would like to thank you for investing your time and effort into this community project.

mmahmoudian:
> Windows apps can be console apps or windows apps, but not both. [...] This commit adds the windows-cli.cpp source, which compiles into flameshot-cli.exe when built on Windows. It is a minimal wrapper around flameshot.exe but ensures that all stdout is captured and output to the console.

I have no solid understanding on C++ development on Windows, so I cannot provide good criticism. From my very limited understanding on Windows C++, and based on you clear explanation and the sources you provided, I believe this wrapper is a good compromise.

@jack9603301 @panpuchkov @veracioux what do you think about this approach and this PR.

README.md
```diff
@@ -152,6 +152,14 @@ A systray icon will be in your system's panel while Flameshot is running.
 Do a right click on the tray icon and you'll see some menu items to open the configuration window and the information window.
 Check out the About window to see all available shortcuts in the graphical capture mode.
 
+### Usage on Windows
+
+On Windows, `flameshot.exe` will behave as expected for all supported command-line arguments, 
+but it will not output any text to the console. This is problematic if, for example, you are 
+running `flameshot.exe -h`.
+
+If you require console output, run `flameshot-cli.exe` instead.
```
mmahmoudian, line 161: Please add this text to the end of the sentence for extra transparency:

> is a minimal wrapper around flameshot.exe but ensures that all stdout is captured and output to the console.

This sentence is borrowed from yourself from the PR's description text :)

mmahmoudian (changes requested):
@niwamo 
It looks clean and simple that I understand the code easily. Thank you. Just one small comment in the README file for extra clarity for users who don't want to inspect the code. After this change, I believe this PR should be ready, and I think it is small and benign enough that I can merge it myself if other devs do not object in the next 30 days.

Thank you again for the PR.

**EDIT:** I think the @RivenSkaye question is valid. Can you test it with some unicode characters and (e.g emoji) for instance in the filename and see if it behaves nicely? e.g `flameshot-cli.exe gui --path '/tmp/in-the-spirit-of-🎃.png'`

niwamo:
Thanks for the reviews and comments! I had not considered the possibility of emoji characters in CLI args. Right now, they result in silent errors. I will make updates within the next week or so.

niwamo:
Just added a commit with:
- support for Unicode characters in CLI args
- the requested additional clarification in the README
- a new workaround for space-containing paths in _popen (the new workaround ensures expected behavior when using relative file paths for output files)

image

src/windows-cli.cpp
```diff
@@ -0,0 +1,62 @@
...
+    if (argc == 1) {
+        std::cout << "Starting flameshot in daemon mode" << std::endl;
+        CallFlameshot(L"", false);
+    } else {
+        std::wstring argString = joinArgs(argc, argv);
+        CallFlameshot(argString, true);
+    }
+    std::cout.flush();
+    return 0;
+}
```
RivenSkaye, line 62: I get that Windows isn't as strict as POSIX with its definition of a line or a textfile, but cross toolchains would be technically allowed to toss this file out due to the missing EOL

niwamo:
Not sure how I (or the linter) missed the line ending. It's fixed.

src/windows-cli.cpp
```diff
@@ -0,0 +1,62 @@
...
+    }
+    return result;
+}
+
+void CallFlameshot(const std::wstring args, bool wait)
+{
+    // generate full path for flameshot executable
+    wchar_t path[MAX_PATH];
+    int pathLength = GetModuleFileNameW(NULL, path, MAX_PATH);
+    std::wstring moduleDir = std::wstring(path, pathLength - 18);
```
borgmanJeremy, line 21: 18 = length of flameshot-cli.exe + null terminator?

If yes could you make a string of "flameshot-cli.exe" and calculate 18 from that? Then it will be more clear if there is a rename / bug here in the future.

borgmanJeremy (changes requested):

src/windows-cli.cpp
niwamo: New approach finds the right-most backslash and uses everything to the left as the directory name. Take a look and see if that works for you

niwamo:
Thanks for following up on this. I've rebased on the latest master commits, adjusted the flameshot.exe path construction, and verified locally that the binaries still build and function properly on Windows. 

(You'll see a force push over my own commits  - at some point last year, I had somehow incorporated unrelated README changes from another PR)

PR #3710, opened 2024-08-31
Added DD-MM-YYYY format option button in configuration menu.

YashKale313:
Added DD-MM-YYYY format option button in configuration menu.

### DIFF
--- a/src/config/strftimechooserwidget.cpp
+++ b/src/config/strftimechooserwidget.cpp
@@ -67,4 +67,6 @@ QMap<QString, QString> StrftimeChooserWidget::m_buttonData{
     { QT_TR_NOOP("Full Date (%m/%d/%y)"), "%D" },
 #endif
     { QT_TR_NOOP("Full Date (%Y-%m-%d)"), "%F" },
+    { QT_TR_NOOP("Full Date (%d-%m-%Y)"), "%d-%m-%Y" },
+
 };


mmahmoudian:
Thanks. Since your proposal was to have even number of options, would you please also add the century (or something similar).

YashKale313:
@mmahmoudian Actually the logic that add format buttons ensures that there are even number of formats passed. So before threre were 21 formats (including century) and it was only adding 20 formats excluding century but now since i added 1 it has become 22 formats and its showing all 22 as its even.

YashKale313:
@mmahmoudian Here is the code code snippets for your refrence 

For adding formats-
image

For adding buttons-
image

let me know if you feels anything wrong here.

YashKale313:
Hi @mmahmoudian Just checking in if you have reviewd the PR and can it be merged.

b0ch3nski:
I've build with this patch (it applied clearly) and indeed the new option is available in the config UI

borgmanJeremy:
I no longer have a Windows machine to test on. Does anyone know if this format is compatible on Windows or if it needs to move to an ifdef block?

ElTh0r0:
> I no longer have a Windows machine to test on. Does anyone know if this format is compatible on Windows or if it needs to move to an ifdef block?

No `ifdef` block required for this PR! It adds another numeric date format, which means there is no issue with potential special characters (but I tested if on Windows anyway). IMO this PR can be merged.

> Thanks. Since your proposal was to have even number of options, would you please also add the century (or something similar).

I created PR #4033, which makes the button layout creation independent of even/odd number of items in the list, so we won't have to take care about this constraint in future once #4033 is merged.

ElTh0r0 (approved):

mmahmoudian:
@borgmanJeremy shall I add this to v13 milestone?

borgmanJeremy (approved):

Issue #3720, opened 2024-09-12
Option to remove border on circle count tool


tomturbu:
### Feature Description

I love this tool!
There is just one thing i would appreciate to get improved:
I would like to request to have the black border around the numbered circles as option, which can be enabled or disabled.
Sometimes sometimes it looks better to have a border, sometimes not....

Example:

Left counter done with flameshot (with border) - right counter done with greenshot (no border)
![grafik](https://github.com/user-attachments/assets/342df1e5-dec0-407d-b33a-6642081c7be0)

Of course some people wanna have borders, some not - therefore i would appreciate to have it as option....

bekirsoylu:
So, do we have this config option ?
```
Flameshot v13.3.0 (da913d7f)
Compiled with Qt 6.9.2
winnt: 10.0.26200
windows: 11
```

mmahmoudian:
@bekirsoylu considering your comment in the PR, I'm going to hide your comment in this issue. Thanks for testing it on Windows.

PR #3746, opened 2024-10-07
Added a config option to remove the contrast ring around the counting circle tool

FelixJochems:
This config option removes the contrasting ring around the counting circle tool but enables it again when the user adds an arrow to the circle. I tested it with removing the circle there as well but it looks really unnatural.

I saw someone made a request for the enhancement in #3720 and I already ran it in my own fork of flameshot so decided to give it an actual config option.

### DIFF
--- a/flameshot.example.ini
+++ b/flameshot.example.ini
@@ -52,7 +52,8 @@
 ;disabledGrimWarning=true
 ;
 ;; Disable the contrast on the counting tool when it has no arrow
-;disabledCountingCircleContrast=false
+;disableCountingCircleContrast=false
+;
 ;; Automatically close daemon when it's not needed (not available on Windows)
 ;autoCloseIdleDaemon=false
 ;
--- a/flameshot.example.ini
+++ b/flameshot.example.ini
@@ -51,6 +51,8 @@
 ;; Disable Grim Warning notification
 ;disabledGrimWarning=true
 ;
+;; Disable the contrast on the counting tool when it has no arrow
+;disabledCountingCircleContrast=false
 ;; Automatically close daemon when it's not needed (not available on Windows)
 ;autoCloseIdleDaemon=false
 ;
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -59,6 +59,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initUploadClientSecret();
     initPredefinedColorPaletteLarge();
     initShowSelectionGeometry();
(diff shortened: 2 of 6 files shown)


FelixJochems:
Oh whoops should have squashed before PR

borgmanJeremy:
@mmahmoudian  Do you think this needs a config option? Or do you think we should just pick one as a default?

mmahmoudian:
@borgmanJeremy 
I believe the outline contrast serves a purpose, but the current version we have is not aesthetically pleasing. I think having a setting is a good middle-ground opinion.

What is your opinion?

borgmanJeremy:
@mmahmoudian  Okay cool, as long as you think _some_ people want the outline we can keep it behind a config..

@FelixJochems would you be able to rebase this off the master branch? That should kick off the CI. Once it passes I can merge it.

mmahmoudian:
@borgmanJeremy in some cases I have also face that the circle is not visible if it is a solid color. In these cases usually a shadow, a glow, or an outline is used. I personally like the shadow option more than the outline (feels more modern), but we should implement shadows for all elements at some point, and then perhaps the outline would not be as useful anymore. The KDE's Spectacle has implemented shadow and few other aesthetics that we can adopt 🤓. Flameshot is way more superior than Spectacle imho, but visually the spectacle looks more appealing in certain features.

@FelixJochems I'm curious to know your opinion as well on the design.

borgmanJeremy:
Ooops sorry this is pending me.

I'm good to merge a config for this. IMO I think if config sets the outline to be disabled, the outline should also be disabled when the arrow is drawn. Thoughts on that?

borgmanJeremy:
@mmahmoudian  do you have thoughts on the question above?

borgmanJeremy:
@FelixJochems  Can you make the suggested changes? If not I can implement on a separate PR.

sjvudp:
> This config option removes the contrasting ring around the counting circle tool but enables it again when the user adds an arrow to the circle. I tested it with removing the circle there as well but it looks really unnatural.

To me the ring seems useful, while not being aesthetically pleasing. So besides the option to remove it, the should be an option to configure the ring itself.

image

Specifically I wonder whether the "black-white-black" thin ring should be replaced by a somewhat thicker ring with a single color only.
Alternatively the ring could be drawn in "XOR mode", so it would have "automatic contrast".

bekirsoylu:
I still couldnt remove that circle border. (windows)

> flameshot: error: Unrecognized setting: 'disableCountingCircleContrast'

PR #3765, opened 2024-10-28
[security] replace pixelation by pseudo-pixelation 

### LINKED ISSUE #3289: Option to obfuscate mosaics to avoid retrieving redacted text from mosaics

### LINKED ISSUE #3763: [security] warn users about security issues with pixelation
In the past there have been several issues regarding the problem of pixelation not being secure. Those were closed with comments like "use a black box instead" or "increase the kernel size". However the average user might not even be aware of the fact that there are security issues. What I find especially worring is the fact that pixelations looks much more secure on first glance than it actually is. So simply telling users to use a "large enough" kernel size doesn't help either.

I'd argue that users view pixelation as a security tool, hence it should be treated as such. It was previously proposed to abolish the feature entirely (#1466) - which I think would be the best solution - but that was rejected.

As a middle ground, I'd suggest to warn users about issues of the tool. E.g. when using pixelation for the first time or after a prolonged period of not using it, flameshot could show a popup saying something like:

Warning! Pixelation is not a security tool! Secrets - especially text - can be recovered from pixelated areas in many situations.
To securely hide contents from the screenshot use a black box instead. [More information](https://bishopfox.com/blog/unredacter-tool-never-pixelation)

jrpie:
Fixes #3289
Fixes #3763

Conflicts with PR #3368

# Changes

Replaces pixelation by pseudo-pixelation to make recovering information with tools like unredactor impossible.
Only the fringe of the selected area is used to calculate a pixelation-like effect; the interior does not effect the output and hence can not be recovered.

# Concerns
* The fringe of the selection can still be recovered (i.e. drawing a black box is still more secure)
* The look of the pixelation effect changes (as the interior of the image is no longer used as input)
* It might be better to use HSV-interpolation instead of RGB
* One might want to add an option to switch between pseudo-pixelation and (insecure) true pixelation.

# Examples of the new effect

image
flameshot_test1
image

### DIFF
--- a/src/tools/pixelate/pixelatetool.cpp
+++ b/src/tools/pixelate/pixelatetool.cpp
@@ -8,6 +8,7 @@
 #include <QGraphicsScene>
 #include <QImage>
 #include <QPainter>
+#include <random>
 
 PixelateTool::PixelateTool(QObject* parent)
   : AbstractTwoPointTool(parent)
@@ -46,6 +47,16 @@ CaptureTool* PixelateTool::copy(QObject* parent)
     return tool;
 }
 
+/**
+ * Since pixelation does not protect the contents of the pixelated area
+ * (see e.g. https://github.com/bishopfox/unredacter),
+ * _pseudo-pixelation_ is used:
+ *
+ * Only colors from the fringe of the selected area are used to generate
+ * a pixelation-like effect. The interior of the selected area is not used
+ * as an input at all and hence can not be recovered.
+ *
+ */
 void PixelateTool::process(QPainter& painter, const QPixmap& pixmap)
 {
     QRect selection = boundingRect().intersected(pixmap.rect());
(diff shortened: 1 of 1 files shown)


b0ch3nski:
I can confirm that this patch indeed works and the pixelation while much more secure, still looks good.

pointydev:
I would love to see this merged as an end-user as I still prefer to pixelate as opposed to using a single block colour.

b0ch3nski:
@borgmanJeremy @mmahmoudian please also take a look at this one :slightly_smiling_face: I have been using this PR along couple of others for a month now on my daily basis and I can confirm that it indeed gets the job done.

borgmanJeremy (approved):

borgmanJeremy:
Thanks for the nudge on this one. @mmahmoudian I've tested both this and #3368  . I appreciate the work by both authors, but prefer the effect on this PR. The pixels are still customizable, and the colors from the surrounding blocks help create the same effect as the legacy pixelate mode.

Are you good with me merging this?

mmahmoudian:
@borgmanJeremy the code is well commented and clean, and the effect as you said is pretty much what a normal user would expect from classic pixelation. Also the comments @jrpie made on #3368 makes sense to me.

If you are ok with the code, then let's merge it.

jrpie:
Great, thank you! I guess #3768 could now be reverted

cweiske:
Related: #4082

Issue #3073, opened 2023-01-26
[Wayland KDE] Flameshot doesn't capture full desktop w/ multi-monitor


Riven-Spell:
### Flameshot Version

Flameshot v12.1.0
Compiled with Qt 5.15.5

### Installation Type

User repository (AUR)

### Operating System type and version

Arch Linux (Linux aetherflow 6.1.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 18 Jan 2023 19:54:38 +0000 x86_64 GNU/Linux)

### Description

Flameshot seems to not realize my second monitor exists, while still taking a picture of it. Ctrl+A'ing the screenshot only gets what's shown on my ultrawide.

My setup is a 16:9 1080p monitor on the left-hand side of a 21:9 1080p monitor.

Moving the left monitor to be right of the ultrawide in software allows me to be able to take a proper screenshot on the ultrawide, but still the length is limited to just the ultrawide.

Setting the other monitor as the primary does not affect the position of flameshot or the size of pictures it takes; it just continues to display on the ultrawide.

### Steps to reproduce

1. Run `flameshot gui` on a multi-monitor KDE Wayland setup

Profit.

### Screenshots or screen recordings

![Screenshot_20230126_035353](https://user-images.githubusercontent.com/13892402/214936752-7cbb1397-dbbf-4c28-8ada-69ff43efb4f0.png)

### System Information

1. Arch Linux (Linux aetherflow 6.1.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 18 Jan 2023 19:54:38 +0000 x86_64 GNU/Linux)
2. ![Screenshot_20230126_040020](https://user-images.githubusercontent.com/13892402/214938173-3c55524c-29d4-4e68-b903-f2ee2876eefd.png)
3. KDE/kwin/Wayland

Chryseus:
Can confirm I'm having the exact same issue here, also running Arch with Zen kernel, KDE and Wayland.
Monitor setup:

![image](https://user-images.githubusercontent.com/44627250/214971628-a253e080-562b-4968-95f5-e1507aac29ab.png)

UVClay:
[...]

PR #3766, opened 2024-10-28
Give pins a different window title

typecasto:
Workaround for a KDE issue where the capture window can't be differentiated from the pin window in any way, making the multi-monitor workaround in #3073 also break pins.

### DIFF
--- a/src/tools/pin/pinwidget.cpp
+++ b/src/tools/pin/pinwidget.cpp
@@ -39,7 +39,7 @@ PinWidget::PinWidget(const QPixmap& pixmap,
     // set the bottom widget background transparent
     setAttribute(Qt::WA_TranslucentBackground);
     setAttribute(Qt::WA_DeleteOnClose);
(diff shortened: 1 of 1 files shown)


FelixJochems:
Especially for wayland where flameshot doesn't really work optimally this is a small and elegant solution.

Personally I'd remove any spaces in the title since it somehow always breaks with different escaping characters, other than that; LGTM

typecasto:
Any update on this? This is a pretty major issue for me, I use pins a good bit, and it's a huge pain in the rear having them stuck in the upper right, not to mention that multiple pins don't work at all. Worth noting that it's a single-line fix, too.

mmahmoudian:
@typecasto thanks for the PR. This is indeed useful especially for tiling window managers.

I agree that the space may cause some issue soe some window managers. Would you please remove the space? Otherwise it looks good to me too.

typecasto:
@mmahmoudian Done

mmahmoudian (approved):

PR #3768, opened 2024-10-29
Add warning about pixelation security into the description

AlexP11223:
Updated the tool description as discussed in #3763


mmahmoudian:
Since this is a simple string change, and for now is the best implemented solution we have for this issue, I'll merge with directly, but I strongly believe that we need to address this issue better as discussed in #3763. So let's keep #3763 open.

mmahmoudian (approved):

Issue #3321, opened 2023-08-31
`flameshot gui` exits with 0 on abort


cyqsimon:
### Flameshot Version

Flameshot v12.1.0 (-)
Compiled with Qt 5.15.5

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

EndeavourOS, Linux 6.4.12-arch1-1 x86_64

### Description

I was trying to use flameshot in scripting, and was surprised to find that `flameshot gui` does not exit with non-zero status when user aborts. This makes scripting quite a bit more annoying.

The behaviour I expected is to exit with `1` (or whatever) on abort.

### Steps to reproduce

1. Run `flameshot gui` in terminal
2. Hit <kbd>esc</kbd> to abort
3. Run `echo $?`, notice that the status is `0`

### System Information

Operating System: EndeavourOS 
KDE Plasma Version: 5.27.7
KDE Frameworks Version: 5.109.0
Qt Version: 5.15.10
Kernel Version: 6.4.12-arch1-1 (64-bit)
Graphics Platform: X11

muzilong:
same problem

serebryakov02:
Question for the developers: How can I run Flameshot from the terminal using the source code? When I launch it from Qt Creator and follow the instructions, I do not receive any messages. However, if I download Flameshot as a standalone program and launch it according to the instructions,  I can reproduce.

guyueshui:
Since this issue marked as "Good first issue" by member of flameshot, and I've used flameshot for quite a while and am happy to make some contributions if possible. So I tried read the code and found this simple solution, hope it helps.

Please let me know if I was wrong, this is my first pull request (#3786) on my life though.

mmahmoudian:
Practically this issue is addressed via the merged PR (#3786), but based on the discussion in the PR I will keep this issue open and will transfer some comments here:

@FelixJochems [mentioned](https://github.com/flameshot-org/flameshot/pull/3786?new_mergebox=true#pullrequestreview-2581745052):

> In the future we might change your exit code to any 128+ code as to differentiate between an actual program crash/error and program termination by the user itself.

@mmahmoudian [mentioned](https://github.com/flameshot-org/flameshot/pull/3786?new_mergebox=true#issuecomment-2623927313):

[...]

Issue #3755, opened 2024-10-16
Make flameshot process return exit code 1 when using `gui` and the screenshot is aborted


rodrigo-morales-1:
### Feature Description

Today I was writing a Python script that captures a region in the screen using `flameshot`. I noticed that flameshot returns the error code `0` when the screenshot is stored and when then screenshot is aborted.

Consider the Python script shown below which run `flameshot`:

```python
import os
import subprocess
import tempfile

temporary_directory = tempfile.mkdtemp()
temporary_file = os.path.join(temporary_directory, 'screenshot.png')
process_flameshot = subprocess.run(
  ['flameshot', 'gui', '--path', temporary_file],
  capture_output = True)

print(f"""stdout: '{process_flameshot.stdout.decode('utf-8').rstrip()}'""")
print(f"""stderr: '{process_flameshot.stderr.decode('utf-8').rstrip()}'""")
print(f"""returncode: {process_flameshot.returncode}""")
```

The following is the output when the user makes the selection and presses the button with a check mark (i.e. the screenshot is saved).

```text
$ python3 /tmp/a.py
stdout: ''
stderr: 'flameshot: info: Capture saved as /tmp/tmpcgi9smow/screenshot.png'
returncode: 0
```

The following is the output when the user presses the key ESC or the user makes the selection and presses the button with an X (i.e. the screenshot is aborted).

```text
$ python3 /tmp/a.py
stdout: ''
stderr: 'flameshot: info: Screenshot aborted.'
returncode: 0
```

Note that `returncode` equals `0` in both cases.

In the current version, if I wanted to check if the user aborted the screenshot using Python, I would need to look for the string `Screenshot aborted` in STDERR (as shown below):

```python
import os
import subprocess
import tempfile

temporary_directory = tempfile.mkdtemp()
temporary_file = os.path.join(temporary_directory, 'screenshot.png')
process_flameshot = subprocess.run(
  ['flameshot', 'gui', '--path', temporary_file],
  capture_output = True)

if "Screenshot aborted" in process_flameshot.stderr.decode('utf-8'):
  print("The screenshot wasn't saved.")
```

If the process returned 1, when the screenshot was aborted, then the code would look as shown below. To me the code below looks cleaner than the code above:

```python
import os
import subprocess
import tempfile

temporary_directory = tempfile.mkdtemp()
temporary_file = os.path.join(temporary_directory, 'screenshot.png')
process_flameshot = subprocess.run(
[...]

PR #3786, opened 2024-11-19
Correct the exit value on abort (#3321)

guyueshui:
`flameshot gui` exits with 0 on abort, we can use the return value of `qApp->exec()` in `requestCaptureAndWait` in main.cpp.

图片

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -45,7 +45,7 @@ void wayland_hacks()
 }
 #endif
 
-void requestCaptureAndWait(const CaptureRequest& req)
+int requestCaptureAndWait(const CaptureRequest& req)
 {
     Flameshot* flameshot = Flameshot::instance();
     flameshot->requestCapture(req);
@@ -66,7 +66,7 @@ void requestCaptureAndWait(const CaptureRequest& req)
         AbstractLogger::info() << "Screenshot aborted.";
         qApp->exit(1);
     });
-    qApp->exec();
+    return qApp->exec();
 }
 
 QSharedMemory* guiMutexLock()
@@ -428,7 +428,7 @@ int main(int argc, char* argv[])
                 req.addSaveTask();
             }
         }
-        requestCaptureAndWait(req);
+        return requestCaptureAndWait(req);
     } else if (parser.isSet(fullArgument)) { // FULL
         reinitializeAsQApplication(argc, argv);
 
@@ -463,7 +463,7 @@ int main(int argc, char* argv[])
         if (!clipboard && path.isEmpty() && !raw && !upload) {
             req.addSaveTask();
         }
-        requestCaptureAndWait(req);
+        return requestCaptureAndWait(req);
     } else if (parser.isSet(screenArgument)) { // SCREEN
         reinitializeAsQApplication(argc, argv);
 
@@ -513,7 +513,7 @@ int main(int argc, char* argv[])
             req.addSaveTask();
         }
 
-        requestCaptureAndWait(req);
+        return requestCaptureAndWait(req);
     } else if (parser.isSet(configArgument)) { // CONFIG
         bool autostart = parser.isSet(autostartOption);
         bool filename = parser.isSet(filenameOption);


mmahmoudian:
@guyueshui Thanks for the PR. I will merge it after we get the opinion of some of the main contributors.

guyueshui:
> @guyueshui Thanks for the PR. I will merge it after we get the opinion of some of the main contributors.

@mmahmoudian OK👌

jack9603301:
@guyueshui I have a question, when we execute exit, shouldn't the program terminate immediately? Is there any meaning of the return value in this case?

mmahmoudian:
@jack9603301 
> @guyueshui I have a question, when we execute exit, shouldn't the program terminate immediately? Is there any meaning of the return value in this case?

This is related to #3321 and #3755

guyueshui:
@mmahmoudian May i ask when would this SIMPLE PR be merged. Seems like other dev do not have any time though. 🤔

FelixJochems (approved):
As I've mentioned in the Matrix chat, I'm OK with these changes for now since it basically passes the exit code from the function as the exit code for the entire program.

In the future we might change your exit code to any 128+ code as to differentiate between an actual program crash/error and program termination by the user itself.

All in all; OK from me

mmahmoudian:
I discussed this issue with a friend IRL and they suggested pretty much the same thing that @FelixJochems suggested. There must be a difference between the actual error within the software and the other type of interrupts we want to signal. Based on my searches, our exit code should be $$1< x < 126$$ because the rest have specific meanings:

- `0`: Success
- `1`: Catch-all for general errors
- `2`: Misuse of shell builtins (according to Bash documentation)
- `126`: Command invoked cannot execute
- `127`: "command not found"
- `128`: Invalid argument to exit
- `128+n`: Fatal error signal "n"
  - > Indicates that the command or program was terminated by a fatal error signal. For example, an exit code of 137 (128 + 9) means that the command was terminated by a SIGKILL signal. [[1](https://www.baeldung.com/linux/status-codes#fatal-error-signal), [2](https://www.agileconnection.com/article/overview-linux-exit-codes), [3](https://stackoverflow.com/a/1535733/1613005)]
- `130`: Script terminated by Ctrl + C
- `255`: Exit status out of range

<br/><br/>

@FelixJochems I couldn't find any documentation for having the first 128 codes reserved. Is there any reference that you can recommend me to read? Because I was about to suggest to start incrementing numbers for such exit codes and for example reserve exit code `2` for this issue.

<br/><br/>

@guyueshui Since it is approved by two other devs this can now be merged, but would it be possible for you to update this PR to return exit code other than 1?

guyueshui:
> I discussed this issue with a friend IRL and they suggested pretty much the same thing that @FelixJochems suggested. There must be a difference between the actual error within the software and the other type of interrupts we want to signal. 

@mmahmoudian I quite agree with it, it's really a good practice to use different exit code to distinguish the different program behavior. But I want to mention that it's will not covered in this simple PR. Because currently I just forward the exit code of the global `qApp` instance (the return value of `qApp->exec()`). 

@FelixJochems says they have a plan (not now) to use some well designed exit code to describe those behavior. And maybe in a near future I can help with it, but not now, cause I haven't read the related code in much more depth. Moreover, such changes will need more careful reviews, which will take a long time to be merged based on current situation. So I suggest that we could merge this PR (simple enough) first.

mmahmoudian (approved):

mmahmoudian (approved):

mmahmoudian:
I collected some of the comments regarding the "exit codes > 1" in the following post:

https://github.com/flameshot-org/flameshot/issues/3321#issuecomment-2626767911

@guyueshui Thank you for the PR.

FelixJochems:
@mmahmoudian I naively thought that most exit codes were standardized, but it appears that only 0 & 1 are really well defined. I saw somewhere that the user defined exit codes are generally 128+ but this was false.

So in general, I'm okay with any exit code other than 1 & 2

PR #3796, opened 2024-12-07
Add Fluxbox configuration snippet

WojciechMula:
Update README


borgmanJeremy:
Thanks! I'm going to trust this is correct as I don't have a fluxbox VM.

PR #3821, opened 2025-01-24
Update build/install instructions for custom install dir translations to work

MatthiasKunnen:
Flameshot uses CMAKE_INSTALL_PREFIX to set APP_PREFIX.
APP_PREFIX is used to determine one of the potential locations of the translations.
This can be observed in https://github.com/flameshot-org/flameshot/blob/10d12e0b54d59de2ac2567c540a93113672cd884/src/utils/pathinfo.cpp#L26
CMAKE_INSTALL_PREFIX defaults to /usr/local on UNIX.

The problem is that APP_PREFIX is set during the build -S step and changing this
using build --install build --prefix ... does not actually change the value of
APP_PREFIX. This leads to the translations not being found when running
from a custom install location.

To combat this, the [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html) env var is used to properly set
the APP_PREFIX. This requires CMAKE 3.29.

Alternatives:
- Get rid of APP_PREFIX and try to find the translations using path traversal.
- Use --install-prefix and --prefix, this requires cmake 3.21. Also possible is to use
  -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR> which would not require a
  newer CMAKE. This requires specifying the install dir twice.
- Add a symbolic link next to the binary and use the same method used
  to make running from the build directory possible.
  `ln -s "$INST_DIR/bin/translations" ../share/flameshot/translations`

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,7 +401,7 @@ Also you can open and build/debug the project in a C++ IDE. For example, in Qt C
 - Qt >= 5.9
   + Development tools
 - GCC >= 7.4
-- CMake >= 3.13
+- CMake >= 3.21
 
 #### Run-time
 
@@ -461,7 +461,7 @@ nix-shell
 
 #### macOS
 
-First of all you need to install [brew](https://brew.sh) and than install the dependencies
+First of all you need to install [brew](https://brew.sh) and then install the dependencies
 ```shell
 brew install qt5
 brew install cmake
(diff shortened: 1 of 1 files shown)


mmahmoudian:
Thanks for the PR. Other than one shellcheck comment, the rest looks perfect. Thanks.

MatthiasKunnen:
@mmahmoudian, the potential word splitting issue? I added a commit to fix it. If you meant something else, let me know and I'll take a look

mmahmoudian (approved):

mmahmoudian:
@MatthiasKunnen Thanks for the PR. I personally tend to run the shell snippets with `shellcheck` in order to maintain a good level of scripting. I basically copied your snippet into a file and ran `shellcheck` on it and commented the warning it rose. It is unlikely that `brew --prefix qt5` returns anything with spaces, but nonetheless, better be safe than sorry 😊

Thanks for yet another clean and detailed PR. It is always great to have such contributors that spend time explaining the issue in details. Cheers.

PR #3822, opened 2025-01-24
Fix English translations not used despite being preferred

MatthiasKunnen:
When specifying the preferred language on Linux, the `$LANGUAGE` env var is used. See <https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html>.

**Example:**
`$ LANGUAGE=en:nl flameshot`

**Expected:**
flameshot should be in English as this is a localization that flameshot provides. If English would not be supported, it should fall back to using Dutch (nl).

**Actual:**
flameshot uses Dutch translations.

**Cause:**
If an empty translation file for the base language (here "en") is not provided, the base language is only used as a fallback.
See <https://bugreports.qt.io/browse/QTBUG-69196>

Additional search keywords:
- $LANGUAGE not respected
- Wrong language used

### DIFF
--- a/data/translations/Internationalization_en.ts
+++ b/data/translations/Internationalization_en.ts
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<!--
+    An empty translation file is required for the base language, or it will only be used as a
+    fallback if the user prefers other supported languages. E.g.
+    $ LANGUAGE=en:nl flameshot
+    will result in Dutch translations being used despite English being higher priority.
+    See https://bugreports.qt.io/browse/QTBUG-69196
+-->
+<TS version="2.1" language="en">
+</TS>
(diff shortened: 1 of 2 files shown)


mmahmoudian:
Thank you for the PR and the detailed explanations. Looks good to me. Let's just wait for one more dev to review the PR too.

mmahmoudian (approved):

FelixJochems (approved):
Had to recreate the situation and even thought the bugtracker says this is fixed in QT5.12 my flameshot (built with QT 5.15) still had the issue, so it's still a valid workaround.

Thanks for providing a fix AND a cause!

MatthiasKunnen:
Happy to help! It was a weird one to root cause, especially because some other QT applications have the same issue leading me to think something on my system was broken :D

It's important to know, though, that this is not just a workaround, but a requirement. Translations must be provided for the base language. I seem to recall that QT recommends this in more recent documentation.

If you are interested in some more history regarding this issue, QT is not the only platform that has this requirement. Quoting some information from an almost identical [GNU gettext bug](https://savannah.gnu.org/bugs/?53516#comment5):

> The source language (the msgid string) should be considered to be of language called "C", not English nor en_US.
> ...
> The reality is that all those applications are technically missing all the localizations and the only way to use those applications in English is to fallback all the way to "C".


**opinion ahead**  
An opinion I personally developed over time working with internationalization is that msgids should just be that, IDs. Using the msgid as the translation results in problems such as:
- Inability to provide context specific localization such as pluralization
- Often problematic unicode support, e.g. GNU gettext msgids should be ASCII 7-bit.
- Misuse of the same msgid where a translation might be changed from e.g. "Yes" to "Agree" in one place but it does not make sense in another.
- Typos in msgids require updating all translation files.

PR #3685, opened 2024-08-02
ci: enhancements and updates

holazt:
- move to quay from dockerhub due to pull rate limit
- add opensue leap 15.5
- the building envrinments of appimage/snap/flatpak update to ubuntu 22.04
- separate RPM spec into two files: one is for fedora/rehl, another is for opensuse leap
- update transferwee.py from upstream

PR #3830, opened 2025-01-31
CI: migrate all upload-artifact actions from v3 to v4

mmahmoudian:
The 4 failed CIs are irrelevant to this PR. The RPM CIs fail because this PR should be merged: #3685, and the Windows CI fails because of a python package which is not used by us. I think we should update the action there too, but that needs it's own PR.


mmahmoudian:
Considering that
1. this is non-destructive change and has nothing to do with out codebase (only updates the CI to bump the Action version)
2. every other PR and our nightly builds are dependent on this one (because literally all the CIs are failing), 
3. it has been 3 days since I created the PR and no objects was raised (no approval from other devs either 🤷🏼 )

I will merge this PR to keep the ball rolling for the repo.

Please comment if you believe this is problematic PR or wrong in some way. it is very easy to roll back if needed.

FelixJochems:
I overlooked this one on accident, but fully agree with your statement. No issues from me

PR #3851, opened 2025-02-21
Remove QEMU dummy package installation from Linux-pack.yml

volovikariel:
Motivation: I don't want the CI to fail when tests are run on ARM

image

Solution reasoning:
The QEMU package is now a dummy package. Installing `qemu-user-static` should keep functionality and fix the failing workflow.

See https://packages.debian.org/buster/qemu and https://packages.debian.org/bullseye/qemu for official mention of it being a dummy package.

image

Current state CI run: https://github.com/volovikariel/flameshot/actions/runs/13458590444
With this change: https://github.com/volovikariel/flameshot/actions/runs/13458428232

### DIFF
--- a/.github/workflows/Linux-pack.yml
+++ b/.github/workflows/Linux-pack.yml
@@ -142,7 +142,7 @@ jobs:
           esac
           if [ -n "${QEMU_ARCH}" ]; then
             sudo apt-get -y -qq update
-            sudo apt-get -y install qemu binfmt-support qemu-user-static
+            sudo apt-get -y install binfmt-support qemu-user-static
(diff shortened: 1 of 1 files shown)


volovikariel:
Putting on hold as there seems to be an issue later on in the build process.

image

volovikariel:
Actually all the others succeeded, only this arm64 and the fedora ones failed.
Thinking to figure these issues out too, but it could be in another PR.

mmahmoudian:
@volovikariel thanks for the PR and being careful about which CI breaks.

@holazt what's your opinion on this?

holazt:
LGTM

mmahmoudian:
@volovikariel thanks for the PR

PR #3859, opened 2025-03-04
Improve the speed of calling grim

lhz07:
When the output file is specified in the "ppm" format, Grim can operate more efficiently. Additionally, loading a pixmap from a file is significantly faster than retrieving it from standard output. In my test, this optimization can reduce the time by approximately 150 milliseconds.

old method:
old_method
new method:
new_method

Already tested this change on archlinux with hyprland.


lhz07:
@mmahmoudian hi, looks like you're the most active maintainer... would you mind giving this a review?

borgmanJeremy:
@lhz07 Do you know what format is used if PPM is not specified? Sorry for the delay, Im reviewing PR's now.

lhz07:
@borgmanJeremy Grim will use PNG by default, and the compression level is 6 by default, which will take more time for grim to compress the image. See https://gitlab.freedesktop.org/emersion/grim/-/blob/master/doc/grim.1.scd

borgmanJeremy (approved):

Issue #3852, opened 2025-02-21
Fractional Scaling. The image is enlarged when the area is captured. (Question)


DmitrySkibitsky:
Are there any options to fix this problem?

Demo Video: https://drive.google.com/file/d/1WXLTLFUxyOz_JUtQi04F0jSp6HzDVJh0/view?usp=drive_link

System
```bash
OS: Manjaro Linux x86_64
Display (TMX1401): 2560x1600 @ 165 Hz (as 2048x1280) in 14" [Built-in]
DE: GNOME 47.4
WM: Mutter (Wayland)
```

flameshot
```bash
extra/flameshot 12.1.0-5 (589.6 KiB 2.8 MiB) (Installed)
```

Settings / Displays
![Image](https://github.com/user-attachments/assets/26fdb80f-6bb6-4f39-9f95-293421910dda)

Activated this parameter in Refine
![Image](https://github.com/user-attachments/assets/5d6d2db7-d9d1-45f4-801e-40f689429ddf)

I really like your app. There aren't even any sane alternatives.

Oopsryuu:
You can run this command in your terminal 
`bash -c -- 'QT_SCREEN_SCALE_FACTORS="1;1.25" flameshot gui'`
Replace 1.25 to your scale (if you are not using 125%).

I save it as a script, like **flameshot.bash**:
`#!/bin/bash`
`QT_SCREEN_SCALE_FACTORS="1;1.25" flameshot gui`

and go Settings -> Keyboard -> View and Customize Shortcuts -> Custom Shortcuts
set your shortcut and set command to `bash ~/path/to/flameshot.bash` .

I use Ubuntu 24.04, but it should solve your problem.

hotdogc1017:
Using it still does not display the screen scaling correctly
```bash
bash -c -- 'QT_SCREEN_SCALE_FACTORS="1;1.25" flameshot gui'
```

By the way, I also set the screen scaling to 125%

DmitrySkibitsky:
@Oopsryuu @hotdogc1017  Thank you for responding. But unfortunately it didn't work. Now it looks like this. I can't decide which option I like better: when the image is smaller or when the image is enlarged 🫠

![Image](https://github.com/user-attachments/assets/0beccdb6-71f9-4f96-891b-17cd75fb697b)

FelixJochems:
Could you try and build with #3869 applied?

DmitrySkibitsky:
[...]

PR #3869, opened 2025-03-11
Fix issue #564, that in wayland fractional scaling situtation, screenshot preview size is wrong.

SFBB:
Fixing screenshot preview size is not correct because wayland dpr is not correct when using fractional scaling. This fix supports the idea that we will use dpr in the xcb platform because it is correct in that case.

Tested on Manjaro wayland session with fractional scaling of 125%, both QT_QPA_PLATFORM=xcb and QT_QPA_PLATFORM=wayland work as expected.

Multi-monitor environment is not tested because I don't have one.

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -80,7 +80,16 @@ void ScreenGrabber::freeDesktopPortal(bool& ok, QPixmap& res)
             QUrl uri = map.value("uri").toString();
             QString uriString = uri.toLocalFile();
             res = QPixmap(uriString);
-            res.setDevicePixelRatio(qApp->devicePixelRatio());
+            QScreen* screen = QGuiAppCurrentScreen().currentScreen();
+            if (screen)
+            {
+                const auto screenRect = screen->geometry();
+                res.setDevicePixelRatio(res.height() * 1.0f / screenRect.height());
+            }
+            else
+            {
+                res.setDevicePixelRatio(qApp->devicePixelRatio());
+            }
             QFile imgFile(uriString);
             imgFile.remove();
         }
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -74,21 +74,25 @@ void ScreenGrabber::freeDesktopPortal(bool& ok, QPixmap& res)
       this);
 
     QEventLoop loop;
-    const auto gotSignal = [&res, &loop](uint status, const QVariantMap& map) {
+    const auto gotSignal = [&res, &loop, this](uint status, const QVariantMap& map) {
         if (status == 0) {
             // Parse this as URI to handle unicode properly
             QUrl uri = map.value("uri").toString();
             QString uriString = uri.toLocalFile();
             res = QPixmap(uriString);
-            QScreen* screen = QGuiAppCurrentScreen().currentScreen();
-            if (screen)
-            {
-                const auto screenRect = screen->geometry();
-                res.setDevicePixelRatio(res.height() * 1.0f / screenRect.height());
-            }
-            else
+            QRect approximatedPhysicalDesktopGeometryBasedOnDevicePixelRatio = desktopGeometry();
+            QRect accurateLogicalDesktopGeometry = logicalDesktopGeometry();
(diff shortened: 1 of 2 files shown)


b0ch3nski:
Unfortunately, this doesn't seem to work on a multimonitor setup (3 horizontal, 1 vertical - 2 of them have a fraction scaling of 1.45). All screens on the screenshot preview seems to be very much upscaled which makes it rather unusable.

This is the screen config I've been using on Sway:
```
output "XXX" enable mode 1920x1080@60.0Hz pos 0 0
output "XXX" enable mode 2880x1800@60.001Hz scale 1.45 pos 1920 0
output "XXX" enable mode 2880x1800@60.001Hz scale 1.45 pos 1920 1241
output "XXX" enable mode 1920x1200@59.95Hz pos 3906 0 transform 270
```

SFBB:
> Unfortunately, this doesn't seem to work on a multimonitor setup (3 horizontal, 1 vertical - 2 of them have a fraction scaling of 1.45). All screens on the screenshot preview seems to be very much upscaled which makes it rather unusable.

Thank you for your testing!

Yes, that's true. In theory, it only works on multiple monitors if all monitors share the same fractional ratio.

I think it is because Qt doesn't support high dpi on Linux Wayland well. It still doesn't use [fractional_scale_v1](https://wayland.app/protocols/wayland-protocols/303) for its QWaylandScreen IIRC. And this will cause if you display a physical size image on your screen, it will not render in the logical size, at least the correct one. As far as I know, Qt should render image based on per monitor fractional scaling, but right now it is still based on a interger instead of float.
Anyway, if they don't support this, I don't think it is a good idea to handle such a complex situation in a downstream application, or even this is impossible.

My PR only fixes simple cases. And such fix should be reverted if Qt does its job well.

b0ch3nski:
@SFBB It seems that I wrote my comment too early, dang! I've tracked the issue back to `grim`. I've disabled it, build with your patch and now it works as described even with the complex multimonitor setup with different orientations and fractional scaling! Thank you for this, it was bugging me for months!

DmitrySkibitsky:
This solved the problem with fractional scaling
Thanks to the author of this fix
https://github.com/flameshot-org/flameshot/issues/3852#issuecomment-2776408188

mmahmoudian:
@borgmanJeremy do you believe this PR is ready to be merged?

borgmanJeremy:
Generally yes, but it would be good for someone to test the 2nd code path. I personally think the variable names are too long but that's more of a preference than an issue.  I think it's at least worth approving the CI so the CI can run. I think the code formatting will be rejected.

This scaling issue still exists on Qt6 btw, I've tested it with a minimal example.

mmahmoudian (approved):

mmahmoudian:
> I think it's at least worth approving the CI so the CI can run. I think the code formatting will be rejected.

This is something I've been struggling with lately. In some PRs the CIs are not available. I approved the code here but still ... . I should look into this.

borgmanJeremy:
@mmahmoudian I have some time lately so I can probably fix the CI if you re-add me as a contributor (or whatever permissions needed to run CI).

@SFBB  Yeah I think thats much better thanks! You can just update this PR. Curious if that triggers a successful CI kick off.

mmahmoudian:
@SFBB 
> I'm considering renaming the variables as follows: `approximatedPhysicalDesktopGeometryBasedOnDevicePi` -> `approxPhysGeo` `accurateLogicalDesktopGeometry` -> `logicalGeo`
> 
> Though it requires a bit of effort to understand, it is good for me. But I'd like your thoughts. Are you okay with me updating the pull request with these changes, or do you prefer I avoid modifying it dynamically?

I think the proposed variable changes are good, but considering that your original variable names are descriptive, I suggest write them as descriptive comments to be more clear about the intention behind each variable.

@borgmanJeremy Thanks, I sent you an invitation.

SFBB:
@borgmanJeremy @mmahmoudian 
Based on the clang-format output, I changed the code style to meet the requirement. I hope we can pass the CI this time.

scanash00:
This fixed an issue that has been around for a really long time, thank you! I have built it and I can confirm that it is working. Good job.

Issue #3471, opened 2024-01-11
snap install failing due to deprecated dependency


nathanielfinkgapmaps:
### Flameshot Version

latest snap

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Xubuntu 23.10

### Description

Requiring kde framework 5-15-3 which is no longer available

sudo snap install flameshot
error: cannot perform the following tasks:
- Ensure prerequisites for "flameshot" are available (cannot install prerequisite "kde-frameworks-5-91-qt-5-15-3-core20": no snap revision available as specified

### Steps to reproduce

run sudo snap install flameshot

### System Information

Xubuntu 23.10

nahojjjen:
I get a similar error on a new Kubuntu 23.10 install when attempting to install through Discover:
![image](https://github.com/flameshot-org/flameshot/assets/8545451/5d6c0bdd-9db4-4d61-b5ed-9e07706ba705)

spintokha:
Same issue on Ubuntu 23.10.
`snap install flameshot`
_error: cannot perform the following tasks:
- Ensure prerequisites for "flameshot" are available (cannot install prerequisite "kde-frameworks-5-91-qt-5-15-3-core20": no snap revision available as specified)_

OnkelTem:
Same here, Kubuntu 22.04:

```
$ snap install flameshot
error: cannot perform the following tasks:
- Ensure prerequisites for "flameshot" are available (cannot install prerequisite "kde-frameworks-5-91-qt-5-15-3-core20": no snap revision available as specified)
```

So a distro-agnostic version depend on the distro obviously? Such a surprise.

P.S. I'm nostalgic for the times when we didn't have all these "distro-agnostic" formats like snap, flatpack, appimage etc. You could just run `apt-get install aTthing` and it got installed.

BhattJayD:
Same issue on Ubuntu 22.04.4

```bash
$ sudo snap install flameshot
error: cannot perform the following tasks:
- Ensure prerequisites for "flameshot" are available (cannot install prerequisite "kde-frameworks-5-91-qt-5-15-3-core20": no snap revision available as specified)
```

MM1132:
Same exact issue on `Linux Mint 21.1`:

```
~$ sudo snap install flameshot
error: cannot perform the following tasks:
[...]

Issue #3714, opened 2024-09-05
Snap installation failure


boukendesho:
### Flameshot Version

v12.1.0+git4.33f48f4a

### Installation Type

Snap from Snapcraft

### Operating System type and version

Ubuntu 24.04.1 LTS

### Description

$ sudo snap install flameshot
error: cannot perform the following tasks:
- Ensure prerequisites for "flameshot" are available (cannot install prerequisite "kde-frameworks-5-91-qt-5-15-3-core20": no snap revision available as specified)

This is obviously because of the kde framework snap has been updated to a newer version, we need build again use latest kde framework snap.

### Steps to reproduce

run `sudo snap install flameshot`

### Screenshots or screen recordings

no

### System Information

Ubuntu 24.04.1 LTS Wayland

ThomasKregelin:
I have the same issue. Just to confirm the issue.

araissaalmeida:
Estou passando por esse problema agora mesmo. Quando será resolvido?

jesseli2002:
Probably duplicate of https://github.com/flameshot-org/flameshot/issues/3119

boukendesho:
Closed for the duplicate

Issue #3119, opened 2023-03-06
Unable to install flameshot using snap in Ubuntu 22.04.2 LTS


Kisaragi-ng:
### Flameshot Version

v12.1.0+git4.33f48f4a

### Installation Type

Snap from Snapcraft

### Operating System type and version

Ubuntu 22.04.2 LTS amd64

### Description

flameshot error during installation due to prerequisites is not available from snap store. flameshot require kde-frameworks-5-91-qt-5-15-3-core20 but it's not found [snapcraft.io/search?q=kde-frameworks-5](https://snapcraft.io/search?q=kde-frameworks-5)

### Steps to reproduce

1. run `sudo snap refresh`
2. `sudo snap install flameshot`

```
error: cannot perform the following tasks:
- Ensure prerequisites for "flameshot" are available (cannot install prerequisite "kde-frameworks-5-91-qt-5-15-3-core20": no snap revision available as specified)

```

### Screenshots or screen recordings

-

### System Information

-

mlec1:
Same for me. 

It seems that on snap, the oldest available kde is kde-frameworks-5-96-qt-5-15-5-core20
https://snapcraft.io/search?q=kde-frameworks-5

berniland:
exactly same prerequisites error on Centos 9

jtournier205:
Same issue on CentOS7, but dependency is kde-frameworks-5-91-qt-5-15-3-core20

kondzerau-ivan:
Same problem into the Ubuntu Studio 22.04.

luzpaz:
CC @borgmanJeremy (are you the [flameshot snap maintainer](https://snapcraft.io/publisher/borgman-jeremy) ?)

borgmanJeremy:
I currently am but stepped down from this project. When someone else wants to take it over I will transfer.

Arkasha-P:
Hi same problem into the Ubuntu 23.04.

IlyaZha:
Same Ubuntu 20

poxeron:
Kali linux 6.1.0-kali9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.27-1kali1 x86_64 GNU/Linux - the same issue

ronilaukkarinen:
Same for Pop_OS!

```
NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
[...]

PR #3870, opened 2025-03-12
Remove link to Snap in README

jesseli2002:
https://github.com/flameshot-org/flameshot/issues/3471 `snap install flameshot` has been failing for some time on some distributions, e.g. possibly at least since Ubuntu 22.04 and possibly earlier. `apt install flameshot` does work and is the method listed at [Flameshot's website](https://flameshot.org/docs/installation/installation-linux/) for various Linux distributions, and it also notes that Flameshot isn't on Snap yet. 

Maybe easiest is to stop directing people towards snap via the first thing you see on the README if it doesn't work consistently/isn't actively being supported. It seems to work maybe on some other distros, but the users of those distros are probably more tech-savvy and will know how to actually figure out install failures.

Recommending that https://github.com/flameshot-org/flameshot/issues/3471 gets closed if this gets merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,6 @@
       <img src="https://img.shields.io/github/release/flameshot-org/flameshot.svg?label=docs" alt="Docs" />
     </a>
     <br>
-    <a href="https://snapcraft.io/flameshot">
-      <img alt="Get it from the Snap Store" src="https://snapcraft.io/static/images/badges/en/snap-store-black.svg" />
-    </a>
     <a href="https://flathub.org/apps/details/org.flameshot.Flameshot">
       <img height="60" alt="Download on Flathub" src="https://flathub.org/assets/badges/flathub-badge-en.svg"/>
     </a>


jesseli2002:
All issues I found that related to this:
- https://github.com/flameshot-org/flameshot/issues/3119 snap maintainer has stepped down; issue reported at least as early as Ubuntu 20
- https://github.com/flameshot-org/flameshot/issues/3714 same issue. There's a proposed merge request to update to snap core22 (to be honest, I don't know what that means or if it should fix the problem) which was closed due to upstream change https://github.com/flameshot-org/flameshot/commit/3fafcf4aa9ae3d620ff691cba3a1a2195d592914 
- https://github.com/flameshot-org/flameshot/issues/3471 already listed above

mmahmoudian:
I believe this is the wrong approach. The correct approach is to fix the snap and push it to snapcraft.

borgmanJeremy:
@mmahmoudian To fix the snap build I need access to the flameshot-org, not just flameshot repo.  I think there is also a way you can somehow authorize the snapcraft app and then maybe I can see it without flameshot-org permissions?

borgmanJeremy:
I'm going to close this as I would rather just fix the snap.

mmahmoudian:
@borgmanJeremy 

>  To fix the snap build I need access to the flameshot-org, not just flameshot repo. I think there is also a way you can somehow authorize the snapcraft app and then maybe I can see it without flameshot-org permissions?

Sorry, my bad. I usually forget that the repo level membership is different from organization membership. I initially intended to [re]invite you to the org rather than repo. Sorry for the inconvenience. Please let me know if some further settings needs to be set (e.g., on Matrix, email, or here)

PR #3888, opened 2025-03-27
Enable high dpi scaling in QT App settings

fishhead2567:
Fixed the gui blinking on my high-dpi monitor


b0ch3nski:
I've build with this patch but since I did not have a blinking before, I can only confirm that it doesn't break anything for me.

borgmanJeremy:
@fishhead2567  under what circumstances do you see blinking?

PR #3895, opened 2025-04-01
Introduce PKGBUILD for Arch Linux builds

b0ch3nski:
This PR introduces `PKGBUILD` file that allows building Flameshot on Arch Linux, based on local repository. With this change, one can quickly modify the source code and build an installable package with no extra hassle.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,0 +1,49 @@
+pkgname=flameshot-git
+_pkgname=flameshot
+pkgver=r1981.86a646c3
+pkgrel=1
+pkgdesc="Powerful yet simple to use screenshot software"
+arch=('i686' 'x86_64' 'aarch64' 'armv7h')
+url="https://github.com/flameshot-org/flameshot"
+license=('GPL-3.0-or-later')
+depends=('qt5-svg' 'hicolor-icon-theme' 'kguiaddons5')
+makedepends=('qt5-tools' 'cmake')
+optdepends=(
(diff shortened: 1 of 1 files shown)


kiecla:
Fair enough. My bad.

borgmanJeremy:
@mmahmoudian thoughts on including a pkgbuild here? Seems fine to me but I would prefer for it to go in the packaging directory.

mmahmoudian:
I think this should be merged with the pgkbuild we have in the AUR. I diffed them last week but I didn't finish the merging.

I also agree with @borgmanJeremy that the better path for this file is in that subfolder. 

@borgmanJeremy 
Perhaps we can accept this PR as-is and then relocate the file and merge it with AUR?

borgmanJeremy:
Sure I am okay with that. I'm thinking we should modify it to build the latest commit similar to the AUR?

mmahmoudian:
I assigned it to myself to have it as a todo.

borgmanJeremy (approved):

b0ch3nski:
@borgmanJeremy @mmahmoudian Thanks for accepting my PR :)

Nevertheless, please do not merge this with the PKGBUILD in AUR - they are intentionally different :) I've mentioned that already in the comment here previously: https://github.com/flameshot-org/flameshot/pull/3895#issuecomment-2778728701

The AUR version is cloning a master branch of the repository while the PKGBUILD I've submitted builds the **local source tree**. This is by design - I want to pull other commits and test them on my own machine by producing a new package that can be handled by pacman itself.

PR #3901, opened 2025-04-05
Bumped minimum Cmake version to fix build issues on Cmake 4.0.

borgmanJeremy:
This resolves the aur issue building flameshot-git. Flameshot also will fail to build in arch until this is bumped. Cmake 3.22 was chosen since that is the version on Ubuntu 22.04.

Ideally all the external dependencies need refreshed, but that's much more work.

### DIFF
--- a/external/Qt-Color-Widgets/CMakeLists.txt
+++ b/external/Qt-Color-Widgets/CMakeLists.txt
@@ -14,10 +14,8 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-cmake_minimum_required (VERSION 3.1 FATAL_ERROR)
-if ( ${CMAKE_VERSION} VERSION_GREATER "3.12.0")
-    cmake_policy(SET CMP0077 NEW)
-endif()
+cmake_minimum_required (VERSION 3.22 FATAL_ERROR)
+cmake_policy(SET CMP0077 NEW)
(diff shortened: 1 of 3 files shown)


FelixJochems (approved):
Workaround

I think the original project has since moved to gitlab. I'll see if we can do an in place upgrade.

mmahmoudian:
We should remove the Ubuntu 20.04 as it is EOL by the end of this month, but I can't figure out why the Debian 11 CIs are failing

borgmanJeremy:
Looks like Debian 11 uses Cmake 3.18. If that’s a valid target I can lower it to 3.18 and see if the CI passes
 Sent from Proton Mail for iOS 
On Sun, Apr 6, 2025 at 04:15, Mehrad Mahmoudian ***@***.***&gt; wrote:   We should remove the Ubuntu 20.04 as it is EOL by the end of this month, but I can't figure out why the Debian 11 CIs are failing—
Reply to this email directly, view it on GitHub, or unsubscribe.
You are receiving this because you authored the thread.Message ID: ***@***.***&gt;  
  mmahmoudian left a comment  (flameshot-org/flameshot#3901)
 We should remove the Ubuntu 20.04 as it is EOL by the end of this month, but I can't figure out why the Debian 11 CIs are failing —

mmahmoudian:
@borgmanJeremy Thanks for checking it out. It was bugging me why it fails and I couldn't figure it out.

Issue #3208, opened 2023-05-22
Allow more than 50 items (>500) in upload history


jesuslalvaro:
### Feature Description

Thanks for this great tool

Ref: https://github.com/flameshot-org/flameshot/issues/3127#issuecomment-1540029508

Config Interface does allow you to select up to 50 elements only

Configuration -> General ->
Latest Uploads Max Size

Because imgur  does not include uploads in your collections, ( anonymous), 
It would be very useful to locally register much more uploadas , even your full uploads history.

Thanks

Amangupta0000:
hello sir! this is my first time in my life contributing for a project. So, please can you assign this task to me.
and Please can you tell  me , how many max item to be set in uoload history.

Thankyou!!

mmahmoudian:
@Amangupta0000 Thanks for considering contributing to this project. I don't think the number actually affects anything, and I don't think anything more than 250 would be practically useful to anyone. The ">500" threshold suggested by @jesuslalvaro would imho just clutter the UI for 99% of the users only to benefit 1% remaining users.

I believe the best course of action is showing the latest 50, and then at the bottom of the list we add a "Show more" button which expands the list to 500 or 1000. This would allow users to delete images that are uploaded way back in their history, which in my book is a good thing to give control to the user.

@veracioux what do you think?

veracioux:
@Amangupta0000 I completely agree with mmahmoudian.

mael-hf:
Hello, I'd like to help with this issue if that's okay. I can see that someone's assigned but nothing has happened for over two years, so I assume that I can work on this.

serebryakov02:
Hello, I would like to help with this issue if it is still relevant.

I checked the current `master` branch and it looks like the upload history setting is still capped at 50 items in the configuration UI. I also saw the previous discussion suggesting that instead of simply raising the visible limit to 500+, the preferred behavior may be to show the latest 50 items first and add a “Show more” button to expand the list to a larger amount, such as 500 or 1000.

Before I start, could you please confirm whether that is still the desired approach? If yes, I would be happy to work on a PR for it.

If this was already solved elsewhere or is no longer wanted, maybe this issue can be closed to avoid confusion.

borgmanJeremy:
Hello this is not relevant with the removal of uploader. Thanks though!

PR #3907, opened 2025-04-13
add loadMore button in Upload History widget (issue #3208)

mael-hf:
- fixes issue #3208 : https://github.com/flameshot-org/flameshot/issues/3208
- I created a new storage called secondaryStorage, which keeps in memory the excess images.
- the total number of images stored is 500 (as defined in src/utils/history.cpp and proposed in the issue).
- the loadMore button is always at the bottom of the UI, except when there aren't anymore images to load, then it disappears.

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -535,7 +535,7 @@ void GeneralConf::initUploadHistoryMax()
     box->setLayout(vboxLayout);
 
     m_uploadHistoryMax = new QSpinBox(this);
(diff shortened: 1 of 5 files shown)


borgmanJeremy:
Hello,

I started a review / testing of this. Would you be able to rebase on master? There was a minor conflict, but something in the qt6 / cpp20 transition made the calls to ```std::min``` unhappy. int and qsizetype are not compatible. 

Once that is done I'll do some local testing and merge (if no issues).

Thanks and sorry for the long delay.

Issue #3399, opened 2023-11-03
USE_WAYLAND_GRIM is not exposed via ccmake


LouisLambda:
### Flameshot Version

v12.1.0

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Fedora Linux 38 

### Description

3ededae5 added  cmake options `USE_WAYLAND_GRIM` and `USE_WAYLAND_CLIPBOARD`, but did not expose USE_WAYLAND_GRIM as an option in `CMakeLists.txt`, so it does not show up via the ccmake TUI and I get confused.

Grim is a screenshot mechanism provided by wlroots-based compositors (like sway).

### System Information

Wayland/Sway

PR #3919, opened 2025-05-03
Replace USE_WAYLAND_GRIM build option with useGrimAdapter setting

q4a:
Since there is already the `disabledGrimWarning` setting, I suggest to replace USE_WAYLAND_GRIM build option with `useGrimAdapter` setting.
People that uses wlroots based Wayland compositors (labwc in my case) will be able to use grim without rebuild flameshot, but just adding line `useGrimAdapter=true` to `~/.config/flameshot/flameshot.ini`.
I don't see any significant negative effects:
- the current behavior does not changed
- grim is still optional
- even if `useGrimAdapter=true`, but grim is not installed, there is message:
`The universal wayland screen capture adapter requires Grim as the screen capture component of wayland. If the screen capture component is missing, please install it!`

The only thing - you will get new setting insted of new build option, but this will lead to more flexibility.
It will also be possible to close this issue: https://github.com/flameshot-org/flameshot/issues/3399
- option `USE_WAYLAND_CLIPBOARD` was added in this commit: https://github.com/flameshot-org/flameshot/commit/00685369a0e0e7f438e051e42119e0360ab84a76
- `USE_WAYLAND_GRIM` was removed in this RP.

### DIFF
--- a/flameshot.example.ini
+++ b/flameshot.example.ini
@@ -51,6 +51,9 @@
 ;; Whether the tray icon is disabled (bool)
 ;disabledTrayIcon=false
 ;
+;; Use grim-based wayland universal screenshot adapter
+;useGrimAdapter=false
+;
 ;; Disable Grim Warning notification
 ;disabledGrimWarning=true
 ;
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -220,15 +220,8 @@ target_link_libraries(
 )
 
 if (USE_WAYLAND_CLIPBOARD)
-  if(!USE_WAYLAND_GRIM)
-    message(WARNING "You activated the USE_WAYLAND_CLIPBOARD option, but did not activate the USE_WAYLAND_GRIM option. Flameshot will use the dbus protocol to support wayland. If you use wlroots-based wayland, it is recommended to enable USE_WAYLAND_GRIM")
-  endif()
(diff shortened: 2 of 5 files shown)


b0ch3nski:
I think this patch makes great sense :+1:

I'm using sway (which is based on wlroots) and explicitly do not use grim since it doesn't handle the multimonitor factorial scaling very well. Current approach puts the decision responsibility at distro package maintainers while this should be user configurable.

jack9603301:
Very good suggestion, I fully support it, this will provide a higher level of flexibility, to ensure compatibility, we can keep both at the same time and then gradually migrate

flameshot.example.ini
```diff
@@ -48,6 +48,9 @@
 ;; Whether the tray icon is disabled (bool)
 ;disabledTrayIcon=false
 ;
+;; Use grim-based wayland universal screenshot adapter
+;useGrimAdapter=true
```
borgmanJeremy, line 52: Can you set this to =false to reflect the initial value in the code?

borgmanJeremy (changes requested):

flameshot.example.ini
q4a: Rebased on top of master branch and use `useGrimAdapter=false` in flameshot.example.ini

borgmanJeremy (approved):

PR #3942, opened 2025-05-14
Add config cli option to enable or disable notifications

borgmanJeremy:


mmahmoudian:
should we add an entry to `flameshot.example.ini`?

borgmanJeremy:
It's already there: https://github.com/flameshot-org/flameshot/blob/master/flameshot.example.ini#L43

This is just exposing the existing option over the cli

mmahmoudian:
Interesting! I used ripgrep to find it in the local clone of the repo but I couldn't find it. Thanks for the link 👍

mmahmoudian (approved):

PR #3950, opened 2025-05-24
Move ARM CI to a new CI

mmahmoudian:
The ARM CIs take 10 times more time to complete compared to AMDs, so it would be more efficient and logical to move them out and only run them on master (after merge) rather than on Pull Requests.

image


mmahmoudian:
It seems all CIs that are meant to run on PRs passed. The ARM CIs are only meant to be run on master, so they didn't run. I will push a temporary commit to add them back to PR just to test if it works, and then I will revert this temporary change.

PR #3952, opened 2025-05-25
Remove Ubuntu 18.04 and 20.04 from documentation

mmahmoudian:
Both 18.04 and 20.04 Ubuntus are EOL already, so there is no point in keeping them in our documentation.


borgmanJeremy (approved):

borgmanJeremy:
Thanks!

Issue #3671, opened 2024-07-18
Flameshot Crashes on start in MacOS


whnr:
### Flameshot Version

12.1.0

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

MacOS 14.4.1 (23E224)

### Description

Flameshot does not start anymore. The console log (full excerpt below) shows this:
```
error	11:20:05.156830-0400	flameshot	SingleApplication: Unable to create block.
error	11:20:05.157023-0400	flameshot	SingleApplication:  7 "QSharedMemory::create: out of resources"
```

I tried reinstalling with homebrew but it failed.

### Steps to reproduce

Try to start flameshot.

Here's the system console log of the whole thread. There are two errors at the bottom.

````
default	11:20:04.965243-0400	flameshot	[0x600002eac000] activating connection: mach=true listener=false peer=false name=com.apple.cfprefsd.daemon
default	11:20:05.010761-0400	flameshot	[0x600002ea81e0] activating connection: mach=true listener=false peer=false name=com.apple.distributed_notifications@Uv3
default	11:20:05.012188-0400	flameshot	No persisted cache on this platform.
default	11:20:05.014791-0400	flameshot	[0x600002ea83c0] activating connection: mach=true listener=false peer=false name=com.apple.tccd.system
default	11:20:05.026776-0400	flameshot	[0x600002ea83c0] invalidated after the last release of the connection object
default	11:20:05.026985-0400	flameshot	server port 0x00002d0f, session port 0x00002d0f
default	11:20:05.037270-0400	flameshot	[0x600002ea83c0] activating connection: mach=true listener=false peer=false name=com.apple.coreservices.launchservicesd
default	11:20:05.037812-0400	flameshot	[0x600002ea85a0] activating connection: mach=true listener=false peer=false name=com.apple.analyticsd
default	11:20:05.039612-0400	flameshot	New connection 0x111c8f main
default	11:20:05.039764-0400	flameshot	Current system appearance, (HLTB: 1), (SLS: 0)
default	11:20:05.039839-0400	flameshot	Current system appearance, (HLTB: 1), (SLS: 0)
default	11:20:05.040739-0400	flameshot	Post-registration system appearance: (HLTB: 1)
default	11:20:05.044190-0400	flameshot	[0x600002ea00f0] activating connection: mach=true listener=false peer=false name=com.apple.pasteboard.1
[...]

PR #3962, opened 2025-05-30
Fixing deprecation for future Qt 6 support

ElTh0r0:
This PR fixes several deprecations as preparation for potential Qt 6 support in future. The code should be compatible with Qt 5 and 6 (at least it is on my machine on Manjaro Linux with Qt 5.15 and Qt 6.9) - let's see the results of the GH actions.

The only major blocker for compiling with Qt 6 is QDesktopWidget / QApplication::desktop(). I didn't had time to dig into that yet, but I marked the corresponding lines with 'ToDo' entries.

Note: When commenting out the lines relating to QDesktopWidget and successfully compiling the app with Qt 6, it crashes when attempting to take a screenshot. But I couldn't really start debugging the root cause, because I could start the app only once - maybe the same/related issue to #3671 - but I am quite sure it is related to 'SingleApplication'.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,20 +37,13 @@ set(PROJECT_NAME_CAPITALIZED "Flameshot")
 include(FetchContent)
 
 #Must be set before fetching external content
-option(USE_QT6 "Enable building aginst Qt6" OFF)
 #QT_DEFAULT_MAJOR_VERSION used by "SingleApplication"
 #QT_VERSION_MAJOR used by Flameshot and "QtColorWidgets"
-if (USE_QT6)
-  set(QT_DEFAULT_MAJOR_VERSION 6 CACHE STRING "")
-  set(QT_VERSION_MAJOR 6 CACHE STRING "")
-  message(WARNING "Attention: The Qt 6 build is still in development, so a properly functioning application cannot be guaranteed!")
+set(QT_DEFAULT_MAJOR_VERSION 6 CACHE STRING "")
+set(QT_VERSION_MAJOR 6 CACHE STRING "")
 
-  #Needed due to linker error with QtColorWidget
-  set(CMAKE_POSITION_INDEPENDENT_CODE ON)
-else()
-  set(QT_DEFAULT_MAJOR_VERSION 5 CACHE STRING "")
-  set(QT_VERSION_MAJOR 5 CACHE STRING "")
-endif()
+#Needed due to linker error with QtColorWidget
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 FetchContent_Declare(
     qtColorWidgets
@@ -131,18 +124,13 @@ set(QAPPLICATION_CLASS
         QApplication
         CACHE STRING "Inheritance class for SingleApplication")
 
-if(USE_EXTERNAL_SINGLEAPPLICATION AND NOT USE_QT6)
+if(USE_EXTERNAL_SINGLEAPPLICATION)
   # look for external QtSingleApplication
   # package dev-qt/qtsingleapplication provides no symlink to current version
   set(qtsingleapplication_libs libQt5Solutions_SingleApplication-2.6 Qt5Solutions_SingleApplication-2.6)
   find_library(QTSINGLEAPPLICATION_LIBRARY NAMES ${qtsingleapplication_libs})
   message(STATUS "Using external QtSingleApplication library")
 else()
-  if(USE_EXTERNAL_SINGLEAPPLICATION AND USE_QT6)
-    message(WARNING "External QtSingleApplication library is not available for Qt 6, switching alternative SingleApplication")
-    set(USE_EXTERNAL_SINGLEAPPLICATION OFF)
-  endif()
-
   FetchContent_Declare(
       singleApplication
       GIT_REPOSITORY https://github.com/itay-grudev/SingleApplication.git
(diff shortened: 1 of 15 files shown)


borgmanJeremy:
Wow great work overall! I was recently porting this to Qt6 again and this will save a lot of time. In the past (and I suspect again) the main blocker will be the Qt Single Application bug. The only work around I have thought of is to just remove the single application feature. IE if the daemon is running and a user invokes flameshot from the CLI a 2nd instance will launch. IMO that 100x better than needing to reboot your PC after a crash.

I will need to take an in depth look at the PR but some initial thoughts:

1.) For the issue with QDesktop see this file: https://github.com/flameshot-org/flameshot/blob/a9cfb004b5b77a4a9ddad2a2069e7ab8a4cf38df/src/utils/screengrabber.cpp 

It's from the qt6 branch in the main flameshot repo. It might be a good branch to look at for anything you are stuck on.

2.) The CMakefiles need to be updated for Qt6, I have a local (not pushed branch) with most of these changes I can share if that will help. Confusingly it's my latest from scratch Qt6 port so it wont be compatible with the existing Qt6 branch

3.) IMO we should only support Qt6 but I want @mmahmoudian  to weigh in. I feel like we have accumulated lots of tech debt and dropping legacy stuff will help. Debian 11 does not have Qt6, but we can support that with flatpak and appimage. Debian 12 does have Qt6

4.) The CI needs a large upgrade for Qt6 (it's passing now since the cmake isn't defaulting to Qt6, I can also help on this)

5.) All the packaging files need updated. I'm okay merged BEFORE all that is fixed and putting a call out to the community for help with that. 


Again, thanks a lot for this PR! If you want any specific help on the CI or cmake updates let me know! I will take a closer look at the other changes this weekend

mmahmoudian:
@borgmanJeremy 
> 3.) [...] IMO we should only support Qt6 but I want @mmahmoudian to weigh in. I feel like we have accumulated lots of tech debt and dropping legacy stuff will help.

I also believe that there is no point in keeping Qt5 around as we are actually quite late at the Qt5 -> Qt6 party and major distros already have Qt6.

> 5.) [...] I'm okay merged BEFORE all that is fixed and putting a call out to the community for help with that.

The only group that is taking our HEAD of master is the AUR users. This group have helped us test Flameshot and get feedback before we have a release candidate or a full release. So maybe it is worth having some of the bells and whistles in before merging so that the master can be built. As far as I understand, CMake is the only part that needs fixing?



Overall this is a great PR to the correct direction. The only question I have is how hard it is to port the open PRs to Qt6? We have 44 other open PRs 6 of which are in v13 milestone.

borgmanJeremy:
@mmahmoudian  OK great. For the AUR the pkgbuild will need some updates but I can make those changes. 

Since both @ElTh0r0  and I will be working on this in parallel I created a new branch 'qt6_2025' @ElTh0r0  Can you please change this PR to  merge into that branch? Then you and I can both submit PR's there without affecting the master branch. I will focus on CMake / CI changes, and you can focus on  the remaining QDesktop issues and fully remove the Qt5 #ifdefs

@mmahmoudian what are your thoughts on the SingleApplication trade off?

I will also try to get to the 6 PR's in the v13 milestone. I'm very keen to finish this port because in theory lots of bugs related to scaling should be fixed. At a quick glance none of those 6 PR's will have many (if any) conflicts.

ElTh0r0:
Many thanks for your kind feedback!

> what are your thoughts on the SingleApplication trade off?

I didn't dig into that topic, but quick online search brought me to [KDSingleApplication](https://www.kdab.com/kdsingleapplication/) from KDAP. Not sure, if it is a better implementation than the other two from Flameshot. I'll leave the decision and discussion up to you, since you know the Flameshot internals much better than I do.

> Since both @ElTh0r0 and I will be working on this in parallel I created a new branch 'qt6_2025' @ElTh0r0 Can you please change this PR to merge into that branch? Then you and I can both submit PR's there without affecting the master branch. I will focus on CMake / CI changes, and you can focus on the remaining QDesktop issues and fully remove the Qt5 #ifdefs

Sounds great, will do!

mmahmoudian:
@borgmanJeremy 

> @mmahmoudian what are your thoughts on the SingleApplication trade off?

I believe I don't quite understand the issue. Do I understand correctly your point that if a Flameshot process already exists, at the moment invoking another Flameshot will not do anything, but with your proposed method we every new run of Flameshot will start a new process?

If so, would it create a new instance if the user do `flameshot gui`? Or would that communicate through the DBus with the existing process?

mmahmoudian:
@borgmanJeremy 

I see, thanks for the detailed explanation. Now I'm the same page.

> I propose removing the SingleApplication dependency until the bug is resolved. Then if you run flameshot gui from the terminal and flameshot is already open, it will just launch a second instance.

So removal of SingleApplication will not affect dbus, right? Linux has dbus, macOS [can have dbus](https://formulae.brew.sh/formula/dbus), and apparently "[the Windows port from the windbus and dbus4win projects was merged into the freedesktop dbus development branch several years ago](https://www.freedesktop.org/wiki/Software/dbus/#windowsport)". Doesn't this mean dbus is already covering all major platforms we support?


Anyways, I suppose this removal of SingleApplication is inevitable until the day that it is fixed in Qt6. And if we can retain the same/similar functionality using dbus, then we can just add to the documentation this information to install dbus if user want to have Flameshot running as daemon.

borgmanJeremy (approved):

PR #3966, opened 2025-06-01
Fix further Qt 6 deprecations

ElTh0r0:
Some further Qt 6 deprecations fixes

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -33,36 +33,6 @@
 #include <desktopinfo.h>
 #endif
 
-#ifdef Q_OS_LINUX
-// source: https://github.com/ksnip/ksnip/issues/416
-void wayland_hacks()
-{
-    int suffixIndex;
-    DesktopInfo info;
-
-    const char* qt_version = qVersion();
-
-    QVersionNumber targetVersion(5, 15, 2);
-    QString string(qt_version);
-    QVersionNumber currentVersion =
-      QVersionNumber::fromString(string, &suffixIndex);
-
-    if (currentVersion < targetVersion) {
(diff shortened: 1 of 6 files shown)


borgmanJeremy:
@ElTh0r0 @mmahmoudian 

This PR cleanly merged with the changes I also have made. The branch now :  
- compiles
- arch PKGBUILD builds
- singleapplication is temporarily commented out
- SEGFAULT is fixed

That is to say, it nominally "works" but needs way more testing.

mmahmoudian:
I will use it in my daily work on Plasma Wayland and report back.

ElTh0r0:
>     * SEGFAULT is fixed
Great! I am really confused how it worked with Qt 5.
https://github.com/flameshot-org/flameshot/blob/88d6e14246bcd2d4be4a0cdf03b3f14c4967875f/src/widgets/capture/capturewidget.cpp#L210-L222

'm_colorPicker->hide()' emits 'ColorPicker::colorSelected' from above connect and triggering 'CaptureWidget::setDrawColor' and in one of the following subroutines it tries to access 'm_panel' which is first created in 'initPanel()'... Maybe the connect was executed later with Qt 5?! 🫤

borgmanJeremy:
@ElTh0r0  Yeah i had read that in Qt6 the timing of connect() and constructors was changed so I tried a fast hack to move all the connect() after the constructors.

This was a quick test that worked and I wanted to push it to unblock you. I want to take a closer look at all our constructors and check them all. I also want to read through more official Qt documentation to make sure I full understand the proper sequence of constructor elements.

PR #3973, opened 2025-06-03
allow networking during build phase so cmake can clone dependencies

borgmanJeremy:
This bug was introduced because the CI builds from master branch. So if a commit breaks flatpak, it wont be detected until after its merged to master branch.


borgmanJeremy:
It also means this won't appear to fix the problem, but I tested it locally.

PR #3976, opened 2025-06-04
Qt6 Windows build

ElTh0r0:
Updating runner to windows-2025 with msvc2022
See successful run of the GH action in [my branch](https://github.com/ElTh0r0/flameshot/actions/runs/15450541033/job/43491277425)

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -86,46 +86,46 @@ jobs:
         # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
         run: ctest -C $BUILD_TYPE
 
-#   windows-build:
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
Terrific,

That means Windows, Flatpak, and PKGBUILD are all working. I might mess with some more packages tonight. @mmahmoudian  I think we have some cleanup to do, but this is probably close to merge to master unless you enounter new bugs in testing.

borgmanJeremy:
In any case I'll merge this to the qt6_2025 branch tonight so qt6_CI can be short lived and we don't have too many divergent branches.

mmahmoudian:
Perfect. I'll do more testing today and will report back.

We can get the first round of feedback from AUR users right after merging to master.

PR #3977, opened 2025-06-04
Port to QT6 on MacOs

borgmanJeremy:
@mmahmoudian  FYI i compiled and tested this on my mac and it works. Ill continue to address the CI on the other branch, but now we have working builds for QT6 on all platforms.
test


mmahmoudian:
@borgmanJeremy wow, this is great ! Thank you and @ElTh0r0 for all the efforts

PR #3993, opened 2025-06-09
Qt6 2025

borgmanJeremy:
@mmahmoudian 

This is a big one so I'll wait for you and probably @ElTh0r0  to agree to merge to main. AFAIK there are no new defects introduced (other than non building packages). I'd like to get this merged in so everyone starts working from the same reference. 

Lot's of bugs to squash before the v13 release, but this will be huge milestone.


mmahmoudian:
@borgmanJeremy 
I've been using `qt6_2025` branch have have not observed any issues that are not present in yhe `master`, so I think merging it would not only be safe, but as you said beneficial as well. Regarding the Flatpak and Snap builds that fail, I don't think anyone is using nightly build of those, so that will not break anyone's workflow either. Regarding deb package and others that you commented out in another PR, I think we can move forward with this PR and fix themin future PRs.

Thank you to both you and @ElTh0r0 for pushing this forward 🍻

ElTh0r0:
I fully agree merging to main; I see no blocker for this step and polishing & bug fixing can happen again in one central place 👍

borgmanJeremy:
Thanks @ElTh0r0  for all the work and @mmahmoudian for all the testing!

Issue #3990, opened 2025-06-09
Fixed screenshot path is overwritten with filename after first save


vladrevers:
### Flameshot Version

Flameshot v12.1.0 (a1dda591)
Compiled with Qt 5.15.2
winnt: 10.0.26100
windows: 10

### Installation Type

Compiled from source (from [Github Actions](https://github.com/flameshot-org/flameshot/actions/runs/15513092898/job/43676432923))

### Operating System type and version

Windows 11 (24H2)

### Description

When saving screenshots with "Use fixed path for screenshot to save" enabled, an error occurs after the first save.

**Minimal configuration to reproduce the issue:**

```
[General]
savePath=C:/Users/Hide/Pictures
savePathFixed=true
```

After the first save, the configuration changes — the save path is updated to include the filename of the previously saved screenshot. On subsequent saves, an error occurs stating that the save path is not a directory:
```
savePath=C:/Users/Hide/Pictures/Screenshot_2025-06-09_23-26-47.png
```

### Steps to reproduce

1. Enable "Use fixed path for screenshot to save" in settings.
2. Press PrtSc to open Flameshot.
3. Press Ctrl+A to select the entire area.
4. Press Ctrl+S to save the screenshot.
5. Press PrtSc again.
6. An error appears because savePath was overwritten with the previous screenshot’s filename.

### Screenshots or screen recordings

![Image](https://github.com/user-attachments/assets/4ef5def4-8530-4734-997c-d876acec4aba)

### System Information

Windows 11
Asus Laptop, 1920x1080

borgmanJeremy:
@ElTh0r0  would you be able to tackle this one? I can't replicate it on linux. If I had to guess the error is here: https://github.com/flameshot-org/flameshot/blob/4edfb2ac1d71e7f75fcdcb850ff6bce5fb148a7b/src/utils/screenshotsaver.cpp#L266C1-L266C49

And maybe on windows the path is not parsing correctly.

ElTh0r0:
@borgmanJeremy I'll take a look 👍

borgmanJeremy:
Fixed with #3997

PR #3997, opened 2025-06-10
Fix for wrong set savePath on Windows (#3990)

ElTh0r0:
Fix for wrong set savePath on Windows (#3990). As Qt is always using '/' internally, we must avoid QDir::separator() here, as it returns '\\' on Windows and because of this the correct path cannot be extracted.


borgmanJeremy:
Thanks! I tested this on linux and it didn't break anything there.

Issue #3979, opened 2025-06-05
Changing opacity in pinned images errors out


mmahmoudian:
### Flameshot Version

Flameshot v12.1.0 (94199582)
Compiled with Qt 5.15.17

Flameshot v12.1.0 (b504b9e)
Compiled with Qt 6.9.1

### Installation Type

AppImage

### Operating System type and version

EndeavourOS - 6.12.31-1-lts (64-bit)

### Description

After pinning the screenshot, I cannot change the opacity of the pinned image.

### Steps to reproduce

1. run `flameshot gui`
2. pin a screenshot
3. rightclick on the pinned image
4. click on "Decrease Opacity"
5. observe the following error:
    > This plugin does not support setting window opacity

The following is the output of the Flameshot daemon:

```console
❯  flameshot
Unable to get current screen, starting to use primary screen. It may be a cause of logical error and working with a wrong screen.
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
This plugin does not support setting window opacity
```

### System Information

```console
❯  inxi --system --graphics --width 80
System:
  Host: pamp-precision-tower Kernel: 6.12.31-1-lts arch: x86_64 bits: 64
  Desktop: KDE Plasma v: 6.3.5 Distro: EndeavourOS
Graphics:
  Device-1: Intel CometLake-S GT2 [UHD Graphics 630] driver: i915 v: kernel
  Device-2: Advanced Micro Devices [AMD/ATI] Lexa XT [Radeon PRO WX 3200]
    driver: amdgpu v: kernel
  Display: wayland server: X.org v: 1.21.1.16 with: Xwayland v: 24.1.6
    compositor: kwin_wayland driver: X: loaded: amdgpu,modesetting
    dri: iris,radeonsi gpu: i915 resolution: 1: 1920x1080~60Hz
    2: 1920x1080~60Hz
  API: EGL v: 1.5 drivers: iris,radeonsi,swrast
    platforms: gbm,wayland,x11,surfaceless,device
  API: OpenGL v: 4.6 compat-v: 4.5 vendor: intel mesa v: 25.1.2-arch1.1
    renderer: Mesa Intel UHD Graphics 630 (CML GT2)
  API: Vulkan v: 1.4.313 drivers: radv surfaces: N/A
  Info: Tools: api: clinfo, eglinfo, glxinfo, vulkaninfo
    de: kscreen-console,kscreen-doctor wl: wayland-info
[...]

PR #4000, opened 2025-06-11
make pin moving compatible with Wayland on Qt6

borgmanJeremy:
First step in fixing #3979. 

Simply fixing the opacity looked terrible with the border added on Wayland. When I first fixed the border on wayland the window became unmovable. This is because on wayland DE's Qt is not able to move non child widgets. So we now request the DE to move for us.

I've only tested on arch with Gnome and KDE. I want to test this on MacOs, and some X11 desktops before merging. If required we can add some if statements to restore the old behavior if the new implementation does not work on all platforms.

See here for more details:

https://doc.qt.io/qt-6/qwindow.html#startSystemMove
https://doc.qt.io/qt-6/qwindow.html#setPosition
https://stackoverflow.com/questions/76665812/qwindow-startsystemmove-stops-hover-events-on-child-widgets

When this is merged I can then add the real opacity fix. 

Tested:
- [x] Wayland KDE
- [x] Wayland Gnome
- [x] X11 KDE
- [x] MacOS
- [x] windows

### DIFF
--- a/src/tools/pin/pinwidget.cpp
+++ b/src/tools/pin/pinwidget.cpp
@@ -148,9 +148,6 @@ void PinWidget::mouseDoubleClickEvent(QMouseEvent*)
 
 void PinWidget::mousePressEvent(QMouseEvent* e)
 {
-    m_dragStart = e->globalPosition();
-    m_offsetX = e->position().x() / width();
-    m_offsetY = e->position().y() / height();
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
I finally have a working windows VM setup again. This breaks the pin widget on Windows so it will need to #ifdef the old implementation on Windows builds.

PR #4004, opened 2025-06-11
fixing windows CMAKE CI

borgmanJeremy:


borgmanJeremy:
Woot no more broken CI (still need to re-enable some linux packages)

Issue #4003, opened 2025-06-11
Remove .ts file from triggering CI


mmahmoudian:
The .ts files are translation files which are typically passed to this repo from the translation platform we use (i.e., Weblate). These files do not affect how Flameshot builds, so it make sense to exclude them from the CIs.

PR #4005, opened 2025-06-12
Make CIs to ignore .ts translation files in PullRequests

mmahmoudian:
Addresses #4003

In all relevant CIs the .ts files are ignored during the PR step. The only exception was the `clang-format.yml` in which I though it should also be unnecessary to run it on master HEAD when a translation file is changed. If that is wrong or ano further modification is needed, let me know and I will add more commits.


borgmanJeremy (approved):
Thanks!

Issue #3812, opened 2025-01-12
Config Option: Default Active Tool Size


sneukam:
### Feature Description

Add a configuration setting in the GUI for the Default Active Tool Size.

For a while, I would have to adjust the Active Tool Size when adding paint tool stuff to the screenshot (lines, boxes, arrows, text, etc). Manually moving the slider worked, but the slider reset after every tool paint.

I eventually stumbled on using shortcuts to modify active tool size (instead the slider) modified the active tool size permanently, even between screenshots.

It would make a lot of sense to have a configuration in the GUI to set the default active tool size. Repeatedly manually moving the slider to increase the active tool size multiple times when editing a screenshot was frustrating. I'm glad I stumbled on a solution, but I think it would be more clear for users if they were able to set this manually, rather than accidentally discovering it if they happen to use a shortcut.

ElTh0r0:
My thoughts on this topic: I think it would be better if the **currently set size is retained** when switching to another tool (and is also reloaded when Flameshot is restarted).
Workflow I have in mind: If, for example, you want to draw several frames, arrows, lines, etc., they should (very often) have the same size for _the current_ screenshot - and the size should not jump back to a set value from the settings every time you change tools.

I can try to take a look on this approach.

mmahmoudian:
I believe it have been suggested before to have different sizes set for different tools. For instance the arrow, line and frame that draw line use the same thickness, but counting tool or pixelation have their own sizes. The logic was that blur in pixel tool happens when size is minimum, but the arrow would need something around size 5 to be visible enough. I personally use a very high value for pixel, but relatively small value for counting bubbles.

ElTh0r0:
Valid points!
So we have the following tools: Pencil, Line, Arrow, Rectangular Selection, Rectangle, Circle, Marker, Text, Circle Counter, Pixelate, Invert.

Proposal:
- Should share the same selected size: Line, Arrow, Rectangular Selection, Circle, Pencil (?)
- Should have its own size each: Text, Circle Counter, Pixelate, Rectangle (as size is used for rounded corners), Marker (?)
- Out of scope: Invert, as it is not using size

I am not sure about the Pencil and Marker. What is your opinion?
And do you agree in not having a configuration option for the sizes, but just retaining the last size value (even after restart of the application)?

mmahmoudian:
> Proposal:
> 
> - Should share the same selected size: Line, Arrow, Rectangular Selection, Circle, Pencil (?)
[...]

PR #4011, opened 2025-06-16
Separate tool size for some of the tools

ElTh0r0:
As discussed in #3812, it would be useful to have separate size for Text, Circle Counter, Pixelate, Rectangle (as the size is used for rounded corners) and Marker. All other tools (Pencil, Line, Arrow, Rectangular Selection, Circle) are still sharing the same selected size.


borgmanJeremy (approved):

borgmanJeremy:
Thank you!

closes #3812

Issue #3824, opened 2025-01-25
mac arm apple silicon support


guyulmaz:
### Feature Description

Hi ,

do you plan to add mac arm apple silicon support

borgmanJeremy:
Hello, I am no longer a maintainer of this project. I provide limited contributions at this point. I appreciate your contribution!, but I'm not in a place to take a look at this right now.

holazt:
Now, macos arm64 should be support https://github.com/flameshot-org/flameshot/pull/3867

holazt:
> [@borgmanJeremy](https://github.com/borgmanJeremy) i thought to [give it a try](https://github.com/electriquo/flameshot/pull/2) by using [workflow matrix strategy](https://github.com/electriquo/flameshot/actions/runs/13204878464). although the builds pass, the artifact produced on macos arm64 do not open on macos arm64. would you mind to take a look?

![](https://github.com/user-attachments/assets/f63366a5-1996-4cdf-9e2a-7def45520612)
That dmg package is not signed, maybe that's the problem.

holazt:
> [@mmahmoudian](https://github.com/mmahmoudian) [@holazt](https://github.com/holazt) although it was [merged](https://github.com/flameshot-org/flameshot/issues/3824#issuecomment-2708629828), the [issue persists](https://github.com/flameshot-org/flameshot/issues/3824#issuecomment-2660838704) in the packages within https://github.com/flameshot-org/flameshot/actions/runs/13758971667
> 
> > the artifact produced on macos arm64 do not open on macos arm64
> 
> so we are back to square one :(

Did you have install arm64 build successfully by changing security settings? Do you have any error log or screenshot?

mmahmoudian:
@electriquo the "do not open" is generally not a useful phrase to help resolving the issue. Do you get any error? do you get any outputs? Also, as @holazt correctly pointed out, please make sure the issue is not related to unsigned dmg. You can find instructions about resolving the signature issue here: https://flameshot.org/docs/installation/installation-osx/

borgmanJeremy:
@mmahmoudian  here is the error from the package built in the CI.  This is different from the security warning due to no code signing. 

<img width="323" alt="Image" src="https://github.com/user-attachments/assets/e1467e41-f90b-40db-aa30-44462550b30c" />

mmahmoudian:
[...]

PR #4020, opened 2025-06-19
Fix macos signing and clean up CI

borgmanJeremy:


borgmanJeremy:
The .dmg built by this pipeline actually works on apple silicon now!

I need to do some clean up so the version is not hard coded but it's close.

borgmanJeremy:
Closes #3824

PR #4022, opened 2025-06-19
Search windeployqt

ElTh0r0:
As mentioned in https://github.com/flameshot-org/flameshot/discussions/4012#discussioncomment-13504609, windeployqt wasn't found anymore and because of this the Windows artifacts were missing the Qt DLLs.
Now cmake searches in CMAKE_PREFIX_PATH first and (as fallback / for backward compatibility) in environment variable QTDIR for the windeployqt.exe

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -391,7 +391,19 @@ endif ()
 
 # windeployqt
 if (WIN32)
-    if (EXISTS $ENV{QTDIR}/bin/windeployqt.exe)
+    # Add CMAKE_PREFIX_PATH to search paths
+    foreach(prefix ${CMAKE_PREFIX_PATH})
+        list(APPEND WINDEPLOYQT_SEARCH_PATHS "${prefix}/bin")
+    endforeach()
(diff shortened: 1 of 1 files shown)


borgmanJeremy (approved):

borgmanJeremy:
Did you happen to test the artifacts? BTW our official release for windows comes from the appveyor build since that is the only way to sign into SignPath. Those builds _did_ seem to have the qt6 dlls. I'll check them again after this commit goes through the SighPath pipeline.

ElTh0r0:
> Did you happen to test the artifacts? BTW our official release for windows comes from the appveyor build since that is the only way to sign into SignPath. Those builds _did_ seem to have the qt6 dlls. I'll check them again after this commit goes through the SighPath pipeline.

Yes, the artifacts of my branch included the DLLs.
Ah ok, I already wondered about the need for the additional appveyor CI. But I saw that the appveyor.yml sets `QTDIR` env variable, and because of that it always worked and wasn't affected by my change #4008, which removed QTDIR from the github action. All good with the Appvyeor build in my opinion, but please check!

https://github.com/flameshot-org/flameshot/blob/89679f5bb6d29d982261dcf1cddeff67f5a50334/appveyor.yml#L12

Issue #877, opened 2020-09-10
Configuration/about window should appear in the middle of screen


hosiet:
**Is your feature request related to a problem?**
When opening the config window or the about window, users may feel somehow weird:

* The config window always somewhere near the top left corner of current screen.
* The about window seems to be appearing in the middle of current screen but it's actually not: the window appear near the center but not the exact center.

I guess we should unify the appear position of those windows to ease user's interaction with them.

See #318 for my previous (somehow unsuccessful) fix for the window position of the about window.

I believe this problem is quite minor and we can dig into it after 0.8.0 release.

borgmanJeremy:
closed with #4024

PR #4024, opened 2025-06-19
Center dialogs on current screen

ElTh0r0:
Fix for #877 
Note: On my main system (Manjaro with KDE Plasma), the dialog's had been automatically centered by Plasma (without this code change), so potentially other Linux users didn't noticed this issue. But it was an issue on Windows, which is fixed now with this PR.

### DIFF
--- a/src/core/flameshot.cpp
+++ b/src/core/flameshot.cpp
@@ -37,8 +37,8 @@
 #endif
 
 Flameshot::Flameshot()
-  : m_captureWindow(nullptr)
-  , m_haveExternalWidget(false)
+  : m_haveExternalWidget(false)
+  , m_captureWindow(nullptr)
 #if defined(Q_OS_MACOS)
   , m_HotkeyScreenshotCapture(nullptr)
   , m_HotkeyScreenshotHistory(nullptr)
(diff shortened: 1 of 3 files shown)


b0ch3nski:
@ElTh0r0 Could similar logic also be applied to `src/widgets/capture/capturewidget.cpp` on `void CaptureWidget::initQuitPrompt()` function? I've noticed that this popup appears on "counter intuitive" places :smile_cat:

ElTh0r0:
> @ElTh0r0 Could similar logic also be applied to `src/widgets/capture/capturewidget.cpp` on `void CaptureWidget::initQuitPrompt()` function? I've noticed that this popup appears on "counter intuitive" places 😸

Thanks for the feedback, I'll take a look! It is highly possible that I missed further pop-up dialogues that I did not notice due to my different workflows and Flameshot settings.

ElTh0r0:
> > @ElTh0r0 Could similar logic also be applied to `src/widgets/capture/capturewidget.cpp` on `void CaptureWidget::initQuitPrompt()` function? I've noticed that this popup appears on "counter intuitive" places 😸

This dialog is centered now as well

borgmanJeremy (approved):
Thanks, I wasn't sure if this would work on Wayland but its working fine. Did you find a reference for calling show() before moving the window? I wonder if thats a solution to the magnifier bugs.

ElTh0r0:
> Thanks, I wasn't sure if this would work on Wayland but its working fine. Did you find a reference for calling show() before moving the window? I wonder if thats a solution to the magnifier bugs.

perplexity.ai "supported" me at this step, after I was frustrated making so many tries, which didn't aligned correctly 😬
grafik

After I just asked the AI again (so that I could attach the above screenshot) it mentioned another improvement:
grafik
This sounds more elegant. I'll take a look once I have time.

The reference links from the AI are (currently ?) not available [[9]](https://www.qtcentre.org/threads/14772-how-to-show-window-in-the-center-of-the-screen), [[10]](https://qtcentre.org/threads/14772-how-to-show-window-in-the-center-of-the-screen?p=197216)

PR #4029, opened 2025-06-21
Qt 6 - DEB/RPM Linux packaging

ElTh0r0:
My changes so far:
- Removed Debian 11 and opensuse-leap-15.6, as they don't support Qt 6 - if I am not wrong.
- According to the [setup-qemu-action manual](https://github.com/docker/setup-qemu-action) the qemu action must be before Buildx, otherwise Buildx may not use the correct platform architecture.
- Add 'Qt Image Formats' package as optional/recommended
- Update README to Qt 6 (please somebody review it!)

The packages are successfully built except for Ubuntu 22.04 (same for the ARM builds) - I have no idea yet, why it fails for 22.04. If anybody has an idea, would be great! Is there any possibility to get more debug output why the build fails? It seems something is wrong with dependencies, but what?

I tested the Fedora 42 and Ubuntu 24.04 package in a VirtualBox VM. On Ubunut 24.04, the installation and taking screenshots works (after providing screenshot permission).
When trying to install the package in Fedora, I get the following:
```
Failed to resolve the transaction:
Problem: conflicting requirements
  - nothing provides libQtColorWidgets.so.2()(64bit), this is required by flameshot-12.1.0-1.fc42.x86_64 from @commandline
You can try adding the following to the command line:
  --skip-broken to skip uninstallable packages
```

Support is highly welcome 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,13 +443,13 @@ apt install git openssl ca-certificates qt6-image-formats-plugins
 
 ```shell
 # Compile-time
-dnf install gcc-c++ cmake qt5-qtbase-devel qt5-linguist
+dnf install gcc-c++ cmake qt6-qtbase-devel qt6-qtsvg-devel qt6-qttools qt6-linguist
 
 # Run-time
-dnf install qt5-qtbase qt5-qtsvg-devel
+dnf install qt6-qtbase qt6-qtsvg
 
 # Optional
-dnf install git openssl ca-certificates
+dnf install git openssl ca-certificates qt6-qtimageformats
 ```
 
 #### Arch
--- a/README.md
+++ b/README.md
@@ -409,7 +409,7 @@ Also you can open and build/debug the project in a C++ IDE. For example, in Qt C
 
 #### Compile-time
 
-- Qt >= 5.9
+- Qt >= 6.0
   + Development tools
 - GCC >= 7.4
 - CMake >= 3.29
@@ -424,18 +424,19 @@ Also you can open and build/debug the project in a C++ IDE. For example, in Qt C
 - Git
 - OpenSSL
 - CA Certificates
+- Qt Image Formats - for additional export image formats (e.g. tiff, webp, and more)
 
 #### Debian
 
 ```shell
 # Compile-time
-apt install g++ cmake build-essential qtbase5-dev qttools5-dev-tools libqt5svg5-dev qttools5-dev
+apt install g++ cmake build-essential qt6-base-dev qt6-tools-dev-tools qt6-svg-dev qt6-tools-dev
 
 # Run-time
-apt install libqt5dbus5 libqt5network5 libqt5core5a libqt5widgets5 libqt5gui5 libqt5svg5
+apt install libqt6dbus6 libqt6network6 libqt6core6 libqt6widgets6 libqt6gui6 libqt6svg6
 
 # Optional
-apt install git openssl ca-certificates
+apt install git openssl ca-certificates qt6-image-formats-plugins
 ```
 
 #### Fedora
@@ -455,13 +456,13 @@ dnf install git openssl ca-certificates
 
 ```shell
 # Compile-time
-pacman -S cmake base-devel git qt5-base qt5-tools
+pacman -S cmake base-devel git qt6-base qt6-tools
 
 # Run-time
-pacman -S qt5-svg
+pacman -S qt6-svg
 
 # Optional
-pacman -S openssl ca-certificates
+pacman -S openssl ca-certificates qt6-imageformats
 ```
 
 #### NixOS
@@ -474,7 +475,7 @@ nix-shell
 
 First of all you need to install [brew](https://brew.sh) and then install the dependencies
 ```shell
-brew install qt5
+brew install qt6
 brew install cmake
 ```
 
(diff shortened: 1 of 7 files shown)


mmahmoudian:
> When trying to install the package in Fedora, I get the following:

I think this is because of conflicting version numbers of libQtColorWidget which one is shipped with Fedora Plasma and one is required by us?

I need to get behind my main machine and test these in VM. But the PR so far sounds logical. This is great.

ElTh0r0:
I think I found the root cause for the not working Fedora package. With my last changes, I was able to build the package and could successfully install it and use Flameshot in my Fedora 42 VM.
Problem was 'BUILD_SHARED_LIBS', which is used by QtColorWidget. I strongly assume we want to use static build everywhere. Some process is still setting 'BUILD_SHARED_LIBS = ON', but I couldn't figure out from where this is coming from. See here:
grafik

If somebody with more packaging knowledge could check it, this would be great.

CMakeLists.txt from QtColorWidgets
grafik

ElTh0r0:
> The packages are successfully built except for Ubuntu 22.04 

I think I found the issue. I didn't do a test run, but in Debian and Ubuntu 24.04+ the required Qt SVG lib package is called [qt6-svg-dev](https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=qt6-svg-dev&searchon=names), but in Ubuntu 22.04 its name is [libqt6svg6-dev](https://packages.ubuntu.com/jammy/libqt6svg6-dev). Does anybody know a smart way how to handle this?

borgmanJeremy:
What about using alternate package options with the | operator? https://www.debian.org/doc/debian-policy/ch-relationships.html

qt6-svg-dev | libqt6svg6-dev,

ElTh0r0:
> What about using alternate package options with the | operator? https://www.debian.org/doc/debian-policy/ch-relationships.html
> 
> qt6-svg-dev | libqt6svg6-dev,

thanks! I wasn't aware, that this is possible for Build-depends. Now the build works!
The only thing I noticed, which needs correction, is that there are some files included in the DEB package, which shouldn't be part of it:
grafik

I think similar change as you did for the RPM (663a056f78bb57ee55acddb7f31ad7ffdcd2afc6) will be needed for the DEB as well.

borgmanJeremy:
The code changes look good to me. I want to test the artifacts in VM's and can do that this evening. I will track the progress below. 

tested:

- [x] Ubuntu 22.04
- [x] Ubuntu 24.04
- [ ] Ubuntu 25.04
- [x] Debian 12
- [x] Fedora 41
- [x] Fedora 42

To make sure the static lib change did not have a negative effect: 
- [ ] Signed Windows build
- [ ] MacOS Build

ElTh0r0:
I noticed a minor leftover of my change in 677ae26. I didn't delete the empty folders. I just pushed another quick change.
grafik

I successfully tested the Fedora 42 package (with KDE Plasma)

ElTh0r0:
Additonally I just added the kguiaddon to the Readme for Fedora and Arch (referring to #4051)

borgmanJeremy:
FYI ubuntu 22.04 is missing a runtime library. Installing qt6-base-dev resolves the issue but I'm trying to pin down exactly which .so that is provided by this fixes the problem so we can include a slimmer dependency.

ElTh0r0:
> I noticed a minor leftover of my change in [677ae26](https://github.com/flameshot-org/flameshot/commit/677ae265f00495b978b35ee0484eaa2278afc9c5). I didn't delete the empty folders. I just pushed another quick change. grafik

borgmanJeremy:
@ElTh0r0 qt6-qpa-plugins is required for ubuntu. Do you mind adding that so I don't have to submit a PR to your PR :)

Not sure on debian 12 yet, I'll set up a clean vm for it.

README.md
```diff
@@ -424,44 +424,45 @@ Also you can open and build/debug the project in a C++ IDE. For example, in Qt C
...
-apt install git openssl ca-certificates
+apt install git openssl ca-certificates qt6-image-formats-plugins
 ```
 
 #### Fedora
 
 ```shell
 # Compile-time
-dnf install gcc-c++ cmake qt5-qtbase-devel qt5-linguist
+dnf install gcc-c++ cmake qt6-qtbase-devel qt6-qtsvg-devel qt6-qttools qt6-linguist kf6-kguiaddons-devel
```
borgmanJeremy, line 446: Also need: qt6-qttools-devel

packaging/debian/control
```diff
@@ -18,7 +20,7 @@ Package: flameshot
 Architecture: any
 Depends:
```
borgmanJeremy, line 21: Also need: qt6-qpa-plugins

ElTh0r0:
> @ElTh0r0 qt6-qpa-plugins is required for ubuntu. Do you mind adding that so I don't have to submit a PR to your PR :)
> 
> Not sure on debian 12 yet, I'll set up a clean vm for it.

Thanks for digging into that! I added the mentioned dependencies.

Note: AppImage is still disabled, I didn't had time to look at that. But I'd say this can be enabled in another PR for not delaying this PR once everything is fine.

borgmanJeremy:
Thanks for all the work on this!!

PR #4030, opened 2025-06-21
Enable Singleapplication again

ElTh0r0:
As mentioned in the [v13 Readiness discussion](https://github.com/flameshot-org/flameshot/discussions/4012#discussioncomment-13502680), Singleapplication can be enabled again, as the main issue in the Qt6 version only occurs when the application is crashing - which should not happen during normal operation!

I removed the option for the deprecated 'Qt5Solutions SingleApplication' and instead made the 'SingleApplication' optional, but enabled by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,11 +119,11 @@ enable_sanitizers(project_options)
 
 # allow for static analysis options include(cmake/StaticAnalyzers.cmake)
 
-set(QAPPLICATION_CLASS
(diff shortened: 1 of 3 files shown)


borgmanJeremy:
FYI if you hit CTRL + C to terminate flameshot after building with single application you are unable to launch another instance without rebooting or building a version of flameshot with no single application enabled. 

I will take a closer look at #4032  soon, thanks for finding that!

Issue #4016, opened 2025-06-18
Confinement issue on Ubuntu Gnome


borgmanJeremy:
I am opening this to track the upstream post I made here: https://forum.snapcraft.io/t/screenshots-on-gnome-in-snap/47452

The issue is the snap build does not capture screenshots on gnome because gnome returns an error that says "Only the focused application is allowed to show a system access dialog"

However this happens even using the launcher _which is_ a focused application window. It also does not happen when building from source.

borgmanJeremy:
Changed title to reflect the post on the snap forums.

This is also an issue with flatpak on ubuntu gnome, but not flatpak on arch

borgmanJeremy:
A workaround for the flatpak is to manually grant permission with:

```
flatpak permission-set screenshot screenshot org.flameshot.Flameshot yes
```

Good info here: https://github.com/NixOS/nixpkgs/issues/351680

Kyuyrii:
I tested the Snap version on Ubuntu 24.04 and Fedora 42, both on Gnome, and no window appears asking if I want to allow Flameshot to take screenshots.

In the terminal, this error appears 
```(flameshot:15206): GLib-GIO-WARNING **: 16:37:21.766: GDBus.Error:org.freedesktop.portal.Error.NotAllowed: This call is not available inside the sandbox```

When I tested it in KDE, it worked without any problems, as well as in Budgie using Labwc.

So I'm really confused about it not working specifically in Gnome.

Edit: I just installed Fedora Cosmic Spin, Flameshot Snap also worked fine on it.

Kyuyrii:
I edited snapcraft.yaml to remove the slot information, installed it and tested it.

It worked fine in KDE with Wayland, as well as in Budgie with Labwc.

So now I'm confused about whether the dbus-flameshot slot is doing anything.

GarThor:
> A workaround for the flatpak is to manually grant permission with:
> 
> ```
> flatpak permission-set screenshot screenshot org.flameshot.Flameshot yes
> ```
> 
> Good info here: [NixOS/nixpkgs#351680](https://github.com/NixOS/nixpkgs/issues/351680)

This solution seems to work for me, however the flatpack version seems to only want to capture my center (main) monitor, and doesn't see the ones on the left/right...

I'm not sure if there's a setting for this somewhere, or if maybe there's a bug, because this wasn't the default behavior in the "apt" version...

[...]

PR #4032, opened 2025-06-22
Kdsingleapplication as SingleApplication alternative

ElTh0r0:
As mentioned in the the [v13 Readiness discussion](https://github.com/flameshot-org/flameshot/discussions/4012#discussioncomment-13539263), I wanted to give [KDSingleApplication](https://github.com/KDAB/KDSingleApplication) ([blog post about it](https://www.kdab.com/kdsingleapplication/)) a try. The integration into Flameshot was quite easy.

From a functional point of view, I'd say that there isn't much difference between SingleApplication and KDSingleApplication in terms of the features we are using for Flameshot. However, KDSingleApplication is currently under active maintenance and the developers are highly experienced with Qt.

@borgmanJeremy and @mmahmoudian please test and let me know your thoughts, if you consider to either keep both modules as an option or to remove one of them completely.

(if you consider this PR useful, please note, that #4030 should be merged first as I used 4030 as basis for this PR)


ElTh0r0:
While resolving the merge conflicts, 'USE_SINGLEAPPLICATION' is enabled as default. Please swap ON/OFF with 'USE_KDSINGLEAPPLICATION' for testing.

borgmanJeremy:
I looked through the code and its pretty straightforward. I'm very much in favor of just replacing the current SingleApplication with this instance. I read through the code and it uses local sockets rather than Shared Memory so it's not subject to the same bug. It worked great on linux. 

I need to test on my Mac before merging but if that goes okay let's just make this the only dependency for SingleApplication.

Good find!

borgmanJeremy:
This also works great on Mac. Assuming you tested with windows I propose removing the old dependency, but keeping a compile time flag to enable / disable SingleApplication. Default should be enable.

ElTh0r0:
> This also works great on Mac. Assuming you tested with windows I propose removing the old dependency, but keeping a compile time flag to enable / disable SingleApplication. Default should be enable.

Yes, I have successfully tested it on both Manjaro and Windows.
With the last commit I removed the "old" SingleApplication and enabled KDSingleApplication by default.

Kyuyrii:
I generated the Snap version with the version from your repository, these changes solve the problem about the Snap version refusing to open.

ElTh0r0:
> I generated the Snap version with the version from your repository, these changes solve the problem about the Snap version refusing to open.

Thanks for sharing @Kyuyrii! I have absolutely no experience with Snaps, but I'm surprised, that this Pull Request affects the Snap build. @borgmanJeremy was working on the Snap build, maybe he can comment on that.

EDIT: Ah ok, now I see you mentioned your findings in the Snapcraft forum as well (referenced in #4016). So it seems there is really a relation with the different SingleApplication modules!

borgmanJeremy:
Yup, I noticed this on the snap but knew we were pivoting. Thanks for the changes!

borgmanJeremy (approved):

PR #4033, opened 2025-06-22
Remove StrftimeChooserWidget button layout creation constraints

ElTh0r0:
Currently there is a constraint, that `m_buttonData` must have even number of items:
https://github.com/flameshot-org/flameshot/blob/172d561fb877c4d5aa7ecdf74230576a94ab1263/src/config/strftimechooserwidget.cpp#L13-L16

This PR makes StrftimeChooserWidget button layout creation independent if number of buttons is even or odd.

This refers e.g. to #3710, which adds a new entry and discussion about the even/odd number of items started.

### DIFF
--- a/src/config/strftimechooserwidget.cpp
+++ b/src/config/strftimechooserwidget.cpp
@@ -9,27 +9,33 @@
 StrftimeChooserWidget::StrftimeChooserWidget(QWidget* parent)
   : QWidget(parent)
 {
+    const quint8 MAXCOLUMNS = 2;
     auto* layout = new QGridLayout(this);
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Did you test a case where there are an odd number of entries? If yes can you post a screenshot of how it looks?

ElTh0r0:
> Did you test a case where there are an odd number of entries? If yes can you post a screenshot of how it looks?

Sure
Odd (was the current status with 21 items - one wasn't shown because of this constraint!!):
grafik

Even (22 items after adding the new format from #3710):
grafik

PR #4034, opened 2025-06-22
Make Imgur conditionally compiled in

borgmanJeremy:
This will conditionally enable or disable imgur uploading. If uploading is disabled it will not be in the toolbar, nor will any config options show in the GUI. However, if a user has settings in their config file they will be properly parsed and not throw errors.


mmahmoudian (approved):
LGTM

mmahmoudian:
@borgmanJeremy I just noticed that the `ENABLE_IMGUR` is `OFF` by default. I think this would cause some backlash from the community as this would disrupt existing workflow. Is it ok if we turn it `ON` by default?

borgmanJeremy:
I intentionally made it OFF by default because I strongly feel it should not be enabled for privacy reasons. My thinking was the official project could leave it off to reflect our views on privacy, but leave the feature there for users and distro's to enable if they disagree. 

For example anyone is welcome to make an AUR package called "flameshot-imgur" and re-enable the feature. 

We even could make a parallel repo that builds with it enabled, and just not put that version out as the official package on flathub. 

If you feel it should be enabled by default lets keep discussing it before the release. But if the only reason to enable it is to avoid backlash, I rather would just explain our stance on why it's not on be default, take the backlash, and have a project we feel better aligns with our values.

mmahmoudian:
My only reason is avoiding breaking workflows. People are lazy and they don't read anything we put out, distro devs are most probably go with the default and not enabling it. Majority of our users are Ubuntu users, and even from our Arch users majority are using the official repo of Arch. This means that upon software update, majority would hit with a disrupted workflow.

Perhaps we can make a compromise: enabling it by default but inactivate it from configurations for the next release, but if anyone activate it via config, we put a clear notice that "from the next version this would get disabled and users can only enable it through the compile flag". This way we have given enough heads-up, and users will be redirected to a blog post of discussion page on Github to read about our stance and the logic of doing this. What do you think?

If we had the plugin system ready, this wouldn't have been an issue and we just needed to educate the users to install the plugin. But now we are deliberately taking a [questionable] feature away.

We can also gauge the users opinion on a poll on Mastodon and our Matrix channel if you think that's useful data.

borgmanJeremy:
What if we keep it off for the beta release, but link to a discussion with a poll? Then whoever helps test the release will get to give their opinions before a final release.

mmahmoudian:
I like the idea👍. But does it make sense if we also add a test to the startup notification that "this is a RFC and please read the release notes" with a link to it? This would encourage AUR and nightly-build users to participate.

That said, something that confuses me is why we haven't heard back anything these past 5 days from these people 🤔 it is very unlikely that no AUR user has a strong opinion about imgur

quiddity-wp:
Hi folks. The release-notes link here, and suggest user-feedback is wanted, so hopefully this is the right place (apologies if not). 

Feedback: I'd suggest the existing option of "Upload image without confirmation" is sufficient (assuming the default is `off`; mine is currently set that way but I tend to change defaults!). I believe that is sufficient to provide the sometimes-helpful 2nd step to prevent accidental uploads. 
My use-case-context: Uploading images _very-rapidly_ is a key feature for me, so that I can quickly respond to discussions in platforms like IRC. However, I'm also technically capable enough to edit a config-file if needed. 

P.S. Thanks for all your efforts on this really well-designed and implemented tool. I use it multiple times almost every day, for many years now.

mmahmoudian:
@quiddity-wp 
Thanks for the feedback. It is indeed valuable to know the use-cases and habits of users. We don't have any form of telemetry and therefore we have no way to get these information other than these types of feedbacks.

Regarding this particular one, it was decided to disable the upload by default for few reasons. From the top of my head:
1. Flameshot is a screenshot and annotation application, and uploading is out of the scope of the software
2. We have taken some steps towards the "plugin system", and we hope that we transfer imgur upload to it's own plugin
3. Imgur related issues are from time to time flooding our issues, and to this day I cannot remember even one case that they were something that we could do anything about it. It would just makes sense to make the imgur upload it's own plugin and put all the issues there.
4. We have got numerous requests to disable the upload due to OpSec. This typically came from companies which use Flameshot as part of their operations software stack and they want to prevent data leakage.

For now imgur is disabled by default, but users can easily enable it during the compile time with a simple flag. We are hoping that those who package Flameshot would carefully choose the flags based on their users demographics.

I'll try to add a section to our website [here](https://flameshot.org/docs/guide/imgur-help/) and [here](https://flameshot.org/docs/installation/source-code/)

Smok3y97:
It's a shame to see that this feature got turned off by default and hidden behind a "build from source" option. Not every user is that tech-savvy or willing to compile something from source every time an update is released. The Imgur upload was a really useful feature, especially for Discord servers that have image uploads disabled.

I genuinely appreciate how lightweight Flameshot is, but this change was a reason for me to downgrade. I tried ShareX, but I really don't like its interface. I hope the Imgur upload feature comes back in the future—even if it's via a plugin, that would be fine with me. Otherwise, I might have to switch back to Greenshot, even though it's not as good as Flameshot.

mmahmoudian:
@Smok3y97 during the testing time of the version 13 release candidate, we asked users through [our social media](https://fosstodon.org/@flameshot/114858430345858922) to give us feedback. We received zero feedback regarding this during the past 20 days. This is not to say imgur upload was not handy for users, but it is to say either it was not critical enough for users to raise their concerns, or people genuinely don't follow the news nor our social media posts. We don't have telemetry and therefore the only way we can gauge users' use-cases is through their feedback. When feedback is non-existent, we can't do much about it 🤷

VicioBR:
If you’re open to feedback, here’s mine.
It’s regrettable that such a useful workflow feature (the automatic uploading of screenshots to the cloud) has been removed.

I understand the privacy concerns, but removing this functionality arbitrarily doesn’t seem like a prudent decision. By doing so, you’re assuming that all users are incapable of deciding for themselves whether they want to use the feature.

My suggestion is simple: keep the feature disabled by default and allow users to enable it manually if they wish. Leave the code available so anyone who wants the function back can compile it themselves—it’s crazy, since most people barely know how to point and click, let alone compile code.

Nowadays, a screenshot app without direct integration to cloud-upload services is like a download manager that can’t download streaming videos—it’s hardly useful! I even reverted to version 12 myself.

I sincerely hope you’ll reconsider this decision. It’s a feature that many users clearly miss in their daily workflows.

nosovk:
actually it would be nice some kind of api, that I can implement for example on my side to upload images to cloudflare r2 by example

ElTh0r0:
> actually it would be nice some kind of api, that I can implement for example on my side to upload images to cloudflare r2 by example

there are already discussions about a plugin system ongoing, see #3953, which could (not only) be used for additional image uploads once it is implemented

PR #4046, opened 2025-07-01
Support for X11 and the standard file picker.

Kyuyrii:
A small change to make Flameshot Snap work on Xorg.

This also allows Flameshot to open the default file picker for the interface you are in.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -27,8 +27,6 @@ apps:
       - kde-neon-6
     environment:
       XDG_DATA_DIRS: $SNAP/share:$XDG_DATA_DIRS
-      QT_QPA_PLATFORM: wayland
-      QT_QPA_PLATFORMTHEME: gtk3
     slots: [dbus-flameshot]
     plugs:
       - home


borgmanJeremy:
This is a regression on wayland desktops though, correct?

Kyuyrii:
What exactly do you mean?

borgmanJeremy:
What I mean is if  QT_QPA_PLATFORM: wayland is not set, the app won't work correctly under wayland sessions (i believe). It's hard to tell for sure since I can't get the snap to work at all on gnome wayland.

Kyuyrii:
I tested it on KDE and Budgie with Labwc, it worked without any problems.

But there's something strange, you know the dbus slot in snapcraft.yaml? I removed it and generated a ```.snap```, installed it and it worked anyway.

borgmanJeremy:
Im confused how it still works on wayland without that :(

I kinda of want to give up on a snap release for this version. @mmahmoudian  I suspect the snap will be the main remaining package issue as I have no line of sight to fix the permission issue on gnome.

Kyuyrii:
I tested Flameshot Snap on Gnome with Xorg, KDE on Xorg and Wayland, Budgie on Xorg and Wayland/Labwc, and it worked fine.

I still don't understand what Gnome does differently that only causes problems on Wayland.

borgmanJeremy:
Ok thanks for all the testing. I'll merge this and put details in the release notes. Does snap have anything like flatseal that would let someone manually grant the permission?

Kyuyrii:
Snap is much more controlled than Flatpak, an example of this is the difference in the use of icon themes, GTK and cursors.

I saw a pull request about Wayland and Gnome, could you take a look at it? If you can fork it and add the changes from the pull request, I can generate a Snap version to test if it changes anything.

https://github.com/flameshot-org/flameshot/pull/3215

borgmanJeremy:
@Kyuyrii see #4055

Issue #4007, opened 2025-06-13
copying to clipboard works only few minutes after pushing it via tray icon


emoxam:
### Flameshot Version

flameshot --version

> Flameshot v12.1.0 (Debian 12.1.0+ds-2)
> Compiled with Qt 5.15.1

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

kubuntu 25.04

### Description

i use shortcut meta+shift+s to make screenshot and enter to copy area to clipboard.
It doesn't work until i click on tray icon and select "take screenshot".
than my shortcut works for some time.. after that it stopped..

here's my config if it's matter

> cat /home/emoxam/.config/flameshot/flameshot.ini
> [General]
> buttons=@Variant(\0\0\0\x7f\0\0\0\vQList<int>\0\0\0\0\x18\0\0\0\0\0\0\0\x1\0\0\0\x2\0\0\0\x3\0\0\0\x4\0\0\0\x5\0\0\0\x6\0\0\0\x12\0\0\0\xf\0\0\0\x16\0\0\0\x13\0\0\0\a\0\0\0\b\0\0\0\t\0\0\0\x10\0\0\0\n\0\0\0\v\0\0\0\r\0\0\0\x17\0\0\0\xe\0\0\0\f\0\0\0\x11\0\0\0\x14\0\0\0\x15)
> contrastOpacity=188
> 
> [Shortcuts]
> TYPE_ACCEPT=Return
> 

### System Information

![Image](https://github.com/user-attachments/assets/41c41e84-679d-4434-946e-afd8ae8bb9f3)

KDE
Wayland

![Image](https://github.com/user-attachments/assets/db2a23ab-4cf9-4314-8e5d-6fc4693dfe77)

borgmanJeremy:
Can you please test using our master branch?

emoxam:
what does it mean?

HarshNarayanJha:
It means to build flameshot from source and test it

emoxam:
I don't know how to do it. Is there any builded release?

HarshNarayanJha:
> I don't know how to do it. Is there any builded release?

you are on kubuntu right? So you must have installed flameshot using `apt install flameshot`. To test the master branch, you must build flameshot from source.

[...]

PR #4051, opened 2025-07-02
make the USE_WAYLAND_CLIPBOARD by default True

mmahmoudian:
This is based on the suggestion by @borgmanJeremy :
https://github.com/flameshot-org/flameshot/issues/4007#issuecomment-3020801944

I'm not sure if what I did is correct, I just followed the logic as far as I understand it, and made sure it build and works. :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,7 +91,7 @@ option(USE_MONOCHROME_ICON "Build using monochrome icon as default" OFF)
 option(GENERATE_TS "Regenerate translation source files" OFF)
 option(USE_KDSINGLEAPPLICATION "Use KDSingleApplication library" ON)
 option(USE_LAUNCHER_ABSOLUTE_PATH "Use absolute path for the desktop launcher" ON)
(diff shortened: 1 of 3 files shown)


mmahmoudian:
Yes, it seems my logic is flawed:

> Could not find a package configuration file provided by "KF6GuiAddons"

https://github.com/flameshot-org/flameshot/actions/runs/16029575932/job/45226346512?pr=4051#step:5:84

ElTh0r0:
> Yes, it seems my logic is flawed:
> 
> > Could not find a package configuration file provided by "KF6GuiAddons"

For e.g. Fedora and Arch this change would be fine, but would need to be disabled for Debian (stable) and Ubuntu <= 24.04.
The required package 'libkf6guiaddons-dev' is currently only available for [Debian testing & unstable](https://packages.debian.org/search?keywords=libkf6guiaddons-dev&searchon=names&suite=all&section=all). For Ubuntu the package is available starting with [24.10](https://packages.ubuntu.com/search?keywords=libkf6guiaddons-dev&searchon=names&suite=all&section=all)

PR #4055, opened 2025-07-03
Add gnome-based wayland universal screenshot adapter

borgmanJeremy:
Rebasing #3215  for @Kyuyrii  to test. Still needs to be reviewed and tested more. 

By default, flameshot uses the program's default dbus protocol interface to communicate with gnome's screenshot protocol, but this is now broken on some gnome installations, so we'll use gnome-screenshot as gnome's wayland screenshot adapter and then communicate directly with the gnome screenshot component!

To enable the GNOME-based generic wayland screenshot adapter, enable it using the following cmake parameter:

```
-DUSE_WAYLAND_GNOME=true
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -229,6 +229,14 @@ if (USE_WAYLAND_CLIPBOARD)
   target_link_libraries(flameshot KF6::GuiAddons)
 endif()
 
+if(USE_WAYLAND_GNOME)
+    target_compile_definitions(flameshot PRIVATE USE_WAYLAND_GNOME=1)
+endif()
+
+if (USE_WAYLAND_GRIM)
+    target_compile_definitions(flameshot PRIVATE USE_WAYLAND_GRIM=1)
+endif()
+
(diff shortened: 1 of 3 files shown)


Kyuyrii:
I assume the modification needs to be made in an updated version, this version of the ```gnome_adapter``` branch does not yet use Kdsingleapplication.

borgmanJeremy:
I'm sorry, I was in a rush and forgot to pull my local master before rebasing. I fixed it and force pushed to this branch.

Kyuyrii:
Unfortunately, nothing has changed here.

One question, does Flameshot access the screen differently than OBS-Studio?

I tested the Snap version of OBS-Studio and it works normally in Gnome.

borgmanJeremy:
I believe it is very similar: https://github.com/obsproject/obs-studio/blob/d91ccdefd250ab2df866ee0545a8ddf4828c3133/plugins/linux-pipewire/screencast-portal.c#L70

I didnt look through their snapcraft file yet.

PR #4064, opened 2025-07-09
Link to translations folder

Kyuyrii:
This modification allows Flameshot Snap to find the translation files.


borgmanJeremy (approved):

Samueru-sama:
This bug is present in the AppImage as well. 

A better fix would be moving the translations dir next to the binary, as it checks that location: 

https://github.com/flameshot-org/flameshot/blob/980cf5352a533463b84ca783eb238c152eb707e5/src/utils/pathinfo.cpp#L21-L23

PR #4076, opened 2025-07-13
revert accidental removal of window title introduced by #3766

mmahmoudian:
This feature initially introduced by #3766 but was removed in 58afdce7414bbe3881610042e4da3772d2d0365f. This can help configuring Flameshot in window managers in a more granular level.

@borgmanJeremy if this removal was intentional, please close this PR, but considering that it plays nicely with Qt6 as far as my testing goes, I'm proposing it back.


borgmanJeremy:
Good catch, thanks. This didn't break anything after re-adding in.

borgmanJeremy (approved):

PR #4078, opened 2025-07-15
Qt 6 - Windows issues loading dis-/enabled tool buttons (fix for #4077)

### LINKED ISSUE #4077: Button Selection not being remembered after restart
### Flameshot Version

v13 beta

### Installation Type

Windows 11

### Description

Pretty straighforward. Change button selection, close Flameshot, open Flameshot, buttons are in default positions again.

### Steps to reproduce

_No response_

_No response_

### System Information

Windows 11, dual screen. Most likely irrelevant.

ElTh0r0:
This PR fixes #4077 

I doubled checked it with the old Qt 5 build and dis-/enabling tool buttons worked on Windows. With Qt 6 loading the saved configuration didn't work anymore and debug console output was showing: "QVariant::load: unknown user type with name QList<int>."

There must have been changes in the Qt meta-type system and platform-specific differences. QList<int> is not a built-in QVariant type. To use custom types we need to register them using Q_DECLARE_METATYPE(QList<int>) and qRegisterMetaType<QList<int>>().

In Qt 5, handling of QList<T> was more permissive, so QList<int> sometimes worked without explicit registration. Starting with Qt 6, the meta-type system has become stricter and does not recognize QList<int> natively anymore. This causes the type to be unknown when deserializing (e.g. from QSettings), unless explicitly registered.

I couldn't find clear explanation for the differences on Windows and Linux, but AI mentioned: "The Qt build configuration or available precompiled meta-types can differ between Windows and Linux, which is why the problem only appears on one platform."

I tested this PR on Windows and on Manjaro and it worked on both systems, so I'd say we don't need explicit `#if defined(Q_OS_WIN)`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -31,6 +31,9 @@
 #include <desktopinfo.h>
 #endif
 
+// Required for saving button list QList<CaptureTool::Type>
+Q_DECLARE_METATYPE(QList<int>)
+
 int requestCaptureAndWait(const CaptureRequest& req)
 {
     Flameshot* flameshot = Flameshot::instance();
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Thanks for the quick fix! I also tested and it didn't break anything.

borgmanJeremy (approved):

PR #4100, opened 2025-07-31
Update shell.nix for QT6

TheToddLuci0:
The namespace for qt6 is `kdePackages` instead of `libsForQt5`


borgmanJeremy:
Thank you!

PR #4121, opened 2025-08-06
Introduce some exit codes (#3321)

guyueshui:
Recall what @mmahmoudian [mentioned](https://github.com/flameshot-org/flameshot/issues/3321#issuecomment-2626767911), this PR introduce these error codes,

1. `E_OK`: exit normally
2. `E_GENERAL`: general error
3. `E_ABORTED`: screenshort aborted by user
4. `E_DBUSCONN`: quit by dbus connection error

Currently, only 1, 3, 4 are used, cause I found only 2 cases of abnormally quits (by call `qApp->exit()`) . And one can easily add more error codes as they need.

Note that I didn't completely follow mmahmoudian's advice. Because at present, there isn't so many abnormal quits. I want to keep things simple and fit the design. We can add a new error code when we encounter a new error case. I must admit these things may be extended and optimized in future. But now, I think it's enough.

Hope it helps.


mmahmoudian:
@guyueshui thanks for the PR. Looks good to me, but let's hear the opinion of @borgmanJeremy and other devs about this.

If you are willing to improve the PR, the manpage can also be updated to include the exit codes. If you don't feel like it, I can later add those in a separate PR when the coding system and codes are finalized.

guyueshui:
@mmahmoudian Of course I am willing to. Do you mean this file `/data/man/man1/flameshot.1`? I took a look at it, seems like it's generated by some program? Or written manually? I haven't wrote any manpage before, could you please give me some instructions or general rules?

src/core/flameshotdaemon.cpp
```diff
@@ -151,10 +151,7 @@ void FlameshotDaemon::copyToClipboard(const QString& text,
     auto m = createMethodCall(QStringLiteral("attachTextToClipboard"));
 
     m << text << notification;
-
-    QDBusConnection sessionBus = QDBusConnection::sessionBus();
-    checkDBusConnection(sessionBus);
-    sessionBus.call(m);
+    call(m);
```
borgmanJeremy, line 154: Why was this change made?

borgmanJeremy (changes requested):
I think this should also be added here:
https://github.com/flameshot-org/flameshot/blob/48f8d1fdd8baccfd8024e6c8a53db2167a9e900d/src/core/signaldaemon.cpp#L48

and the other signal hander in that file.

src/core/flameshotdaemon.cpp
guyueshui: It has nothing to do with exit code, but I find these 3 lines are equivalent to `FlameshotDaemon::call`, just change it by the way. Did I make something wrong?

guyueshui:
> I think this should also be added here:
> 
> https://github.com/flameshot-org/flameshot/blob/48f8d1fdd8baccfd8024e6c8a53db2167a9e900d/src/core/signaldaemon.cpp#L48
> 
> and the other signal hander in that file.

Updated.

src/core/flameshotdaemon.cpp
borgmanJeremy: It's probably fine, but it would be better to split it into 2 PR's or commits where 1 commit is the new feature, and 1 is the code cleanup.

I have had to bisect the code quite a bit lately, and having separate thoughts broken out to separate commits helps with that.

mmahmoudian:
Regarding the manpage, it has been written and updated manually.

Here is a very quick guide:

https://www.cyberciti.biz/faq/linux-unix-creating-a-manpage/

I know the format is a bit unintuitive. If you had any questions, feel free to ask and I'll fo my best to help :)

src/core/flameshotdaemon.cpp
guyueshui: I have rewrote the commit history to separate code clean against with the new feature. Hope it helps later. Feel free to post any further issues if you have ;)

guyueshui:
@mmahmoudian Thanks for your link. I have made some changes in the manpage, but it's really very simple and basic, it doesn't have any style. Please check if it is ok when you have some time. And of course, further advice is always welcome.

guyueshui:
@borgmanJeremy May I ask when would you like to review these changes (just a reminder)? A feedback helps me make further improvements.

borgmanJeremy:
@guyueshui  sorry for the delay. I'll review soon. I've been trying to wait to merge new features until the v13 bug fixes have slowed down.

guyueshui:
Not sure why this [CI](https://ci.appveyor.com/project/borgmanJeremy/flameshot/builds/52749637) failed. I update it to resolve the conflicts with master. Please give it a review when you have some time. @borgmanJeremy

borgmanJeremy:
The appveyor failure is fixed if you rebase again. Sorry I will try to merge this over the weekend.

I really appreciate the contribution and sorry its been slow on my end.

borgmanJeremy (approved):

borgmanJeremy:
Apologies for the delay on this. Looks good, thanks for the contribution.

guyueshui:
You are welcome, I am glad to see flameshot becomes better and better, thanks for your efforts.

PR #4130, opened 2025-08-07
build: Add an option to use a bundled KDSingleApplication

EbonJaeger:
When packaging, distros like to not use bundled dependencies when possible. This adds an option to use a bundled version of KDSingleApplication, or a system version of the library.

I briefly explored doing the same for Qt-Color-Widgets, but it seems to expect to be bundled in a project.


borgmanJeremy:
Thank you!

PR #3387, opened 2023-10-22
Add shortcut for Grab Color

AlexP11223:
Resolves #3384, adds a shortcut (default G) for starting color grabbing.

After finishing it opens the side panel even if it was hidden before. I am not sure if there is a simple way to fix that, but it may be the right way to do it anyway, so that you can see what color was actually grabbed.

PR #4153, opened 2025-08-13
Add shortcut 'G' to set cursor color to position's color

Gai-shi:
This PR introduces a new shortcut functionality: pressing the 'g' key will set the brush color to match the pixel color at the cursor's current position.
﻿
Key implementation details:
- Added 'g' key listener in the input handling module to trigger color sampling
- Implemented logic to extract the exact color values from the pixel under the cursor's coordinates
- Integrated the sampled color into the brush's active color configuration, ensuring immediate application to subsequent drawing actions
- Verified no conflicts with existing shortcut keys by cross-checking against the project's key binding registry
- Tested in multiple scenarios: different image resolutions, transparent backgrounds, and with various brush sizes (all function as expected)
﻿
This feature streamlines the workflow for users who need to quickly match colors from the canvas, reducing the need to manually select colors through the palette, thus improving editing efficiency.

### DIFF
--- a/src/tools/capturetool.h
+++ b/src/tools/capturetool.h
@@ -87,7 +87,7 @@ public:
     {}
 
     // TODO unused
-    virtual void setCapture(const QPixmap& pixmap){};
+    virtual void setCapture(const QPixmap& pixmap) {};
 
     // Returns false when the tool is in an inconsistent state and shouldn't
     // be included in the tool undo/redo stack.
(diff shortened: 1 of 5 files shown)


ElTh0r0:
Thanks for your effort! But there is a similar PR already waiting for merging: #3387
The approach of 3387 is slightly different: After pressing 'G' it is opening the side panel and starting the color grab from there.

Just my two cents: I'd personally prefer 3387, because it is using the build-in color grab. Your PR is reinventing the existing color grab functionality. (but it is not on me to finally decide on this).

borgmanJeremy:
@ElTh0r0  Good call, lets merge #3387

PR #4162, opened 2025-08-15
bump for v13.1.0

borgmanJeremy:


orhanbsnkimya:
Hello. I'm using a dual-screen laptop and monitor. If I take a screenshot while in the monitor mode, the screen resolution changes and the image gets bigger. That's why I went back to 12.1.0

PR #4127, opened 2025-08-07
Fix multi monitor offset bug in Qt6 multi-monitor setups

ionsquare:
The issue was caused by using QApplication::primaryScreen() to capture the entire desktop, but in Qt6 the primary screen is not necessarily the leftmost screen. This caused coordinate system misalignment in multi-monitor setups where the primary screen is positioned to the right.

Additionally, the previous approach failed in edge cases where screens have different heights and the desktop bounding box includes virtual space that doesn't correspond to any physical screen.

Changes:
- Replace single-screen desktop capture with composite approach
- Capture each screen individually and composite them together
- Position capture widget to cover entire desktop geometry
- Remove devicePixelRatio division that caused coordinate issues
- Handle virtual desktop space by filling with black background

This solution works with any multi-monitor configuration including:
- Primary screen not being leftmost
- Screens with different resolutions and alignments
- Bottom-aligned or top-aligned screen arrangements
- Complex layouts with virtual coordinate space

Fixes #4111

PR #4166, opened 2025-08-16
Minor RPM spec file tweaks

pointydev:
Was just looking at the RPM spec file now that I can actually use the newer builds thanks to #4127, but just noticed some minor issues that I thought I'd correct:

- Corrected version to match actual release
- Removed mention of Imgur from description since its no longer included in the build
- Corrected RHEL typo in the comment at the top for the Fedora spec
- Add RPM vendor for use with [`allow_vendor_change=False`](https://dnf.readthedocs.io/en/latest/conf_ref.html#main-options) (in Fedora, this is set true by default)
- Bumped release for previous issues

Thanks,
Elliott

### DIFF
--- a/packaging/rpm/fedora/flameshot.spec
+++ b/packaging/rpm/fedora/flameshot.spec
@@ -1,9 +1,9 @@
 #
-# spec file for package flameshot on fedora, rehl
+# spec file for package flameshot on fedora, rhel
 #
 Name: flameshot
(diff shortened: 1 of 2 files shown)


mmahmoudian (approved):

mmahmoudian:
@pointydev thanks for the PR.

borgmanJeremy:
thanks!

PR #4179, opened 2025-08-19
Fix translator memory leak

borgmanJeremy:
This fixes #4168

PR #4190, opened 2025-08-21
adding sha to all builds

borgmanJeremy:


pointydev:
Works for me, tested with AppImage to avoid having to replace my install from #4179:

image

```
$ ./Flameshot-13.1.0.x86_64.AppImage --version
"No Qt translation found for English"
Flameshot v13.1.0 (c8f6a441)
Compiled with Qt 6.2.4
```

borgmanJeremy:
Thanks for confirming, I added the same pattern to all the other piplines now.

PR #4202, opened 2025-08-23
Added hash to CI builds

borgmanJeremy:
This is a temporarly fatal error to more easily see which CI pipelines still are not pulling in the commit hash.


borgmanJeremy:
The issue was packpack is a container and the env variables were not properly passed through

Issue #4118, opened 2025-08-06
Copy selection to clipboard don't work on cli windows


brunolima545:
### Flameshot Version

Flameshot v13.0.0 (a29fada2)

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Windows 11

### Description

I can't copy a selection to the clipboard when using the Windows CLI with the "flameshot.exe gui" command, using the application in the system tray works normally

### Steps to reproduce

1. run 'flameshot gui'
2. select a region on any monitor
3. press CTRL+C or click on "copy to clipboard button"

When trying to paste somewhere the image was not loaded, the clipboard continues with what was there before

### System Information

Windows 11 24H2 26100.4770

borgmanJeremy:
I suspect that the daemon does not stay active and thus the clipboard loses its content. Maybe @ElTh0r0 can help debug this

shouya:
I have the same problem and it's solved by enabling the "Use JPG format for clipboard" option.

~~I suspect that the [`QApplication::clipboard()->setPixmap(capture);`](https://github.com/flameshot-org/flameshot/blob/23fade7bb943882d0a84cccc3d25a052ec25edd6/src/utils/screenshotsaver.cpp#L208-L216) method may not work as expected.~~

Update: problem resolved by installing xdg-desktop-portal-xapp. So the problem should be unrelated to Windows.

felixoakz:
Same issue here after last update on Arch Linux with GNOME (Wayland). Copy to clipboard stopped working.

Tried:
- Enabling "Use JPG format for clipboard" in Flameshot settings
- Installing `xdg-desktop-portal-xapp`
- Restarting `xdg-desktop-portal` and Flameshot
- Confirmed `DBUS_SESSION_BUS_ADDRESS` is set correctly
- Clipboard utilities (`wl-copy`/`wl-paste`) work fine independently
- Tested `flameshot gui -c` — no errors but clipboard remains empty
- Running `flameshot` without tray option (default Wayland behavior)

I'm commenting because I don't know what else to test for now, and want to keep track on this issue.
I have downgraded to version 12.1.0 and it just working back ok.

ndzuki:
After update arch linux (Wayland) . flameshot copy to clipboard not work.

ElTh0r0:
> I suspect that the daemon does not stay active and thus the clipboard loses its content. Maybe [@ElTh0r0](https://github.com/ElTh0r0) can help debug this

[...]

PR #4211, opened 2025-08-27
Windows cli copy (fix for #4118)

ElTh0r0:
As mentioned in #4118, copying a screenshot is not working when the flameshot CLI is used on Windows. Currently Flameshot daemon is using D-Bus to forward the CLI request to the main Flameshot instace and D-Bus is not available on Windows (and MacOS as far as I know).

This PR is using a feature from the recently added KDSingleApplication, with which we can send a message from the second (CLI) Flameshot instance to the first instance without D-Bus.
I successfully tested it on Windows 10; I don't have MacOS, but I'd assume it should work there as well - if somebody could test it on mac it would be great! Additionally I added further ifdefs, so that the Windows and MacOS build has no dependency to Qt D-Bus module anymore.

_Side note:_ I tested this functionality on Manjaro linux as well (after adjusting the code to use the KDSingleApplication message feature instead of the usual D-Bus code for linux) and it worked as well. So I am raising the question if we should consider to remove `flameshotdbusadapter` completely and use KDSingleApplication messages on all platforms?! (for sure D-Bus will remain in general for the Linux builds, as D-Bus is used e.g. by `ScreenGrabber`). Discussion #1999 might be a better place to discuss this in detail.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,7 +10,7 @@ find_package(
         LinguistTools
 )
 
-if (UNIX)
+if (UNIX AND NOT APPLE)
 find_package(
         Qt${QT_VERSION_MAJOR}
         CONFIG
@@ -220,7 +220,7 @@ target_link_libraries(
         Qt${QT_VERSION_MAJOR}::Widgets
         QtColorWidgets
 )
-if (UNIX)
+if (UNIX AND NOT APPLE)
 target_link_libraries(
         flameshot
         Qt${QT_VERSION_MAJOR}::DBus
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -7,9 +7,17 @@ find_package(
         Widgets
         Network
         Svg
-        DBus
         LinguistTools
 )
+
+if (UNIX)
+find_package(
+        Qt${QT_VERSION_MAJOR}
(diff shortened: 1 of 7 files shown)


xefiry:
Hi,
I just tested the portable version on Windows 11 by running `flameshot-cli.exe gui`.
The copy (Ctrl + C) works, but **only** if flameshot.exe is running in the background.
If flameshot.exe is not running, Ctrl+C does not copy and the command line is still running a few seconds after pressing Ctrl + C.

ElTh0r0:
> The copy (Ctrl + C) works, but **only** if flameshot.exe is running in the background.

Thank you for testing!
According to my understanding this is the intended/default behaviour (independent of the operating system) - please anybody correct me if I am wrong.

borgmanJeremy:
@ElTh0r0  you are correct, that is how it works on linux too because a running application needs to host the clipboard data.

borgmanJeremy:
I tested this on KDE but not on Mac yet. Once I am able to do that I'll go ahead and merge this.

borgmanJeremy:
This PR has no impact on the macos behavior. But there is a seperate macos bug where the clipboard doesnt work when invoked with flameshot gui. So I will merge this.

Thanks!

PR #4228, opened 2025-09-07
OCR Tesseract integration 

rokokol:
**Title:** Add Tesseract OCR integration

**Description:**
This PR introduces integration with Tesseract OCR. Users can now copy text from a selected area to the clipboard using `SelectionTextTool`. The tool automatically detects and uses all available language packs. Just use the button with tesseract icon

**Changes include:**

* Implementation of OCR logic in `selectiontexttool.cpp`.
* Automatic tessdata path detection across Windows, macOS, and Linux.
* Automatic gathering of all installed languages.
* Clipboard support for both standard and selection buffers, including Wayland compatibility.

This feature enables seamless text recognition from screenshots or selected regions.

**Deployment**
I use Linux Mint Cinnamon 22.2. You should to install the tesseract and, for example, the English package:
```sh
sudo apt install tesseract-ocr tesseract-ocr-eng
```


borgmanJeremy:
Thank! I will review more closely at my PC. But at a high level I think we should make this an optional integration so tesseract is not a hard dependency. @mmahmoudian what are your thoughts?

mmahmoudian:
@borgmanJeremy I remember this being discussed before. OCR is out of the scope of a screenshot tool and therefore it is better if we have this as a plugin. I think it worth the effort to push the plugin system forward.

**Update:** I collected these:
The #702 is the feature request.
This is already implemented 3 times before this PR:
- #3074
- #2843
- #1239

PR #4248, opened 2025-09-19
refreshed translation files

borgmanJeremy:


borgmanJeremy:
Note: weblate locked until this is merged

PR #4249, opened 2025-09-20
Add UI language selection

ElTh0r0:
Currently Flameshot is using the detected system language for loading the UI language. Some users raised the question, if it would be possible to select a different language than the system language (see [translation-instruction/issues/25](https://github.com/flameshot-org/translation-instruction/issues/25) and discussion: #3513).

I used the word 'auto' to refer to the 'automatically' detected language as the default option in the language selection combo box. I'm not a native English speaker, so I'm not 100% sure if this is well understood when used in this context. In addition, this word must remain untranslated; otherwise, this configuration option will not work. It should therefore be understandable all over the world. Or do you have any other ideas on how this could be solved?

My current implementation requires a restart to load the new selected translation. I added a "TODO" in the source, so maybe somebody adds a direct re-translation of all UI elements without restart of the application in future.

### DIFF
--- a/flameshot.example.ini
+++ b/flameshot.example.ini
@@ -18,6 +18,9 @@
 ;; Default file extension for screenshots
 ;saveAsFileExtension=.png
 ;
+;; UI language (auto = detected system language)
+; uiLanguage=auto
+;
 ;; Main UI color
 ;; Color is any valid hex code or W3C color name
 ;uiColor=#740096
--- a/src/config/visualseditor.cpp
+++ b/src/config/visualseditor.cpp
@@ -7,8 +7,10 @@
 #include "src/config/extendedslider.h"
 #include "src/config/uicoloreditor.h"
 #include "src/utils/confighandler.h"
+#include <QDirIterator>
 #include <QHBoxLayout>
 #include <QLabel>
+#include <QMessageBox>
 
 VisualsEditor::VisualsEditor(QWidget* parent)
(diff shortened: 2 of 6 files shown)


borgmanJeremy:
This works great in my testing and I think it solves many corner cases. "Auto" makes sense to me, and has an advantage of always being on the top of the list. 

Also did some testing where I set the config file to an invalid language and it was gracefully handled.

borgmanJeremy (approved):

mmahmoudian:
@borgmanJeremy to facilitate the next release note, I suggest creating a new milestone and add PRs to it. I just need to know if you are aiming at "13.2.0" or "14.0.0". When I create the milestone I'll go back and start adding PRs. Just give me the version number and the rest will be handled 😊

devZu9:
I still don't understand how to set the language I want. There's no such problem on Windows, but on my MacBook running MacOs Tahoe 26, the flameshot program only has one option, auto. However, the entire MacOs interface is in Russian. Everything is in Russian except for flameshot. It's in English, and the language is set to auto. There are no other languages available. I want to set my own language, but it seems like the localization files are not being added. How can I add them? I'm ready to do it manually, but please tell me where to get them and where to put them. I want the familiar Russian interface. What should I do to set the Russian language instead of auto?

borgmanJeremy:
@devZu9 can you please open a bug report and be sure to include your version and how you installed?

PR #4253, opened 2025-09-23
Fix: multi-screen DPI screenshot black edges by adjusting logic pixel

Gai-shi:
Adjusted the logical pixel calculation logic in src/utils/screengrabber.cpp and src/widgets/capture/capturewidget.cpp to adapt to multi-screen DPI differences, ensuring the screenshot area fits the screen accurately.
Tested on Windows system with 2 screens of different DPIs (2K-150% and 2K-125%). The screenshots have no black edges and fit the screens perfectly.
Verified that the single-screen scenario works normally with no regression issues.
I would also like to attach two images: one showing the issue before the fix, and one showing the result after the fix.
before the fix:
m_context__screenshot_diff

after the fix:
fixed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ data/flatpak/.flatpak-builder
 # Miscellaneous
 !docs/dev/Makefile
 .direnv
+"KDSingleApplication/" 
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -268,15 +268,17 @@ QPixmap ScreenGrabber::grabScreen(QScreen* screen, bool& ok)
 QRect ScreenGrabber::desktopGeometry()
 {
     QRect geometry;
-
+    qreal dpr = QGuiApplication::screens()[0]->devicePixelRatio();
     for (QScreen* const screen : QGuiApplication::screens()) {
         QRect scrRect = screen->geometry();
         // Qt6 fix: Don't divide by devicePixelRatio for multi-monitor setups
         // This was causing coordinate offset issues in dual monitor
         // configurations
         // But it still has a screen position in real pixels, not logical ones
-        qreal dpr = screen->devicePixelRatio();
         scrRect.moveTo(QPointF(scrRect.x() / dpr, scrRect.y() / dpr).toPoint());
+        scrRect.setWidth(scrRect.width() * screen->devicePixelRatio() / dpr);
+        scrRect.setHeight(scrRect.height() * screen->devicePixelRatio() / dpr);
+        // dpr = screen->devicePixelRatio();//这里应该是除以第一个屏幕的dpr
         geometry = geometry.united(scrRect);
     }
     return geometry;
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -132,8 +132,20 @@ CaptureWidget::CaptureWidget(const CaptureRequest& req,
                 topLeft.setY(topLeftScreen.y());
             }
         }
+// qDebug()<<topLeft;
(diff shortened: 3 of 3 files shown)


mmahmoudian:
@Gai-shi 
Thanks for the PR. I wonder how portable the code is on Linux and Mac. I can try to test it on Linux.

As a side note, I would be thankful if you can change the code comments to English.

Gai-shi:
> @Gai-shi感谢您的公关。我想知道代码在 Linux 和 Mac 上的可移植性如何。我可以尝试在 Linux 上测试它。
> 
> 顺便说一句，如果您能将代码注释更改为英文，我将不胜感激。

Thank you for your comment and suggestion to test on Linux. I look forward to your Linux testing results!
You may need to modify the code in the following section of src \ widgets \ capture \ capturewidget.cpp:
image
I will update all code comments to English next time. Thank you for your reminder

borgmanJeremy:
I'm not understanding why this fixes an issue. With the new code, the variable "dpr" is set just using screen 0. Then screen 1 is multiplied by the pixel ratio of screen1, and divided by screen 0? 

It's not making sense to me _why_ its working, and im not convinced it would work for 3 screens or in any general sense. Can you help walk me through why you think this fixes the issue?

Gai-shi:
> I'm not understanding why this fixes an issue. With the new code, the variable "dpr" is set just using screen 0. Then screen 1 is multiplied by the pixel ratio of screen1, and divided by screen 0?
> 
> It's not making sense to me _why_ its working, and im not convinced it would work for 3 screens or in any general sense. Can you help walk me through why you think this fixes the issue?

Okay, let me explain:
Assuming the dpr of the main screen is MDPR. For any screen s, we assume that its physical pixel width are PW, and the corresponding dpr is DPR. So the width x of its logical pixels is PW/DPR (which is the logical pixel of the screen itself), assuming that the width of the logical pixels it requires is c, when there is only the main screen, `c * DPR=x * DPR`, Obviously, c and x are equal. But in a multi screen environment, the equation we need to satisfy becomes:
`c * MDPR=x * DPR`
This is because QT is always displayed using the MDPR of the main screen, and all other screens are also forced to display using MDPR. According to this equation, `c=x * DPR/MDPR` can be derived, so what you call "the variable'dpr 'is set just using screen 0". The dpr in the code is actually what I call MDPR here, so the screen s is multiplied by the pixel ratio of screen1, and divided by the MDPR of the main screen.

Gai-shi:
> @Gai-shi Thanks for the PR. I wonder how portable the code is on Linux and Mac. I can try to test it on Linux.
> 
> As a side note, I would be thankful if you can change the code comments to English.

Hello, do you have the test results so far? If you need me to make further code modifications during the process, feel free to let me know anytime!

mmahmoudian:
I got the math, although I didn't understand how Qt calculates the other screens based on screen 0. Regardless, I can test this on a 3 monitor setup and report back.

Gai-shi:
> I got the math, although I didn't understand how Qt calculates the other screens based on screen 0. Regardless, I can test this on a 3 monitor setup and report back.

Okay, thank you for the test. I will reconsider whether this math is completely correct.

panpuchkov:
There is no issue with calculating and generating a proper screenshot. I resolved it a long time ago, and it is in another MR. 
The problem lies in determining how to get the offset for painting it. The primary screen changes depending on the screen location. Try moving your secondary display around the primary one to see the issue.

panpuchkov:
Here is a PR: https://github.com/flameshot-org/flameshot/pull/4230/
I created a new DesktopCapturer class, which can capture a proper screenshot with proper scaling for one or all screens.

borgmanJeremy:
Will merge #4230 instead, but need to fix it for wayland. @Gai-shi  if you want to take a stab at that it would be great

Issue #3637, opened 2024-06-13
Mac OS: Show Cmd+Shift+X shortcut in the menu


ilyagr:
### Feature Description

This is not a major thing, but I would find it very helpful if the shortcut for taking a screenshot (and possibly the other operations) was shown in the application menu pictured below:

<img width="156" alt="image" src="https://github.com/flameshot-org/flameshot/assets/4123047/219d7c31-9836-45eb-94b4-a2da2ce3fdfd">

mmahmoudian:
Do you have examples of other software that put their keyboard shortcuts in their tray icon menu?

ilyagr:
Sure, here's [Maccy](https://github.com/p0deje/Maccy):

![image](https://github.com/flameshot-org/flameshot/assets/4123047/8a39304a-ddf6-4f13-aaa8-95a94dc3d749)


Here's [Rectangle](https://rectangleapp.com/):

![image](https://github.com/flameshot-org/flameshot/assets/4123047/536f08ed-eaa0-4b59-9218-08cc0c04eebc)

mmahmoudian:
Thanks for the examples, but these two doesn't count. They both are always running in the background and providing their services (tiling management and clipboard management), and therefore both should have "global" key bindings. Flameshot on the other hand doesn't listen to anything unless it is explicitly called by the user. This means the only "global" binding is the one to trigger `flameshot gui` (and all the other keybindings you yourself define. [More info about CLI here](https://flameshot.org/docs/advanced/commandline-options/)).

Maybe, **maybe** the screenshot one can be added, but then, that is the only thing that makes sense to me.

ilyagr:
> Thanks for the examples, but these two doesn't count. They both are always running in the background and providing their services (tiling management and clipboard management), and therefore both should have "global" key bindings. Flameshot on the other hand doesn't listen to anything unless it is explicitly called by the user. 

On second thought, if viewed through the lens of global shortcuts, I agree that Maccy wasn't exactly what you asked for. The shortcuts it lists are only valid if Maccy is open. It does have a global shortcut to open that window, but it's not listed inside the window.

The Rectangle example, OTOH, is completely analogous to Flameshot's "Take Screenshot" and "Open GUI" commands.  It sits in the background doing nothing and then, if you press one of its shortcuts, it activates and does its thing (e.g. tiles the current window and puts it in some screen quadrant).

[...]

PR #4258, opened 2025-09-25
Show keyboard shortcut in the tray icon menu

artptr:
Inspired by: #3637

Unfortunately I don't have a Linux or Windows machine to build and check how it would work on these platforms. Anyway, it doesn't seem too difficult to add "PrtSc" to the menu, even though there is no explicitly set shortcut for taking a screenshot in the `ConfigHandler`. I would appreciate it if anybody could build and run it, and provide feedback.

image

### DIFF
--- a/src/widgets/trayicon.cpp
+++ b/src/widgets/trayicon.cpp
@@ -186,12 +186,14 @@ void TrayIcon::initMenu()
 
 void TrayIcon::updateCaptureActionShortcut()
 {
+#if defined(Q_OS_MACOS)
     if (!m_captureAction) {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


mmahmoudian:
Looks good to me

borgmanJeremy:
FYI this causes a config parsing error on Linux (probably windows too). I haven't dug into it much, but i suspect its from this:

https://github.com/flameshot-org/flameshot/pull/4258/files#diff-aa00e694dbae9f7bed89af16ac60043b4bef487a64f903c3b6cbe29e9be3eec9R193

artptr:
@borgmanJeremy thank you, I forgot a guard directive. This shortcut is available only on macOS.

borgmanJeremy:
Looks good now. Thanks!

PR #4268, opened 2025-09-30
CI: Add Windows ARM64 Build support

namoen0301:
Qt >=6.8 can native compile on Windows ARM64 machine, and GitHub also provides windows-11-arm runner. so I think It's time to add support.

I've tested from my branch and Snapdragon 8cx gen3 computer. all ci build passed.


borgmanJeremy:
We need windows binaries to be submitted via appveyor because thats how SignPath ties in to code sign the binary.

namoen0301:
CI built artifact file cannot use to appveyor? or can we temporarily skip signing arm64 files?

borgmanJeremy:
The CI build must use appveyor for windows. I'm not going to release arm64 builds that are unsigned because of the massive number of bug reports it will spawn for being unsigned. 

Edit: But thank you for the contribution. If you can switch it to appveyor I can set up signpath to sign Arm64 windows binaries.

namoen0301:
I'll open a new pr and switch to appveyor. @borgmanJeremy

Issue #4353, opened 2025-10-31
When no notification manager is running Flameshot hangs and causes ubuntu to hang after taking screenshot


samwcoulter:
### Flameshot Version

Flameshot v13.3.0 (88c951e7)
Compiled with Qt 6.2.4

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Ubuntu 25.04

### Description

When no notification manager (ie. dunst or the one provided by gnome) is running: Flameshot hangs after attempting to take a second screenshot. The "command completion" time for the first screenshot also seems excessively long.

### Steps to reproduce

Reproduction steps:
1. Start flameshot (via terminal)
2. run `flameshot gui` command in terminal, select a portion of the screen and click "save" button
3. Right click flameshot taskbar icon, and take another screenshot, save it.

After Step 2, the saved file appears instantly, but the log messages in the flameshot window take about ~45 seconds to indicate the file has been saved, and another minute or so before the command terminates completely:

```
sam@stash-pc:~[main]$ flameshot gui
QLayout: Attempting to add QLayout "" to SidePanelWidget "", which already has a layout
QPainter::begin: Paint device returned engine == 0, type: 2
QPainter::setRenderHint: Painter must be active to set rendering hints
QPainter::setCompositionMode: Painter not active
QPainter::translate: Painter not active
QPainter::setPen: Painter not active
QPainter::setBrush: Painter not active
QPainter::drawEllipse: Painter not active
QPainter::setBrush: Painter not active
QPainter::drawEllipse: Painter not active
flameshot: info: Capture saved as /home/sam/snips/2025-10-31_09-46.png
QThreadStorage: entry 2 destroyed before end of thread 0x5b7d394996e0
QThreadStorage: entry 1 destroyed before end of thread 0x5b7d394996e0
sam@stash-pc:~[main]$ 
```

There is about a 45 second gap between these two messages:
```
QPainter::drawEllipse: Painter not active
flameshot: info: Capture saved as /home/sam/snips/2025-10-31_09-46.png
```
And then another minute or so before the command returns and I get the bash prompt again.

After Step 3, again the saved file appears instantly, but the entire flameshot UI hangs, and if I try to interact with it at all, my entire desktop hangs for 1 to 2 minutes

### System Information

dbus:
```
sam@stash-pc:~/Downloads[main]$ dbus-monitor --session sender=org.flameshot.Flameshot
[...]

PR #4354, opened 2025-11-01
Fix: Handle case where there is no system notification service registered in DBus

samwcoulter:
Issue #4353

Adding a check when looking up the `org.freedesktop.Notifications` service to see if the service actually exists before trying to fetch it. 

The 5 second default timeout in the `sendMessage` function won't cover this case, as it's actually the notification service lookup in the constructor that times out (default 25 seconds). When no notifications service exists, you still get a `QDBusInterface` handle (after the timeout period), but it is not "valid".

I also added a similar check in the `screenGrabber::freeDesktopPortal` function, with an explicit error message which should be easier for users to debug than an unexplained hang/crash if the portal.Desktop service is ever unavailable.

I'm not sure I updated the translation files properly. I'm not familiar with how those are generated/used.

### DIFF
--- a/data/translations/Internationalization_en.ts
+++ b/data/translations/Internationalization_en.ts
@@ -1787,27 +1787,32 @@ You may need to escape the &apos;#&apos; sign as in &apos;\#FFF&apos;</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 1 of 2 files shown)


samwcoulter:
Closing this PR in favor of #4357 which better follows contribution guidelines.

PR #4357, opened 2025-11-02
Fix hang caused by no Notifications service registered in DBus

samwcoulter:
Issue #4353

Add a check when looking up the `org.freedesktop.Notifications` service to see if the it actually exists before trying to fetch it

Add a check on QDBus service lookup in screengrabber freedesktop

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -61,8 +61,18 @@ void ScreenGrabber::freeDesktopPortal(bool& ok, QPixmap& res)
 {
 
 #if !(defined(Q_OS_MACOS) || defined(Q_OS_WIN))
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
Can you please remove the translation file updates? That will mess up the weblate backend.

Can you post a picture of what this looks like for a system that lacks a notification manager? Otherwise looks ready to merge.

samwcoulter:
@borgmanJeremy , amended the commit to remove the translation file changes. 

On a system with no notification manager, no desktop notifications will appear even if the desktop notification option is enabled. If you'd like I can add a warning log if that is the case. 

Here is a screenshot of my desktop immediately after taking a screenshot. No notification appears:

borgmanJeremy:
Awesome thank you!

As soon as this CI finishes ill merge.

PR #4355, opened 2025-11-01
Fix copy failure on GNOME

no7076:
On GNOME, it briefly creates a standard, invisible helper window to provide a trusted context for the clipboard operation (I think this is needed for Mutter). Also implemented a custom QMimeData subclass for GNOME clipboard manager to cache the data before the application exits.

Fixes #4116, #4298

Issue #4298, opened 2025-10-17
Screenshot copying to clipboard doesn't work


Borfak:
### Flameshot Version

Flameshot v13.1.0

### Installation Type

Flatpak from Flathub

### Operating System type and version

Fedora Linux 42

### Description

On a Fedora 42 GNOME Wayland session, copying a screenshot to the clipboard (using Ctrl+C or the copy button) does not work. The application shows a notification that the image has been copied, but the clipboard remains empty. Saving the screenshot to a file works correctly. The issue persists even with the `wl-clipboard` package installed.

### System Information

Fedora Linux 42
GNOME, Wayland

falcie1337:
I second this, having the exact same problem on my end using Ubuntu 25.10 on GNOME, Wayland. The .deb and .appimage are also giving me trouble, even though it is different, still have to open a separate issue for that.

All3xJ:
I have the same problem

hydr0nium:
Same problem. Maybe similar to #4285. I did some analysis on my end finding this:

> After some testing I am pretty sure that the problem is not with the pasting (at least for me) but rather that flameshot freezes for a couple of seconds. I can not interact with flameshot at all for like 10s after pressing ctrl-c to safe the screenshot to clipboard.

hydr0nium:
Until a fix I have to switch to `shutter`


**Update**:
After updating my system just now and restarting, flameshot is working again. Maybe some other package broke flameshot. Anyway for me it is working again

sorashi:
#4116

mmahmoudian:
The #4363 have been merged. It would be great if users can try the latest builds and provide feedback if this has fixed the clipboard issue and copying in Gnome

sorashi:
@mmahmoudian works perfectly on Arch Linux with Gnome 49.2, flameshot commit e9a8aed8

mmahmoudian:
@sorashi thanks for testing

Bubu-Droid:
@mmahmoudian The issue persists for me. I use Arch with i3. I have tried using `flameshot full -c` which fails to copy to the clipboard. Here is the output from the CLI:
```
QThreadStorage: entry 1 destroyed before end of thread 0x5641f9d99eb0
QThreadStorage: entry 0 destroyed before end of thread 0x5641f9d99eb0
```
I asked my friend to try it out on his Debian build and it runs perfectly fine for him. This is what his output shows:
```
[...]

PR #4363, opened 2025-11-04
Fix copy failure on GNOME via GUI

### LINKED ISSUE #4116: `Copy selection to clipboard` not work any more on GNOME wayland (mutter 48.4)
### Flameshot Version

flameshot-git-r2098.93d456d2-1

### Installation Type

Arch Linux on GNOME wayland (mutter 48.4-1)

### Description

Click the `Copy selection to clipboard` button (Ctrl + C),  not copy the current selection into system clipboard without any error tips, even, I can see the notification said:  `Capture saved to clipboard`, but system clipboard not update.

Save to file still work.

This issue happen after recent rolling update on arch linux.

I tried delete the `.config/flameshot/flameshot.ini`, no luck.

Thanks

### Steps to reproduce

_No response_

_No response_

### System Information

1. Arch Linux on GNOME wayland (mutter 48.4-1)
2. one 4K monitor

no7076:
My suspicion is that GNOME and specifically Mutter requires specific trusted context for the QtClipboard (from 'kf.guiaddons: Could not init WaylandClipboard, falling back to QtClipboard.') clipboard operation. This means that a visible window must provide the clipboard data. Since the daemon has no visible window, this is a workaround that makes the 'flameshot gui' window handle the copying with a slight delay using QTimer to ensure that GNOME has requested the clipboard data and saved it.

This seems like a better solution to #4355 as no helper windows are needed, since we already have the 'flameshot gui' window and can use it. I've limited it to GNOME only as I cannot easily test in other environments.

Fixes #4116, #4298

### DIFF
--- a/src/utils/screenshotsaver.cpp
+++ b/src/utils/screenshotsaver.cpp
@@ -25,11 +25,11 @@
 #include <QBuffer>
 #include <QClipboard>
 #include <QFileDialog>
+#include <QGuiApplication>
 #include <QMessageBox>
 #include <QMimeData>
-#include <QStandardPaths>
-#include <QGuiApplication>
 #include <QPointer>
+#include <QStandardPaths>
 #include <QTimer>
 #include <qimagewriter.h>
 #include <qmimedatabase.h>
@@ -275,7 +275,6 @@ private:
     QPointer<QWidget> m_owner;
 };
 
-
 bool saveToClipboardGnomeWorkaround(const QPixmap& pixmap, QWidget* keepAlive)
 {
     auto* mimeData = new ClipboardWatcherMimeData(pixmap.toImage(), keepAlive);
@@ -286,7 +285,8 @@ bool saveToClipboardGnomeWorkaround(const QPixmap& pixmap, QWidget* keepAlive)
 
     // Safety net: force close after 500ms if compositor never fetches
     QTimer::singleShot(500, keepAlive, [keepAlive]() {
-        qWarning() << "GNOME workaround timed out, compositor did not request clipboard data within 500ms. Force closing.";
+        qWarning() << "GNOME workaround timed out, compositor did not request "
+                      "clipboard data within 500ms. Force closing.";
         if (keepAlive)
             keepAlive->close();
     });
--- a/src/utils/screenshotsaver.h
+++ b/src/utils/screenshotsaver.h
@@ -17,4 +17,3 @@ void saveToClipboard(const QPixmap& capture);
 // GNOME Wayland: keeps the widget alive until clipboard data is fetched
 bool saveToClipboardGnomeWorkaround(const QPixmap& pixmap, QWidget* keepAlive);
 bool saveToFilesystemGUI(const QPixmap& capture);
-
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -629,7 +629,8 @@ void CaptureWidget::closeEvent(QCloseEvent* event)
             m_clipboardWorkaroundDone = true;
             m_context.request.removeTask(CaptureRequest::COPY);
             AbstractLogger::info()
-              << "GNOME Wayland detected; keeping capture window alive until clipboard data is fetched.";
+              << "GNOME Wayland detected; keeping capture window alive until "
+                 "clipboard data is fetched.";
             saveToClipboardGnomeWorkaround(pixmap(), this);
             return;
         }


WizardlyBump17:
I can confirm this fixes it on Ubuntu 25.04 with Wayland
image

EDIT: weirdly when trying to take a screenshot from the keyboard it just sends an error notification. I am going to investigate that tomorrow

WizardlyBump17:
> I can confirm this fixes it on Ubuntu 25.04 with Wayland image
> 
> EDIT: weirdly when trying to take a screenshot from the keyboard it just sends an error notification. I am going to investigate that tomorrow

Well, that is interesting. If I set the command to be `bash -c "flameshot gui"` it works fine. If I create a file with `flameshot gui` as its content and make it executable, it also works fine, but setting the command to be just `flameshot gui` doesnt...

B4T3S:
> > I can confirm this fixes it on Ubuntu 25.04 with Wayland image
> > EDIT: weirdly when trying to take a screenshot from the keyboard it just sends an error notification. I am going to investigate that tomorrow
> 
> Well, that is interesting. If I set the command to be `bash -c "flameshot gui"` it works fine. If I create a file with `flameshot gui` as its content and make it executable, it also works fine, but setting the command to be just `flameshot gui` doesnt...

I don't think this is a new issue though. I had to use the `bash -c` workaround a while ago when I first started using flameshot

mmahmoudian:
Thanks for the PR. This looks like a much cleaner approach. I wonder if 150 nanoseconds in QTimer is  arbitrary, based on trial and error, or it has some connections with mutter's interal logic?

All3xJ:
I tried to compile and install it, but clipboard copy still does not work. I have Fedora 42

no7076:
> Thanks for the PR. This looks like a much cleaner approach. I wonder if 150 nanoseconds in QTimer is arbitrary, based on trial and error, or it has some connections with mutter's interal logic?

It is arbitrary. Unfortunately I couldn't find a more precise method for this, so I just arrived at this 150ms delay. I'd be happy if someone comes with something more deterministic, maybe I can look into it further too in the following days.

> I tried to compile and install it, but clipboard copy still does not work. I have Fedora 42

I just tested this on a freshly installed Fedora 42 Workstation and clipboard copy worked as expected after the patch. Could you confirm you've fully rebuilt after applying the patch, and also ensure that you're using the newly-built binary?

borgmanJeremy:
Thanks for this, it at least can put us on the right track. I'm a bit hesitant to merge it since its a somewhat arbitrary delay. 

Is there a reason the full screen capture widget cannot act as the visible window before its closed? For example have the clipboard data saved while the UI is fullscreen, before its closed?

Do you have any references from Gnome that explains this behavior we could link to to keep track of it?

no7076:
Thanks @borgmanJeremy for the feedback, I took another shot at this.

The arbitrary 150ms delay is now gone. The GUI capture widget now provides the clipboard data directly and stays alive only until the compositor actually requests it. For this deterministic behavior, I introduced a small helper class, `ClipboardWatcherMimeData`, which inherits from `QMimeData`. This subclass overrides `retrieveData()` so we get a callback the moment Mutter pulls the image bytes. When that method is invoked, we immediately signal the capture widget to close itself. The 500ms timer that’s still in the code is only a safety net so we don’t hang forever if GNOME never pulls the data; under normal conditions this shouldn't fire.

> Is there a reason the full screen capture widget cannot act as the visible window before its closed? For example have the clipboard data saved while the UI is fullscreen, before its closed?

That’s exactly what happens now: the capture window itself calls `QClipboard::setMimeData()` while it’s still the focused, visible client, then hides and only exits once GNOME has read the data. The dbus+daemon path is now never used for copying on GNOME.

> Do you have any references from Gnome that explains this behavior we could link to to keep track of it?

The behaviour comes from Mutter’s Wayland clipboard implementation: [src/wayland/meta-wayland-data-device.c](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/wayland/meta-wayland-data-device.c), the key method is `meta_wayland_data_device_set_selection()`, which checks ([I think here](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/wayland/meta-wayland-data-device.c#L1158-1163)) whether the requesting client is currently the focused one; if not, the selection request is rejected or the source is cancelled. This is why the copy fails on GNOME when Flameshot's GUI window closes before calling `setMimeData()`.

Once GNOME has accepted the selection, its clipboard manager copies the data into its own buffer so it can keep serving it after the original client goes away. That logic is defined in [src/core/meta-clipboard-manager.c](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/core/meta-clipboard-manager.c), primarily in `owner_changed_cb()`. The compositor iterates over the offered MIME types, chooses one (`image/png` in our case), and requests the data via `meta_selection_transfer_async()`. If the original owner disappears, the manager takes over the clipboard using its cached data.

Just a heads up that I’m by no means deeply familiar with Mutter. But with the current patch the copy succeeds on GNOME 48.

borgmanJeremy (commented):
Thanks! This approach makes much more sense to me in the long run. I want to test on gnome before merging but can do that in the next couple of days. Thanks again.

simonhammes:
> Thanks @borgmanJeremy for the feedback, I took another shot at this.

Just to give an additional data point: I tested your changes on Fedora 43 (gnome-shell 49.1, mutter 49.1.1) and copying to the clipboard worked flawlessly :tada:

borgmanJeremy:
Sorry about the clang-format, I'm ready to merge this once thats resolved (assuming the rest of the CI passes).

no7076:
Of course, applied clang-format, hopefully it is good now.

pmorch:
This works on NixOS 25.11 under Gnome with Wayland too, where the stock v13.3.0 was broken. Here is the nix overlay file I used to build it: [flameshot-on-gnome-wayland.nix](https://github.com/user-attachments/files/24443014/flameshot-on-gnome-wayland.txt)

(I had to rename it to a `.txt` file for github to allow me to upload it, but it should end in `.nix`)

B4T3S:
Does anyone know when we can expect a new release? I would love to have this fix available without building from source, but I couldn't find anything regarding a release schedule ^^

mmahmoudian:
@B4T3S no stable release is planned yet. There are few Wayland issues that we are trying to fix before a release. Stay tuned, and until then , building from source is super easy and straightforward:

 https://flameshot.org/docs/installation/source-code/

B4T3S:
> @B4T3S no stable release is planned yet. There are few Wayland issues that we are trying to fix before a release. Stay tuned, and until then , building from source is super easy and straightforward:
> 
> https://flameshot.org/docs/installation/source-code/

Alright! Thanks for the reply ^^

Just would've been nice to be able to rely on my package manager to update for me, instead of having to check in every now and then to see when the new release is out and I can switch back

mkomko:
> Just would've been nice to be able to rely on my package manager to update for me, instead of having to check in every now and then to see when the new release is out and I can switch back

Haha, same for me :wink:

brianjmurrell:
This is a serious enough bug that it deserves a bug-fix release at least.

> @B4T3S no stable release is planned yet. There are few Wayland issues that we are trying to fix before a release. 

Can we not let the perfect be the enemy of the good?  New releases don't need to wait for everything known to be fixed.  You will never do a new release if you follow that philosophy.  New releases should be made any time there is any bugfix available, IMO.

Anyone wanting a new package on Fedora could go vote on the [Fedora Bugzilla issue](https://bugzilla.redhat.com/show_bug.cgi?id=2437059) I just opened to have this patch landed to their package while we wait here for a new release.  While you wait for that, I have rebuilt the current Fedora flameshot with this patch included in my [copr repo](https://copr.fedorainfracloud.org/coprs/brianjmurrell/fedora/) for Fedora 42, 43 and rawhide.

PR #4371, opened 2025-11-06
Add support for logging to a log file in the qt standard state directory

samwcoulter:
By default, logging to file is disabled. It can be enabled via the configuration menu. Support for filtering by log level severity and choosing a different path to save the  log files has also been added.

The log file implements a simple roller that saves only the two most recent sets of logs generated by flameshot. This should stop log messages from eating users' disk.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -539,3 +539,7 @@ if (APPLE)
 
 
 endif ()
+
+if(${Qt6Core_VERSION} VERSION_GREATER_EQUAL 6.7)
+    target_compile_definitions(flameshot PRIVATE QT_STATE_DIR_SUPPORTED=1)
+endif()
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -2,7 +2,9 @@
 // SPDX-FileCopyrightText: 2017-2019 Alejandro Sirgo Rica & Contributors
 #include "generalconf.h"
 #include "src/core/flameshot.h"
+#include "src/utils/abstractlogger.h"
 #include "src/utils/confighandler.h"
+#include "src/utils/logfile.h"
 #include <QCheckBox>
 #include <QComboBox>
 #include <QFile>
@@ -56,6 +58,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initAntialiasingPinZoom();
     initUndoLimit();
     initInsecurePixelate();
+    initLogToFile();
 #ifdef ENABLE_IMGUR
     initCopyAndCloseAfterUpload();
     initUploadWithoutConfirmation();
@@ -121,6 +124,9 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
     if (allowEmptySavePath || !config.savePath().isEmpty()) {
         m_savePath->setText(config.savePath());
     }
+    m_logToFile->setChecked(config.logToFile());
+    m_logFilePath->setText(config.logFilePath());
+
 #if defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
     m_showTray->setChecked(!config.disabledTrayIcon());
 #endif
@@ -244,6 +250,7 @@ void GeneralConf::resetConfiguration()
     if (reply == QMessageBox::Yes) {
         m_savePath->setText(
           QStandardPaths::writableLocation(QStandardPaths::PicturesLocation));
+        m_logFilePath->setText(LogFile::defaultLogFilePath());
         ConfigHandler().setDefaultSettings();
         _updateComponents(true);
     }
@@ -706,6 +713,33 @@ void GeneralConf::changeSavePath()
     }
 }
 
+void GeneralConf::changeLogFilePath()
+{
+    QString path = ConfigHandler().logFilePath();
+
+    if (path.isEmpty()) {
+        path = LogFile::defaultLogFilePath();
+    }
+
+    path = QFileDialog::getExistingDirectory(
+      this,
+      tr("Choose a Folder"),
+      path,
+      QFileDialog::ShowDirsOnly | QFileDialog::DontResolveSymlinks);
+
+    if (!QFileInfo(path).isWritable()) {
+        QMessageBox::about(
+          this, tr("Error"), tr("Unable to write to directory."));
+        path = QString();
+    }
+
+    if (!path.isEmpty()) {
+        m_logFilePath->setText(path);
+        ConfigHandler().setLogFilePath(path);
+        LogFile::resetLogFile();
+    }
+}
+
 void GeneralConf::initCopyPathAfterSave()
 {
     m_copyPathAfterSave = new QCheckBox(tr("Copy file path after save"), this);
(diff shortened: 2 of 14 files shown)


samwcoulter:
Looks like the failing builds is caused by some builds only having access to older version of Qt6. The `StateLocation` enum is available in QT 6.7+. 

```
/build/flameshot-13.3.0/src/config/generalconf.cpp:719:56: error: ‘StateLocation’ is not a member of ‘QStandardPaths’
  719 |       QStandardPaths::writableLocation(QStandardPaths::StateLocation);
```
I can either add a CMAKE branch to see if it's available, or just use `AppLocalDataLocation` in all cases, which is something like:

```
"~/.local/share/<APPNAME>", "/usr/local/share/<APPNAME>", "/usr/share/<APPNAME>"
```
or
```
"C:/Users/<USER>/AppData/Local/<APPNAME>", "C:/ProgramData/<APPNAME>", "<APPDIR>", "<APPDIR>/data", "<APPDIR>/data/<APPNAME>"
```
or
```
"~/Library/Application Support/<APPNAME>", "/Library/Application Support/<APPNAME>". "<APPDIR>/../Resources"
```

Putting logs in the "data" location isn't terrible, but it's also not ideal. I'll see how tricky a cmake configuration branch would be.

samwcoulter:
Fixed the clang format with latest push. The version I was using before is newer and makes some strange decisions.

samwcoulter:
I added a ifdef switch to check the QT6 minor version to see if the `StateLocation` enum exists. Also fixed the default initialization of the log file path in an empty config.

samwcoulter:
Here's a picture of the new entries in the configuration menu (default configuration): 
flameshotConfig


And here are example contents of the log file (after enabling logging and setting loglevel to info in the configuration)

```
[Sun Nov 09 08:38 PM] sam@stash-pc:~[main]$ cat ~/.local/state/flameshot/flameshot/flameshot.log
2025-11-09T20:51:48.856:[info]: Capture saved to clipboard.
2025-11-09T20:51:56.550:[info]: Capture saved as /home/sam/Pictures/2025-11-09_20-51.png
```

mmahmoudian:
Thanks for the PR and tracking the builds. This is a very welcome change imho.

The log can have two use-cases:
1. The user can travk what they do
2. The user teport to us a bug

I wonder if there are ways to improve the UX and make it more simple for the user. For example if user can use the About window to also open the log file path in their file manager, or have a Save As window to save the log file into the party they want.

samwcoulter:
@mmahmoudian My pleasure!

I can add a shortcut button to the about page, but if we do that I think we should enable logging by default. 

~~I'll also see if I can set up `stderr` and `stdout` redirection into the log file, as that info would probably be useful during debugging if users are going to be instructed to upload their log files.~~ Edit: This proved to be infeasible in a cross platform safe way

samwcoulter:
@mmahmoudian I added a button in the about page to open the logging directory:

flameshotAbout

I also enabled logging to file by default and set the default level to info. Now you should be able to direct users to upload the log file there without them having to first enable file logging.

I also added some other safety checks and improvements.

samwcoulter:
@mmahmoudian Anything else I can do to get this merged?

mmahmoudian:
@samwcoulter Thanks for pinging me, and thanks for caring about cross-platform part. This looks good to me. I believe it is now ready for @borgmanJeremy to review this PR. He might also have some further ideas about this as well to make this even better 🤓 .

mmahmoudian:
Upon running, I opened About and clicked on the button and was faced with this error. If this is the intended path, which I'm not sure if is a proper place to have log files, I think the path should be created upon running the software if it does not already exist. This of course goes away after the first capture. So not a major issue.

image


I also find the naming in the config a bit confusing, in my head the logging filter should be about verbosity. So "error" only saves the errors, but not warnings or general stuff like capturing the screenshot, "warning" should be warnings + errors, and info should be everything. If you have the same understanding and we are on the same page, I suggest reversing the order of the drop-down box in config, and set the default to "Error".

As a last comment, I wonder if we should do something about log rotation. Maybe we should always keep the last 1k lines, although for bug report I don't think we would ever need anything longer than 50 lines.

samwcoulter:
@mmahmoudian Oh I see, the default path is set correctly but until a log message is sent that path may not exist. I'll fix that.

As far as I understand, that path is the correct path to have log files (on linux), the path is found using the QT6 standard path function for the "state" directory, which respects the [XDG Base Directory specification](https://specifications.freedesktop.org/basedir/latest/). On other platforms QT6 should also supply an appropriate directory path for such files.

Log levels in the configuration: The way it currently works is that you specify a log level and everything at that level or below it gets saved. So if you specify "Error" only error messages get saved, if you specify "Warning" both warning and error messages get saved, and of course if you specify info all messages get saved. I can reverse the drop down order, and add a tool tip or something for a detailed explanation. 

Log rotation is currently implemented per application startup, every time you start the app (or specify a new logging directory in the configuration) a new `flameshot.log` file is created. If that log file already exists, it is moved to `flameshot.log.old` in the same directory before the new one is created. If `flameshot.log.old` already exists, it is deleted. This preserves the log messages from the last 2 application startups. I imagine for a bug report we could instruct the user to start flameshot, perform whatever actions expose the bug, then close flameshot and send us the logging file.

Log Rotation was implemented this way due to simplicity, but I would be happy to implement a different proposal. For example we could also allow the user to configure the "max log file length" and "max log file age", and then write to a time stamped log file until we reach the max length, then open a new timestamped log file and delete all log files present older than the configured age. That is how frameworks like log4net generally configure rotation, but it seemed like overkill for flameshot.

borgmanJeremy:
I can take a closer look at this tonight. Have you tested on windows? I will be able to test on linux + macos.

mmahmoudian:
@samwcoulter 

> Log rotation is currently implemented per application startup,

That makes sense. The rest was clear for me, but somehow it didn't click for me that the log is for each run. Sorry for that.



> I can reverse the drop down order, and add a tool tip or something for a detailed explanation.

A tooltip is a near idea. But considering the log rotation, keep the default at "info". I can later explain this in our documentation as well.

Thanks for your valuable contribution.

samwcoulter:
@borgmanJeremy @mmahmoudian 

I'm happy to help. I use flameshot almost every day.

Latest changes:

- I fixed the bug that occurred when you tried to open the logging directory from the about menu before it was created by creating the path if it didn't already exist
- That same shortcut is now disabled if logging to file is disabled, and I added a tooltip instructing the user to enable logging to file in the configuration menu
- I set the default log level back to "Info"
- I updated the language in the configuration menu to be easier to read. I sometimes forget not everyone is a developer who is intimately familiar with the info < warn < err log level hierarchy. The language now should be clear even to non technical users.
- I updated the way the default log file path is found. Before we were using the Qt6 `StandardPath::StateLocation`, which rendered to something like `~/sam/.local/state/flameshot/flameshot`. I changed it to be `StandardPath::GenericStateLocation + '/flameshot'` which will now render to `~/sam/.local/state/flameshot`.

borgmanJeremy:
This looks great! Thanks again

PR #4381, opened 2025-11-10
bump CI to see if that picks up newer Qt in appimage

borgmanJeremy:


QuentiumYT:
I can confirm that using a recent OS will not break on previous Ubuntu versions because libc & other dependencies are bundled into the AppImage :)

mmahmoudian (approved):

PR #4389, opened 2025-11-13
feat: color selection based on angle with widget

livethg:
The current process for selecting a new colour requires opening the colour picker with a right click and then selecting a colour with a left click on the chosen disk. This quickly become tedious when multiple colours are involved in the drawing.

This pull request aims to fix this issue by having the colour picker chose the colour that corresponds to the current orientation of the cursor, removing the requirement for a second mouse press.

This also fixes an inconvenience (possibly a bug) where if no tool is selected and the wheel is spawned on the sides of the selection panel, when choosing a colour outside the bounds of the selection, the selection will consider the left click as a new selection, effectively removing the previous selection area.

### DIFF
--- a/src/tools/capturetool.h
+++ b/src/tools/capturetool.h
@@ -87,7 +87,7 @@ public:
     {}
 
     // TODO unused
-    virtual void setCapture(const QPixmap& pixmap){};
+    virtual void setCapture(const QPixmap& pixmap) {};
 
     // Returns false when the tool is in an inconsistent state and shouldn't
     // be included in the tool undo/redo stack.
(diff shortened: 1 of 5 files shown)


mmahmoudian:
Thanks for the PR. This is an interesting approach. I just wonder how it behaves on touch devices. We have users that use Flameshot on tablets and touch screens, and as far as I understand, the concept of holding right-click depends on the interface. In those with pen/stylus, this is typically easier than those using their fingers.

Alternatively, this can be something to be configured by the user.

I'm interested to know @borgmanJeremy opinion over this PR. If/when @borgmanJeremy approves, we can ask as for feedback from the community who use tablets and touch devices.

Issue #4390, opened 2025-11-13
Cannot correctly install in newer Ubuntu releases with t64 (questing)


QuentiumYT:
### Flameshot Version

Flameshot 13.3.0

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Linux Ubuntu Questing (25.10)

### Description

Ubuntu decided to roll newer libraries and software with t64 suffix for 64 bit time values, but they kinda mixed several dependencies and it is very hard for a developer to maintain these needs. I think they choose to do 64 bit time values now without suffix for newer Ubuntu releases afaik. Is it possible to do packaging using the OR `|` value in debian control?

Like `libqt6dbus6 | libqt6dbus6t64` as libqt6dbus6t64 is no longer available in the Ubuntu apt repos, so it will satisfy multiple versions?

Here are the logs when installing the 24.04 deb version:
```
Preparing to unpack flameshot-13.3.0-1.ubuntu-24.04.amd64.deb ...
Unpacking flameshot (13.3.0-1) over (12.1.0+ds-2) ...
dpkg: flameshot: dependency problems, but configuring anyway as you requested:
 flameshot depends on libqt6dbus6t64 (>= 6.2.0~); however:
  Package libqt6dbus6t64 is not installed.
 flameshot depends on libqt6gui6t64 (>= 6.4.0); however:
  Package libqt6gui6t64 is not installed.
 flameshot depends on libqt6network6t64 (>= 6.2.0~); however:
  Package libqt6network6t64 is not installed.
 flameshot depends on libqt6widgets6t64 (>= 6.2.0~); however:
  Package libqt6widgets6t64 is not installed.
```

Note: it works fine if I use 22.04 deb version because t64 were not a thing back in the day :)

### Steps to reproduce

1. Install latest flameshot 24.04 from github releases in a newer version of Ubuntu
2. See unmatch dependencies

### System Information

Linux Ubuntu Questing (25.10)
2K monitor
Wayland

borgmanJeremy:
Sure can you make a PR?

QuentiumYT:
Will look at it tomorrow and test that

QuentiumYT:
Well, this should have been generated with `${shlibs:Depends}, ${misc:Depends}`, as you can see in ubuntu packages repos, only noble have the t64 version suffix.

https://packages.ubuntu.com/jammy/libqt6gui6
https://packages.ubuntu.com/noble/libqt6gui6t64

[...]

PR #4393, opened 2025-11-14
Explicit depends to fix t64 packaging in Ubuntu >noble (#4390) + drop Ubuntu focal + bump dependencies

QuentiumYT:
As discussed in #4390, noble renamed its packages with t64 to indicate 64 bit for time. But they got back to the original naming to prevent package issues in the future. As a result, many (if not all) packages ending with t64 have also the non-suffix alias. By explicitly defining the runtime dependencies without t64, it allows Ubuntu versions superior to noble to use the latest dependencies (from Noble) and not have any conflict or missing dependencies.

Bumping to CMake 3.16 in the meantime ensure more recent features, but current runners (jammy & noble) have even more recent versions.

| | focal | jammy | noble |
| - | - | - | - |
| Default CMake version | 3.16.3 | 3.22.1 | 3.28.3 |

I still have to test on my OS matrix, here are the workflow packages https://github.com/Quentium-Forks/flameshot/actions/runs/19370660408

Hope you will find this useful :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@
       <img alt="Get it from the Snap Store" src="https://snapcraft.io/static/images/badges/en/snap-store-black.svg" />
     </a>
     <a href="https://flathub.org/apps/details/org.flameshot.Flameshot">
-      <img height="60" alt="Download on Flathub" src="https://flathub.org/assets/badges/flathub-badge-en.svg"/>
+      <img height="60" alt="Get it on Flathub" src="https://flathub.org/api/badge?locale=en"/>
     </a>
   </p>
 </div>
@@ -416,7 +416,7 @@ Also you can open and build/debug the project in a C++ IDE. For example, in Qt C
 
 - Qt >= 6.2.4 (available by default on Ubuntu Jammy)
   + Development tools
-- GCC >= 12
+- GCC >= 11
 - CMake >= 3.22
 
 #### Run-time


borgmanJeremy:
Thanks for the help! This is still in draft, but is it ready for review / merging?

QuentiumYT:
I was testing a bit on another branch, but yeah the only thing I see is manually adding dependencies and remove shlibs

QuentiumYT:
Alright, I'm good now @borgmanJeremy :) I expanded manually shlibs for global dependencies to even allow Ubuntu 20.04 versions minimum (glibc, libgcc-s1 and libstdc++6). But Ubuntu 20.04 only have Qt5, they will need to build Qt6 themselves anyway (or PPA).
Also lowering the Qt version to 6.0.0 will match rpm spec files and README requirements.

The final depends using `apt info` now shows `Depends: libc6 (>= 2.31), libgcc-s1 (>= 3.3.1), libstdc++6 (>= 10), hicolor-icon-theme, qt6-qpa-plugins (>= 6.0.0~), libqt6core6 (>= 6.0.0~), libqt6dbus6 (>= 6.0.0~), libqt6gui6 (>= 6.0.0~), libqt6network6 (>= 6.0.0~), libqt6widgets6 (>= 6.0.0~), libqt6svg6 (>= 6.0.0~)` which will be more flexible.

Installing on Ubuntu noble works fine, I can test on Ubuntu questing this evening with these latest changes before merging :)

QuentiumYT:
Alright, works fine in Ubuntu 25.10 questing as well :)

image

borgmanJeremy:
FYI I didn't kick off the copilot comments, it says it automatically happened. Is that a setting you have? I looked through our project settings and didn't see anything obvious. I will take a look at this in the next day or so. Our commitement is to build for the latest 2 ubuntu release which would be 22.04 and 24.04 right now.

Thanks!

QuentiumYT:
Yes it's a personal setting that triggers automatically. Sometimes it's useful to not miss something :)
Alright so if you support active Ubuntu versions (currently jammy and above), I can bump qt6 to 6.2.4 and a few other things. Currently, almost the whole project can support focal with the requirements I set. Tell me if I should bump the requirements as well in this PR.

borgmanJeremy:
Yeah I would prefer to use the latest Qt as possible. In April when 26.04 comes out I will also increase the Qt version and drop some #ifdef guards

QuentiumYT:
Alright, will change that and drop support for Ubuntu focal

EDIT: Workflow is fine & tested in ubuntu 22.04 & 24.04

EDIT again: Ubuntu jammy support will end in 1 year and 4 months (01 Apr 2027), so I suggest waiting a bit before dropping support :)

PR #4418, opened 2025-11-30
Fix Wayland screenshot scaling on multi-monitor setups

anzelesnik:
The newly added if condition enables correct screenshot scaling in CaptureWidget on multi-monitor setups with scaling when running the app with the XCB platform plugin on Wayland. Previously the else branch would be taken, which would apply an incorrect device pixel ratio. Since the screenshot may already be scaled, we need to convert it back to physical dimensions and compare it with desktop geometry again.

Apologies that I do not have enough time to come up with a proper explanation as to why this works. I did test with Wayland, X11 and multiple platform plugins and this change did not affect the functionality of any other configuration.


borgmanJeremy:
Closing in favor of #4498

PR #4426, opened 2025-12-04
Fix scaling issue with last-region option

Kenshin9977:
When memorising the last region captured on my Mac, I had an issue where the next screenshot's region would be smaller. Each time I did a new screenshot it would be smaller and smaller.
It seems to be a scaling issue similar to what I already saw on Windows with other softwares.
This PR fixes that (I compiled and tested this commit).

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -303,7 +303,7 @@ CaptureWidget::~CaptureWidget()
     }
 #endif
     if (m_captureDone) {
-        auto lastRegion = m_selection->geometry();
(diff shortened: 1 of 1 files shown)


mmahmoudian:
Thanks for the PR.

Out of curiosity, what is your display aspect ratio?

Kenshin9977:
Of course, it's a 16:10, 2560 * 1600 and MacOS uses it as a 1280 * 800 display (that's their "Retina" feature / marketing concept)

borgmanJeremy:
I think i fixed this with #4530 , would you mind retesting?

Kenshin9977:
I confirm it's fixed, I'll close my PR

PR #4443, opened 2025-12-14
Simplification of snapcraft.yaml

Kyuyrii:
This is just to simplify snapcraft.yaml and add common-id.

The removed plugs are already added thanks to the kde-neon-6 extension.

Common-id is for Gnome Software and Plasma Discover to know that the Snap version and the Flatpak version are the same app, instead of showing them separately.


borgmanJeremy:
Thanks!

PR #4454, opened 2025-12-22
fix: Update KDE Plasma keyboard shortcut config file

LeviSnoot:
Wasn't able to follow the current docs for [setting up keybinds in KDE](https://flameshot.org/docs/guide/key-bindings/#on-kde-plasma-desktop), the config file wouldn't import.

After a bit of digging around I found that the spec in the current repo doesn't appear to be used in current versions of Plasma, so I created and exported the shortcuts manually and placed them here with the correct file extension.
An associated PR will be linked at [flameshot-org/flameshot-org.github.io](https://github.com/flameshot-org/flameshot-org.github.io) for updated documentation.


mmahmoudian:
Sorry for the delay in approving and merging. I forgot about this PR.

Thanks again for your contributions and for keeping an eye on the project.

PR #4455, opened 2025-12-23
clang-format update on the main branch (as per development guidelines)

markcunninghamgit:
According to https://github.com/flameshot-org/flameshot/blob/master/docs/CONTRIBUTING.md you require 
```
clang-format -i $(git ls-files "*.cpp" "*.h")
```
to be run on the code base

# this update
Purely just runs clang-format on the main code base to fix the format up to the standard that's been asked of

# reason
I wanted to submit a PR for my own feature but the master branch in this project isn't currently updated with this.
Can we bring the main branch inline so i don't have to include these changes along with my own changes when i run clang-format  on my own code for my own PR?

### DIFF
--- a/src/tools/capturetool.h
+++ b/src/tools/capturetool.h
@@ -87,7 +87,7 @@ public:
     {}
 
     // TODO unused
-    virtual void setCapture(const QPixmap& pixmap){};
+    virtual void setCapture(const QPixmap& pixmap) {};
 
     // Returns false when the tool is in an inconsistent state and shouldn't
     // be included in the tool undo/redo stack.
--- a/src/utils/confighandler.cpp
+++ b/src/utils/confighandler.cpp
@@ -212,7 +212,7 @@ ConfigHandler::ConfigHandler()
         QObject::connect(m_configWatcher.data(),
                          &QFileSystemWatcher::fileChanged,
                          [](const QString& fileName) {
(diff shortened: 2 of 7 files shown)


borgmanJeremy:
Which version of clang format are you using? The issue is clang format outputs different results depending on the version. We probably need to update the readme to specify the version.

markcunninghamgit:
> Which version of clang format are you using? The issue is clang format outputs different results depending on the version. We probably need to update the readme to specify the version.

Ubuntu clang-format version 18.1.3 (1ubuntu1)
which version should be used?

borgmanJeremy:
I use a ubuntu:20.04 distrobox image with clang-format 10:

```
clang-format --version
clang-format version 10.0.0-4ubuntu1

```

markcunninghamgit:
so where do we go from here? do you want to ask people to run a pinned version of 
clang-format version 10.0.0-4ubuntu1
or upgrade the clang-format version and accept the pr or..?

ElTh0r0:
The Github CI workflow is currently using version 11 for its check: https://github.com/flameshot-org/flameshot/blob/bee989ec55e7034a782fd18bd4e694b74a62ac32/.github/workflows/clang-format.yml#L36

borgmanJeremy:
Yes for now pin to version 11.

It's always going to be a bad version for someone.

PR #4458, opened 2025-12-27
Update README.md

AtomicRobotMan0101:
trivial edit to add flatpak `PrtSc` button instructions for Ubuntu


mmahmoudian:
Thanks for your kind contribution. Two questions though:
1. Is there any particular reason that the full path for flatpak is mentioned? I believe this is not universal path on linux distros. For instance on Guix and Nix the flatpak binary is not in `usr/bin/`.
2. Are the `--branch=stable --arch=x86_64` necessary?

PR #4467, opened 2026-01-04
Update README.md, add shortcut for ubuntu 25.10

zhangfuwen:
`/usr/bin/flameshot gui` does not work under ubuntu 25.10, `systemd-run --user --scope flameshot gui` works. tested.


mmahmoudian:
I don't understand why providing absolute path shouldn't work. I guess I have to replicate this in a VM.

itnomisco:
but copying to the clipboard does not work.

mmahmoudian:
@itnomisco is this issue related this particular PR?

Issue #4062, opened 2025-07-07
Text tool is ignoring first edit


jonathan-f-silva:
### Flameshot Version

master branch - https://github.com/flameshot-org/flameshot/commit/f33d7f266c266532adbbcb924830244a59f62143

### Installation Type

Compiled from source

### Operating System type and version

Windows 11

### Description

On first editing, the text is ignoring the changes and reverting to the old contents. But on subsequent tries it changes ok.

### Steps to reproduce

1. Launch flameshot
2. Take a screenshot
3. Use text tool and type some text
4. Unselect the tool
5. Double-click on text to start editing
6. Edit the text
7. Click outside once to finish the changes
8. The text change is reverted to the original content
9. If try to edit again, it works

### Screenshots or screen recordings

https://github.com/user-attachments/assets/7182b6dc-5450-48cb-88ca-4a84449a2ae9

### System Information

Windows 11 24H2 build 26100.4349

Info copied from about box:
Flameshot v12.1.0 ()
Compiled with Qt 6.9.1
winnt: 10.0.26100
windows: 11

borgmanJeremy:
I replicated this on Linux as well on the master branch. Thanks for the detailed procedure to reproduce the bug.

borgmanJeremy:
I've made enough progress on this to know its not going to be easy to find the root cause. It's probably a one line fix once we fully understand the root cause. Here is what I've learned.

- It was introduced in this commit: 534f05b specifically in this function
```
int ConfigHandler::toolSize(CaptureTool::Type toolType)
```
- In today's code base, the code has changed quite a bit, but if you adjust that function in the config handler to always return the same size for each tool, the problem goes away. Additionally if the 2nd tool you click after creating the text is configured with the same size as the text the problem goes away. 

- If this line of code is comment out the problem also goes away: https://github.com/flameshot-org/flameshot/blob/3a206fce8b6db657cdf0e1a4fba3c26267dceb69/src/widgets/capture/capturewidget.cpp#L1445


Based on this, I am suspecting there is a race condition related to tool size and the undo/redo stack. I don't know exactly what yet, but I'm pretty sure there is a timing/race condition.

MikePryadko:
I have this problem too!
```
Flameshot v12.1.0 (Debian 12.1.0-2build2)
Compiled with Qt 5.15.12
```

PR #4472, opened 2026-01-09
Fix: Text tool ignoring first edit (#4062)

DsChauhan08:
Prevent duplicate undo commands by checking if m_captureToolObjectsBackup has already been consumed.


borgmanJeremy:
Thanks for the PR, but this did not resolve the issue for me. If I type with the text tool, deselect the text tool, double click the existing text, edit the text, click off it, it does not save the edit.

borgmanJeremy:
This behaves the same as the first PR, are you able to replicate the issue locally so you can test for it being fixed?

PR #4479, opened 2026-01-15
grim: set device pixel ratio on captured screenshots

BackSlasher:
The grim screenshot adapter was not setting the device pixel ratio on captured images, causing them to appear zoomed/scaled incorrectly on HiDPI displays.

Apply the same DPR correction logic used by the freeDesktopPortal method: compare the captured image size against expected physical and logical desktop geometries to determine the correct ratio.

# Testing
Environment: Sway on Arch Linux

Before:
image

After:
image


borgmanJeremy:
Rather than duplicating the code, could you move to a function and call the function in the two locations with duplicate code?

I don't have a DE setup for grim, so can you also leave a comment on the PR for which DE you tested in? Looks maybe like sway?

BackSlasher:
Thanks for the review, deduped and tested

borgmanJeremy (approved):

BackSlasher:
Let me fix the styling.
Should I squash or will you do on merge?

*Nitzan Raz*

borgmanJeremy:
I can squash and merge... thanks!

PR #4490, opened 2026-01-23
Readme: Include mention of gnome and how to set screenshot hotkey when installed via flatpak

zitr0y:
https://gist.github.com/linuxkathirvel/5e8bb416f487b4b47a97daa42288502f

found this solution here and it worked for me so I think it should be in the docs

should also be on the website

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@
   - [Local](#local)
   - [Global](#global)
     - [On KDE Plasma desktop](#on-kde-plasma-desktop)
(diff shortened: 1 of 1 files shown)


mmahmoudian:
@zitr0y thanks for the contribution. The issue is that many Ubuntu users don't know what Gnome is. Your changes are correct. Let's hope if helps users, otherwise I have to revert the header back to Ubuntu.

PR #4496, opened 2026-01-27
fix: check QPainter::isActive() before drawing to prevent Wayland crashes

benignobjunior:
## Problem
On Wayland (tested on KDE Plasma 6 + Qt 6.10.1), Flameshot crashes intermittently with errors like:
```
QPainter::begin: Paint device returned engine == 0, type: 2
QPainter::setRenderHint: Painter must be active to set rendering hints
QPainter::drawEllipse: Painter not active
```

## Cause
The backing store may not be ready when `paintEvent` is called on Wayland. The code was drawing without checking if the QPainter was successfully initialized.

## Solution
Add `QPainter::isActive()` checks before drawing in:
- `selectionwidget.cpp`
- `notifierbox.cpp`
- `magnifierwidget.cpp`

This is the standard Qt pattern for handling this case - if the painter isn't ready, we simply skip that paint cycle and let the next one handle it.

## Testing
- CachyOS (Arch-based) with KDE Plasma 6 Wayland
- Qt 6.10.1
- Flameshot 13.3.0
- No more crashes after the fix

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -644,6 +644,9 @@ void CaptureWidget::paintEvent(QPaintEvent* paintEvent)
 {
     Q_UNUSED(paintEvent)
     QPainter painter(this);
+    if (!painter.isActive()) {
+        return;
+    }
     GeneralConf::xywh_position position =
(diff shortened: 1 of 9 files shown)


borgmanJeremy:
thank you!

PR #4497, opened 2026-01-27
fix: use runtime detection for Wayland clipboard

benignobjunior:
## Problem
When compiled with `USE_WAYLAND_CLIPBOARD=ON`, the clipboard only works on Wayland sessions. Running the same binary on X11 (e.g., via XWayland or `QT_QPA_PLATFORM=xcb`) results in clipboard not working.

## Solution
Add runtime detection for the platform name:
- `wayland` → uses KSystemClipboard (KF6 GuiAddons)
- otherwise → uses QApplication::clipboard() (standard Qt)

## Benefits
- Distros can ship a single binary that works on both Wayland and X11
- No separate builds needed for different display servers
- Zero impact on builds without `USE_WAYLAND_CLIPBOARD`

## Testing
Tested on CachyOS (Arch-based) with KDE Plasma 6:
- ✅ Wayland native: clipboard works
- ✅ X11 via `QT_QPA_PLATFORM=xcb`: clipboard works

### DIFF
--- a/src/utils/screenshotsaver.cpp
+++ b/src/utils/screenshotsaver.cpp
@@ -171,11 +171,16 @@ void saveToClipboardMime(const QPixmap& capture, const QString& imageType)
         auto* mimeData = new QMimeData();
 
 #ifdef USE_WAYLAND_CLIPBOARD
-        mimeData->setImageData(formattedPixmap.toImage());
-        mimeData->setData(QStringLiteral("x-kde-force-image-copy"),
-                          QByteArray());
-        KSystemClipboard::instance()->setMimeData(mimeData,
-                                                  QClipboard::Clipboard);
+        if (QGuiApplication::platformName() == "wayland") {
+            mimeData->setImageData(formattedPixmap.toImage());
+            mimeData->setData(QStringLiteral("x-kde-force-image-copy"),
+                              QByteArray());
+            KSystemClipboard::instance()->setMimeData(mimeData,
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
If KGuiAddons will compile on windows and Macos I would prefer to remove it as a compile time option and always build with it. That would simplify it even more.

If its not cross platform then we can leave it like this.

benignobjunior:
Hey! Good call. I checked the KGuiAddons source and yeah, it builds on all platforms. KSystemClipboard is always compiled and just uses QClipboard when Wayland isn't available, so no platform issues there.

One thing though, I looked at the Flameshot CI and the Windows/macOS workflows don't install any KDE deps (pure Qt builds). Making KF6GuiAddons a hard dep would break those.

What about auto-detection instead?

```cmake
find_package(KF6GuiAddons QUIET)
if(KF6GuiAddons_FOUND)
    target_compile_definitions(flameshot PRIVATE USE_WAYLAND_CLIPBOARD=1)
    target_link_libraries(flameshot KF6::GuiAddons)
    message(STATUS "KF6GuiAddons found, Wayland clipboard support enabled")
else()
    message(STATUS "KF6GuiAddons not found, using Qt clipboard")
endif()
```

No cmake flag needed, if it's there it gets used. Linux distros pick it up automatically, Win/Mac CI stays untouched.

borgmanJeremy:
Thanks for your patience!

PR #4501, opened 2026-01-30
Add Windows installation instructions for Scoop

ofek:
I chose to put this before Chocolatey because that package hasn't been updated in years.

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,6 +386,7 @@ After following all those steps above, `flameshot` will open without problems in
 
 ### Windows
 
+- [Scoop](https://github.com/ScoopInstaller/Extras/blob/master/bucket/flameshot.json): `scoop install flameshot`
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
@ofek  do you know if chocolatey is maintained at all? If not we should remove it from the list

README.md
```diff
@@ -386,6 +386,7 @@ After following all those steps above, `flameshot` will open without problems in
 
 ### Windows
 
+- [Scoop](https://github.com/ScoopInstaller/Extras/blob/master/bucket/flameshot.json): `scoop install flameshot`
 - [Chocolatey](https://chocolatey.org/packages/flameshot)
```
ofek:
```suggestion
```

ofek:
It doesn't appear to be so I just removed it.

PR #4514, opened 2026-02-06
fix: pre-grab screenshot on X11 to preserve popups and dropdowns

EricBlanquer:
## Summary

- On X11 with non-KDE desktop environments (Cinnamon, MATE, XFCE, etc.), pressing Print Screen to take a screenshot causes open popups, dropdowns, and menus to close before the screenshot is captured
- This introduces an XInput2-based raw key event listener that detects Print Screen without causing any focus changes
- The screen is grabbed immediately before creating the capture widget, preserving the current UI state including open popups

## Root cause

Both the window manager's keybinding handler and X11 key grabs (`xcb_grab_key`) generate `FocusOut` events that cause transient UI elements (dropdowns, popups, menus) to close before the screenshot can be captured.

## Approach

1. A dedicated thread with its own X11 connection listens for `XI_RawKeyPress` events via XInput2. Raw events are delivered without causing any focus changes.
2. When Print Screen is detected, the screen is grabbed immediately (before any widget is created).
3. The pre-grabbed screenshot is passed to `CaptureWidget`, which uses it instead of performing its own grab.

## Build notes

- The feature requires `libxi-dev` and `libx11-dev` at build time
- Dependencies are **optional**: if not found, the feature is simply disabled and behavior is unchanged
- Controlled by the `USE_X11_SHORTCUT` compile definition
- No effect on Wayland, macOS, or Windows builds

### DIFF
--- a/src/core/x11shortcutfilter.cpp
+++ b/src/core/x11shortcutfilter.cpp
@@ -2,9 +2,30 @@
 // SPDX-FileCopyrightText: 2026 Flameshot Contributors
 
 #include "x11shortcutfilter.h"
+#include <QSet>
+#include <QString>
 #include <X11/Xlib.h>
 #include <X11/extensions/XInput2.h>
 #include <X11/keysym.h>
+
+static QSet<int> virtualDeviceIds(Display* dpy)
+{
+    QSet<int> ids;
+    int numDevices = 0;
+    XIDeviceInfo* devices = XIQueryDevice(dpy, XIAllDevices, &numDevices);
+    if (!devices) {
+        return ids;
+    }
+    for (int i = 0; i < numDevices; ++i) {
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Why do the menu's close on X11 when you press screenshot in the first place?

EricBlanquer:
> Why do the menu's close on X11 when you press screenshot in the first place?

I don't know
I cannot take this screenshot without a delay, when I use the "Print screen" button as shortcut, this dropdown button appears closed 

with a delay:
image

without delay and the dropdown open 1st:
image

EricBlanquer:
Closing this PR — the approach needs more work to properly handle multi-monitor setups. Keeping the changes locally for now.

PR #4522, opened 2026-02-17
Sort monitor previews by screen position

brianjcohen:
## Summary

- The monitor selection dialog displayed previews in Qt's default screen order (by connector name, e.g. DP-1 before DP-2), which can differ from the physical arrangement configured in display settings (e.g. KDE's kscreen)
- This sorts the preview widgets by their X geometry position so they match the user's actual left-to-right monitor layout

## Test plan

- [ ] Configure two or more monitors where the physical left-to-right order differs from the connector name order (e.g. DP-2 is physically left of DP-1)
- [ ] Trigger a screenshot and verify the monitor selection previews match the physical arrangement

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -19,6 +19,7 @@
 #include <QScreen>
 #include <QTimer>
 #include <QWidget>
+#include <algorithm>
 
 #ifdef FLAMESHOT_DEBUG_CAPTURE
 #include <QDebug>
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Thank you! This is such a great solve for 99% of use cases. If you don't mind fixing clang-format I'll merge this into the other branch and like merge it all to main tomorrow.

brianjcohen:
@borgmanJeremy Should be fixed now.

PR #4524, opened 2026-02-18
Add custom colors for key action buttons (Save, Copy, Exit) for increased usability

BJayJ7:
Basically what the title says. I noticed a good first contribution error that stated the difficulty differentiating the important buttons with the less important ones when the capture screen was small. Therefore I changed the colors of the buttons Save, Copy, and Exit to make it clearer for users.


borgmanJeremy:
Thanks for the contribution. Could you adjust this to maybe use the "contrast color" from the config menu rather than hard coding the colors? As a side note I would prefer if the source comments do not include any emoji / unicode.
image

Thanks!

PR #4531, opened 2026-02-21
Add option for portable binary (using flameshot.ini in app folder)

### LINKED ISSUE #3440: Portable "creates .ini" in "AppData" folder.
### Flameshot Version

12.1.0.1

### Installation Type

Win 7 x64 SP1

### Description

After of download the portable zip file from [here](https://github.com/flameshot-org/flameshot/releases/download/v12.1.0/flameshot-12.1.0-win64.zip), the exe still creates the .ini in "%AppData%Roaming\flameshot\flameshot.ini" instead of locate it inside the program folder.

### Steps to reproduce

_No response_

_No response_

### System Information

Same as above,

ElTh0r0:
This PR adds the option to build Flameshot using flameshot.ini stored in the application folder. Most likely this is only used on Windows to have a portable setup.

Closes #3440
Partly solving #3848 - as mentioned in the comments there, I don't know what will happen, if users won't have write permission in the folder which includes the ini.

Side notes: I aligned the Qt version used by Appveyor and GH workflow. If I am not wrong, the exe files from Appveyor (which are signed) are copied into the ZIP archive, which is created by the GH workflow when a new official release is published. For this reason, the Qt version should be the same for both.
(Appveyor seems not to provide Qt 6.10 yet; otherwise I would have switched to that version)

Additionally I removed some win32 lines from CMakeLists as we don't have any Windows 32bit builds anymore.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,15 +16,9 @@ build_script:
     set "OPENSSL_ROOT_DIR=C:/OpenSSL-v111-Win64"
     cmake -S C:\projects\source -B build -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE=Release -DENABLE_OPENSSL=ON -DUSE_PORTABLE_CONFIG=%PORTABLE_CONFIG%
     cmake --build build --parallel 2 --config Release
-    cd build
-    if "%PORTABLE_CONFIG%"=="OFF" (
-      cpack -G WIX -B package
-    )
-    if "%PORTABLE_CONFIG%"=="ON" (
-      mkdir portable_exe
-      copy src\Release\*.exe portable_exe\.
-      7z a -tzip portable_exe.zip portable_exe\
-    )
+- cmd: cd build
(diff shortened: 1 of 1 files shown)


ElTh0r0:
After testing the files, which had been compiled by Appveyor, I noticed that the exe files are not signed by SignPath.
Did my changes break the signing or are only official releases signed?

borgmanJeremy:
Only official releases are signed, it's a manual process.

src/utils/confighandler.cpp
```diff
@@ -197,10 +197,15 @@ static QMap<QString, QSharedPointer<KeySequence>> recognizedShortcuts = {
 // CLASS CONFIGHANDLER
 
 ConfigHandler::ConfigHandler()
+#ifndef USE_PORTABLE_CONFIG
```
borgmanJeremy: I think this should say USE_WINDOWS_PORTABLE_CONFIG so its not misleading on builds like appimage. other than that im good with this.

PR #4498, opened 2026-01-28
Multimonitor fix

borgmanJeremy:
This (should) universally fix issues related to dpi, monitor placement, monitor orientation, etc. It changes X11, Wayland, and windows clients to match the MacOS behavior of screenshoting one monitor at a time. I've additionally unified the backend to use xdg-portal everywhere now that it consistently works. This cleans up a good bit of code. 

I also fixed a bug related to DPI and the placement of the toolbar / help screen.

In addition to the examples below I have also tested KDE/Waylaid and Cosmic on Arch. I've also tested on MacOS.

Here are some examples

Hyprland / wlroots:
[hyprland.webm](https://github.com/user-attachments/assets/832cf2ba-6648-441c-b8e6-b9bb40e4f73f)

Ubuntu 22.04 Gnome on X11:
[ubuntu_x11.webm](https://github.com/user-attachments/assets/43b16b7e-d900-410c-8345-e32a831ed015)

Windows 10:
[Windows10.webm](https://github.com/user-attachments/assets/427e76d8-1767-4460-ba20-8c4af0d0106e)

Need to fix before merging:

- [x] CLI screen command 
- [x] Launcher should not prompt for monitor on launch
- [x] Do not allow multiple captures to be started before one is finished
- [x] Quitting monitor selection should not throw an error
- [x] fix region issue relating to pin presenting on the wrong screen
- [x] Delay option is delaying the monitor preview not the actual screenshot.

Need to investigate before merging (not blocking issues it may require larger rework)
- [x] last-region option behaves strange with multi monitor
- [x] Consider moving X11 to portal backend to unify all of linux
- [x] Add menu item to select which screen to capture and bypass preview

Known Issues:
- The cli needs a bit of cleanup related to the ```screen``` command
- When using mixed DPI, sometimes the xdg-portal returns a blurry screenshot for the monitor of higher DPI. There's nothing I can do about this, its returned by the underlying system that way. 

@mmahmoudian  and @ElTh0r0  it would be great to have you also test this before I merge it.

PR #4532, opened 2026-02-21
Add tests for #4498

Saibamen:
Add tests for #4498


borgmanJeremy:
I appreciate the effort to add tests. However, the main issue is really in the integration of Qt into the system. It's straightforward to get the screens to appear right if you already know the "true" sizes, scales, and positions. What has been tricky is how the different operating systems report those, and how they return the screenshot.

Therefore I'm not really inclined to add in these specific tests.

I do like some of the refactoring to screenshot_geometry helper functions. It makes the intentions more clear.

PR #4533, opened 2026-02-22
Add a color block and a text area below the magnifier widget to show the RGB/hex value under the mouse, and a tool for copying the hex string.

zyf2007:
I noticed that there are currently almost no suitable screen color pickers available for Wayland. And I consider Flameshot to be well-suited as a screen color picker tool, so I have created this PR.

image

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -72,8 +72,7 @@ if (APPLE)
 
     add_executable(flameshot MACOSX_BUNDLE main.cpp ${APP_ICON_MACOSX})
 else ()
-    add_executable(flameshot
-        tools/screencolorpicker/screencolorpickertool.cpp)
+    add_executable(flameshot)
 endif ()
 
 add_executable(Flameshot::flameshot ALIAS flameshot)
--- a/src/tools/CMakeLists.txt
+++ b/src/tools/CMakeLists.txt
@@ -6,7 +6,7 @@ target_sources(flameshot PRIVATE copy/copytool.h copy/copytool.cpp)
 target_sources(flameshot PRIVATE exit/exittool.h exit/exittool.cpp)
 target_sources(flameshot PRIVATE sizeincrease/sizeincreasetool.h sizeincrease/sizeincreasetool.cpp)
 target_sources(flameshot PRIVATE sizedecrease/sizedecreasetool.h sizedecrease/sizedecreasetool.cpp)
-
+target_sources(flameshot PRIVATE screencolorpicker/screencolorpickertool.h screencolorpicker/screencolorpickertool.cpp)
 if (ENABLE_IMGUR)
 target_sources(
   flameshot


mmahmoudian:
Thanks for contributing to Flameshot. In the spirit of FLOSS, I personally use [KColorChooser](https://apps.kde.org/kcolorchooser/). I personally am not against this change, due to its very minimal effect on the UI/UX. Let's see what @borgmanJeremy thinks about this as an idea and this as code.

As a side-note, it would be better if features are first discussed in a feature request issue, so that we can gather more opinions from the community.

Thanks again for the code.

**Update:**  after skimming through the code:

1. Perhaps the name `SCREENCOLORPICKER` is a bit confusing considering the existing color picker in the sidebar. Personally I think the `MAGNIFIERCOLORPICKER` is a more clear name.
2. If I understand the code correctly, you are introducing a new executable, right? That feels a bit outside the scope of a screenshot tool tbh. I have mixed feelings about this. Let's wait and see what's @borgmanJeremy feedback.

zyf2007:
> 1. Perhaps the name `SCREENCOLORPICKER` is a bit confusing considering the existing color picker in the sidebar. Personally I think the `MAGNIFIERCOLORPICKER` is a more clear name.

I haven’t finished reading the full Flameshot code yet,  maybe the `SCREENCOLORPICKER`  might be more suitable as a simple key binding rather than being introduced as a separate tool? As a pure shortcut feature, it doesn’t make sense to add an extra option in the config menu and show an icon on the screenshot interface. I’ll read through the full project later and implement it as a shortcut-only feature, not a tool.

> 2. If I understand the code correctly, you are introducing a new executable, right? That feels a bit outside the scope of a screenshot tool tbh. I have mixed feelings about this. Let's wait and see what's @borgmanJeremy feedback.

Sorry, I'm new to Qt and that new executable was accidentally added when I was working with Qt Creator—**it’s actually not needed**. I'll fix it immediately.

> In the spirit of FLOSS, I personally use [KColorChooser](https://apps.kde.org/kcolorchooser/).

Thank you for the recommendation, but the 'Pick Screen Color' button on KColorChooser is hidden on my Gnome+Wayland environment.
image

borgmanJeremy:
I agree we should not introduce another executable. We already show the the selected color in the sidebar and that can be activated with a hotkey ("g" by default). What is the issue with the existing color picker?

zyf2007:
> I agree we should not introduce another executable. We already show the the selected color in the sidebar and that can be activated with a hotkey ("g" by default). What is the issue with the existing color picker?

The new executable is actually not needed. It was accidentally added when I was working with Qt Creator because I'm new to Qt, I've removed it in the newest commit.

About the color picker, actually, I was inspired by Snipaste. By default, it displays the hex value of the color under the mouse cursor below the magnifer and I can simply press C on the screenshot gui to copy it. I think this can greatly improve efficiency for developers who frequently use the color picker function.

What's more, the color picker which activated by hotkey 'g' will cause the magnifer component to be hidden. I didn't know how to fix it, so I simply wrote the colorpicker as a tool which can be used outside the 'grab color' tool.If you are not satisfied with my work, just close the PR😊

PR #4535, opened 2026-02-24
fix: X11 multi-monitor overlay and mixed-DPI screenshot grab

### LINKED ISSUE #4171: Changes in v13.1.0 Broke Screenshot on Linux
### Flameshot Version

13.1.0-1

### Installation Type

Arch Linux x86_64 6.16.1-arch1-1

### Description

After updating from Flameshot v13.0.1 to v13.1.0, the screen capture widget overlay (that allows you to preview and edit the screenshot) incorrectly scaled and smaller than it should be. This makes it harder to edit the image and the resulting image has aliasing artifacts.

### Steps to reproduce

1. Install Flameshot v13.1.0 on Linux (tested on i3wm with X11)
2. Take a screenshot in GUI mode
3. Observe that the screenshot preview/edit widget has the wrong scale
4. After copying/saving the screenshot, it's in the wrong size and aliased

Here's a screenshot taken by v13.0.1, which functions correctly:

Image

Here's a screenshot by v13.1.0:

Image

You can observe that the screenshot edit/preview is at the wrong scale (smaller than it should be) so it's not overlaying the entire display and also causing aliasing artifacts over the text. 

### System Information

- OS: Arch Linux x86_64
- Kernel: Linux 6.16.1-arch1-1
- WM: i3 4.24 (X11)
- Display: 3440x1440 @ 160 Hz in 34" [External]

erik55:
## Summary

Three fixes for X11 multi-monitor screenshot capture in Qt6:

### 1. Overlay spans entire virtual desktop on X11

`showFullScreen()` in Qt6 sets `_NET_WM_STATE_FULLSCREEN`, which restricts the window to a single monitor. Qt5 ignored this hint when `BypassWindowManagerHint` was set, but Qt6 respects it.

**Fix:** Use `show()` on X11 instead of `showFullScreen()`. Wayland keeps `showFullScreen()` since the compositor handles per-output fullscreen correctly.

**History:** `showFullScreen()` was introduced in [commit 3199059e](https://github.com/flameshot-org/flameshot/commit/3199059e) (Jul 2017) during a refactor from `Qt::Popup` to `Qt::Tool`. Windows was fixed in [commit a9b0c213](https://github.com/flameshot-org/flameshot/commit/a9b0c213) (Jan 2018, "Fix Windows multimonitor capture") by switching to `show()`, but Linux was left unchanged because Qt5 masked the bug.

### 2. Per-monitor grab with native pixel coordinates on X11

PR #4498 unified all platforms to use xdg-desktop-portal for screenshot capture. However, the portal may not be available or functional on all X11 window managers (e.g., tiling WMs like Xmonad, i3).

**Fix:** On X11, bypass the portal and use direct per-monitor `grabWindow()` with native pixel coordinate compositing.

`QScreen::geometry()` in Qt6 returns **mixed coordinates** by design:
- **Position** = native pixels (X11 coordinates, unscaled)
- **Size** = logical pixels (native / devicePixelRatio)

This is intentional Qt6 behavior (see `QScreenPrivate::updateGeometry()` → `fromNativeScreenGeometry()` in qtbase `qscreen.cpp` line 69). Previous code used these mixed coordinates directly, producing incorrect composites at non-1.0 DPR. The fix reconstructs native geometry: position is already native, size is multiplied by DPR.

### 3. `flameshot full` returns composite without monitor selection

PR #4498 added a monitor selection dialog in `grabEntireDesktop()` for multi-monitor setups. This causes `flameshot full` to hang waiting for user input when it should capture all monitors.

**Fix:** When `preSelectedMonitor < 0` (default for `flameshot full`), return the composite screenshot directly.

## Test Matrix

| Test | Before | After |
|------|--------|-------|
| X11 Baseline DPR 1.0 | ✅ 1366×1792 | ✅ 1366×1792 |
| X11 QT_SCALE_FACTOR=1.5 | ⚠️ 911×1536 | ✅ 1367×1792 |
| X11 Xft.dpi=144 (DPR 1.5) | ⚠️ 911×1536 | ✅ 1367×1792 |
| Wayland (GNOME, portal) | ✅ works | ✅ 2646×1024 |

## Environment

- Fedora 43, Qt 6.10.1
- X11: Xmonad WM, HDMI-1 (1280×1024) + eDP-1 (1366×768), vertical stack
- Wayland: GNOME, same monitors side-by-side

Fixes #4171, #4172

### DIFF
--- a/src/core/flameshot.cpp
+++ b/src/core/flameshot.cpp
@@ -21,6 +21,7 @@
 #endif
 
 #include "src/utils/confighandler.h"
+#include "src/utils/desktopinfo.h"
 #include "src/utils/screengrabber.h"
 #include "src/widgets/capture/capturewidget.h"
 #include "src/widgets/capturelauncher.h"
@@ -130,8 +131,17 @@ CaptureWidget* Flameshot::gui(const CaptureRequest& req)
         m_captureWindow->activateWindow();
         m_captureWindow->raise();
 #else
-        m_captureWindow->showFullScreen();
-//        m_captureWindow->show(); // For CaptureWidget Debugging under Linux
+        // On X11, showFullScreen() in Qt6 restricts the window to a single
+        // monitor even with BypassWindowManagerHint. Use show() instead to
+        // allow spanning the entire virtual desktop (v12 behavior).
+        // On Wayland, showFullScreen() works correctly per-output.
+        if (DesktopInfo().waylandDetected()) {
+            m_captureWindow->showFullScreen();
+        } else {
+            m_captureWindow->show();
+        }
+        m_captureWindow->activateWindow();
+        m_captureWindow->raise();
 #endif
         return m_captureWindow;
     } else {
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -205,23 +205,75 @@ QPixmap ScreenGrabber::grabEntireDesktop(bool& ok, int preSelectedMonitor)
     return screenshot;
 
 #elif defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
-    freeDesktopPortal(ok, screenshot);
-    if (!ok) {
-        AbstractLogger::error() << tr("Unable to capture screen");
-        return QPixmap();
+    if (m_info.waylandDetected()) {
+        freeDesktopPortal(ok, screenshot);
+        if (!ok) {
+            AbstractLogger::error() << tr("Unable to capture screen");
+            return QPixmap();
+        }
+    } else {
+        // X11: Per-monitor grab in NATIVE pixel coordinates.
+        //
+        // QScreen::geometry() on X11 returns mixed coordinates:
+        //   position = native pixels (X11 coords, not scaled)
+        //   size     = logical pixels (native / devicePixelRatio)
+        // This is intentional Qt6 design (fromNativeScreenGeometry).
+        //
+        // To composite correctly with mixed DPI, we reconstruct
+        // native geometry: position is already native, size needs
+        // multiplying by DPR to get back to native.
+
+        auto nativeScreenGeometry = [](QScreen* s) -> QRect {
+            QRect g = s->geometry();
+            qreal dpr = s->devicePixelRatio();
+            return QRect(g.topLeft(),
+                         QSize(qRound(g.width() * dpr),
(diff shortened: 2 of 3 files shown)


borgmanJeremy:
Many of these changes are intentionally and I don't want to revert them.

erik55:
Hi @borgmanJeremy,

Thanks for the quick response. I understand #4498 was an intentional architectural change to unify all platforms on a per-monitor portal-based approach – and I appreciate the cleanup that came with it.

However, on X11 with multi-monitor setups, the current state has a few concrete regressions:

1. **`showFullScreen()` restricts the overlay to a single monitor.** Qt6 respects `_NET_WM_STATE_FULLSCREEN` even with `BypassWindowManagerHint`, unlike Qt5. This means users can only select a region on one monitor at a time – the overlay no longer spans the full virtual desktop. This breaks the workflow many X11 users depend on (see #4171).

2. **`flameshot full` hangs** instead of showing a monitor selection dialog, requiring a force-kill (#4172).

3. **Mixed-DPI compositing is broken** because `QScreen::geometry()` in Qt6 returns mixed coordinates (native position, logical size). The portal path doesn't account for this.

I'm happy to rework the PR to minimize the diff – for example, keeping the portal path as the default but falling back to direct `grabWindow()` only on X11 multi-monitor, and only replacing `showFullScreen()` → `show()` on X11. Would that be a more acceptable approach?

borgmanJeremy:
I'm closing this because I suspect I'm talking to a bot and I'm starting to lose patience with the AI generated PRs.

No, my goal in this is to _explicitly_ change the paradigm to only display on one monitor.

erik55:
Understandable. I will continue to use it, as it behaves exactly like version 12 and I am an X11 user (Xmonad). And no workarounds are needed with this. Without this patch flameshot was just hanging when started with the `gui` option.

Minimal patch to just get flameshot working on X11: #4538

Alex-rk:
@borgmanJeremy I had hoped this paradigm change was temporary until the scaling issues were properly solved (see [my comment](https://github.com/flameshot-org/flameshot/issues/4172#issuecomment-3969977597) in #4172), but if I understand correctly this is not possible until QT fixes their bug. Or do you want to keep the restricted behavior even if they did fix it? It would be a pity but did you explain this choice somewhere? (either way, I'll keep using v12 on my side)

PR #4538, opened 2026-02-25
fix: X11 gui hang and mixed-DPI screenshot grab

erik55:
Minimal fix for two X11-specific regressions introduced by #4498. Wayland path is untouched.

**1. `flameshot gui` hangs on X11 multi-monitor**
Qt6 `showFullScreen()` restricts the window to a single monitor even with `BypassWindowManagerHint`, preventing the monitor selection dialog from appearing. Fixed by using `show()` on X11 only (Wayland keeps `showFullScreen()`).

**2. Mixed-DPI screenshot grab broken on X11**
`QScreen::geometry()` in Qt6 returns native position but logical size. The per-monitor grab now accounts for this.

Both changes are behind `waylandDetected()` guards. The single-monitor paradigm from #4498 is preserved.

See #4535 for detailed discussion and test results.

### DIFF
--- a/src/core/flameshot.cpp
+++ b/src/core/flameshot.cpp
@@ -21,6 +21,7 @@
 #endif
 
 #include "src/utils/confighandler.h"
+#include "src/utils/desktopinfo.h"
 #include "src/utils/screengrabber.h"
 #include "src/widgets/capture/capturewidget.h"
 #include "src/widgets/capturelauncher.h"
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
For issue #1, are you saying that on some x11 desktops the monitor selection screen never appears? Can you explain more how to reproduce the issue?

erik55:
Well, I am using Xmonad, and there is no dialog asking for which monitor. It hangs forever. I guess that would be the same with xfce, i3, etc etc, but haven’t tried.

borgmanJeremy:
The first other platform I tried this on was cinnamon and it breaks the scaling there. I'm going to add a fallback path for desktops with no portal, but I am going to put it behind a config option that makes it clear its not a supported path going forward. 

The desktops I am aware of where this would be required are just i3 and qtile.

erik55:
ok. so i3 and qtile and xmonad.

borgmanJeremy:
@erik55  Can you test #4588? Make sure to enable the new option in the config menu.  I have tested only on i3 as I am not familiar with Xmonday.

PR #4423, opened 2025-12-03
change source for mac icon

cmpute:
Fixes https://github.com/flameshot-org/flameshot/issues/4422, use the high-resolution colored icon file to generate the icon pack for MacOS.

PR #4540, opened 2026-02-26
Update cached iconset for MacOS

cmpute:
This is a PR following #4423. The previous PR fixed the icons generation pipeline, but the generated dmg files still use the old iconset.

I found that when packaging, the icons actually used is the .icns file in the repo (instead of the one generated during CI). Related code:

https://github.com/flameshot-org/flameshot/blob/76d883362fa1872f3e0aa31c179c98ebbd0effff/src/CMakeLists.txt#L61-L64

I'm not sure why the icon generation fails on CI, but this PR updates the fallback iconset, which should make sure that the generated DMG files use the correct colorful icon.

The reason why .icns file is not correctly generated needs further investigation


borgmanJeremy:
Thanks!

Issue #4541, opened 2026-02-27
Subcommands `full` and `screen` are not correctly behaving after #4498


mmahmoudian:
### Flameshot Version

Flameshot v13.3.0 (76d88336)
Compiled with Qt 6.10.2

### Installation Type

User repository (AUR)

### Operating System type and version

EndevourOS

### Description

After #4498 the subcommands `screen` and `full` are misbehaving. As far as I understand, the `gui` is supposed to show the monitor selector, but:
1. the `flameshot screen -n 1` should not bring up the monitor selector, and instead should select the monitor based on the `-n`.
2. the `flameshot screen -n 1` should not bring up the monitor selector, and instead should capture the entire landscape of monitors.

I also noticed that when I run `flameshot screen -n 1` but select the other monitor, the picture is cropped based on the `-n 1` but the content is from `-n 2`.

### Steps to reproduce

```sh
# create a folder to store the screenshots
MY_PATH="$(mktemp --directory)"

flameshot screen -n 1 --path "${MY_PATH}"

flameshot full --path "${MY_PATH}"

xdg-open "${MY_PATH}"
```

### System Information

```console
❯  inxi --system --graphics --width 80
System:
  Host: pamp-precision-tower Kernel: 6.12.74-1-lts arch: x86_64 bits: 64
  Desktop: KDE Plasma v: 6.6.1 Distro: EndeavourOS
Graphics:
  Device-1: Intel CometLake-S GT2 [UHD Graphics 630] driver: i915 v: kernel
  Device-2: Advanced Micro Devices [AMD/ATI] Lexa XT [Radeon PRO WX 3200]
    driver: amdgpu v: kernel
  Display: wayland server: X.org v: 1.21.1.21 with: Xwayland v: 24.1.9
    compositor: kwin_wayland driver: X: loaded: amdgpu,modesetting
    dri: iris,radeonsi gpu: i915 resolution: 1: 1920x1080~60Hz
    2: 1920x1080~60Hz
  API: EGL v: 1.5 drivers: iris,radeonsi,swrast
    platforms: gbm,wayland,x11,surfaceless,device
  API: OpenGL v: 4.6 compat-v: 4.5 vendor: intel mesa v: 26.0.1-arch1.1
    renderer: Mesa Intel UHD Graphics 630 (CML GT2)
  API: Vulkan v: 1.4.341 drivers: radv surfaces: N/A
  Info: Tools: api: clinfo, eglinfo, glxinfo, vulkaninfo
[...]

PR #4544, opened 2026-02-28
Fixed screen subcommand to fit per screen paradigm

borgmanJeremy:
Tested on KDE Wayland and Windows (3 monitor + fractional scaling VM)


borgmanJeremy:
closes #4541

borgmanJeremy:
This should fix the inconsistencies when running screen or full with the new paradigm. Full no longer accepts "region" since it doesnt make sense to take a region of the full screenshot.

Issue #4534, opened 2026-02-23
Windows: Allow unbinding Print Screen & overriding Win+Shift+S via Low-Level Hook


kishikuun:
### Feature Description

Flameshot on Windows faces two major hotkey limitations:
- Print Screen is hardcoded via **RegisterHotKey**, making it impossible for users to unbind it via GUI/.ini configure file if they want to free up the key.
- Win + Shift + S cannot be natively bound. Windows Explorer aggressively registers this combo at startup, causing standard **RegisterHotKey** calls to fail (**ERROR_HOTKEY_ALREADY_REGISTERED**).

To demonstrate this, I wrote an lightweight C++ daemon (~17KB) that uses **WH_KEYBOARD_LL** to perfectly bypass these restrictions. It intercepts VK_SNAPSHOT and Win+Shift+S, virtually releases modifier keys (preventing sticky keys), and redirects to Flameshot's CLI (full -c and gui).
My implementation & workaround for users: [Github Repo](https://github.com/kishikuun/Flameshot_Win-Shift-S)
I understand implementing Win32-specific hooks in a cross-platform Qt app is architecturally challenging. If native integration isn't possible right now, feel free to point Windows users to my repo as a viable workaround.

Thanks for maintaining this amazing tool!

mmahmoudian:
@borgmanJeremy  @ElTh0r0 I'm curious to know your opinion on this.

@kishikuun what is the licence of your project?

kishikuun:
> [@borgmanJeremy](https://github.com/borgmanJeremy) [@ElTh0r0](https://github.com/ElTh0r0) I'm curious to know your opinion on this.
> 
> [@kishikuun](https://github.com/kishikuun) what is the licence of your project?

Oh, I forgot about this. I've decided to release this project under the MIT License.

ElTh0r0:
Thanks for sharing your workaround!

> Win + Shift + S cannot be natively bound. Windows Explorer aggressively registers this combo at startup, causing standard RegisterHotKey calls to fail (ERROR_HOTKEY_ALREADY_REGISTERED).

It is already possible to use Win+Shift+S with a registry change; see the official [Flameshot docs](https://flameshot.org/docs/guide/windows-help/#overriding-windows-default-shortcut), so in my opinion the shown Win32 hook may be a little over-engineered. Please correct me if I am overlooking something, but I can't see any additional advantages of this approach over the registry change.

[...]

Issue #4500, opened 2026-01-29
prt sc not working to open flameshot on windows


emilykal:
Hi, new user and I can't capture my screen when I press the print screen button. I disabled it on windows but it feels like I need someone to hold my hand through the entire setup. Are there any steps Im missing past installation and creating the desktop shortcut?

gonzaleztroyano:
Hello:

Having Flameshot v13.3 open, the _Print Screent_ is working for me on Windows 10. 
Could you share technical details of your setup? Thanks.

ElTh0r0:
Please checkout #4556 - testing is highly appreciated. You can find test builds in the [GH actions artefacts](https://github.com/flameshot-org/flameshot/actions/runs/22808854912) of that pull request.

PR #4556, opened 2026-03-07
Windows: Global "take screenshot" shortcut

ElTh0r0:
This PR adds multiple optimizations for Windows user for using global shortcut for calling Flameshot to take a screenshot.

1. Enable qHotkey for registering arbitrary global shortcut
2. Check if PrintScreen key is forced by Windows to call its snipping tool and provide the option to change the corresponding registry key by Flameshot to disable it (see [Flameshot docs](https://flameshot.org/docs/guide/windows-help/#how-to-disable-windows-snipping-tool-when-i-press-printscreen))
3. Option to register Flameshot as MS-SCREENCLIP application, if registered one can select Flameshot as default screenshot application. With this, one can take a screenshot using Win + Shift + S (see [Flameshot docs](https://flameshot.org/docs/guide/windows-help/#overriding-windows-default-shortcut)). Administrative privileges are required for this; a warning will be shown, if registry change fails due to missing permissions.

I had to implement some workarounds. I had to keep the 'old' GlobalShortcutFilter as I couldn't get PrntScr working with qHotkey; don't know, must be a Windows-thing, that Windows still is fetching that key and never forwarding to any third-party application. So PrntScr remains available + one can define **another** global shortcut to take a screenshot.
For registering the Win key, I had to use a nasty timer workaround ([see here](https://github.com/flameshot-org/flameshot/commit/b6c2c9e0c41cd2ebe6dbe44ec0e68eb691d61e15#diff-f869bfeb100f8862fe8d0cc86598dc595cea048d07a2ae24dd5fde1a0332fe1bR64)) to be able to call grabKeyboard(), otherwise SetShortcutDialog didn't receive the Win key. In the end the Win key is used by several other "hardcoded" Windows shortcuts, so it is difficult to find a free shortcut anyway..!

Testing this PR before merging is highly appreciated! Mainly by Windows users, but also short double check by macOS users, that I didn't destroy the qHotkey part, which is used by macOS as well.

I proposed Win+Shift+X as the default (custom) shortcut. I have no personal preference regarding this, so please let me know if you can think of a better one.

Fixes: #1341, #3783, #4534, #4500, #3462, #3465
Proposing to close #3593 and #2168 for this PR

### DIFF
--- a/src/config/shortcutswidget.cpp
+++ b/src/config/shortcutswidget.cpp
@@ -310,7 +310,7 @@ bool ShortcutsWidget::disablePrintScreenKeyForSnipping()
     return this->isPrintScreenKeyForSnippingDisabled();
 }
 
-ShortcutsWidget::initMsScreenclipCheckbox()
+void ShortcutsWidget::initMsScreenclipCheckbox()
 {
     m_registerMsScreenclip =
       new QCheckBox(tr("Register Flameshot as MS-SCREENCLIP application "


borgmanJeremy:
I've tested that this does not break hotkeys on MacOS. If I change the hotkey, I do need to restart flameshot for it to take effect. Not sure if that is the same on windows.

I still want to take some time to review the code.

ElTh0r0:
> I've tested that this does not break hotkeys on MacOS. If I change the hotkey, I do need to restart flameshot for it to take effect. Not sure if that is the same on windows.

Thanks for testing on MacOS! Restart for registering the Take-screenshot shortcut was required before as well (this is mentioned on the SetShortcutDialog too); it is also required on Windows.

> I still want to take some time to review the code.

No hurry 🙂
Still hoping for some Windows testers!

kishikuun:
Hi @ElTh0r0, thanks for the quick action and the massive effort you put into #4556 !

I downloaded the **win-x64-portable** artifact and ran a thorough test on my Windows 11 IoT. Here is the exact feedback based on my testing:

1. Standard User Mode (No Admin Rights) - Works nicely:
- The warning dialog "The registry could not be changed!" pops up perfectly when trying to register **MS-SCREENCLIP** without privileges. Excellent fallback handling here.

2. Admin Mode & **MS-SCREENCLIP** - Failing on Win 11:
I ran Flameshot as Administrator and successfully checked the **MS-SCREENCLIP** box.
- Bug 1 (UI): I couldn't clear the secondary keybinding row, so I changed it to **Shift + Space**. Trying to bind the Windows key in the UI directly results in the Meta+Shift+Shift glitch.

- Bug 2 (Functional): Even with the **MS-SCREENCLIP** box successfully checked and running as Admin, pressing Win + Shift + S globally still opens the default **Windows Snipping Tool**. Flameshot does not launch. It seems Windows 11 aggressively ignores the registry override in this scenario.

3. Print Screen / QHotkey fallback: The fallback to the old GlobalShortcutFilter for Print Screen works perfectly.

Moving Forward (A Native Hook Proposal)
Since the **MS-SCREENCLIP** Registry method appears unreliable on modern Windows builds (and strictly requires Administrator rights), would the team be open to natively integrating the **WH_KEYBOARD_LL** Win32 API Hook approach directly into Flameshot?
By utilizing a low-level keyboard hook (like the one in my standalone C++ daemon), Flameshot can intercept Win+Shift+S and Print Screen at the OS level before Windows Explorer even registers them.
- Pros: 100% reliable overriding, completely **bypasses Snipping Tool**, and **Zero Admin Rights** required.

I would be more than happy to help adapt my C++ logic into a clean Qt/C++ module for the Windows build if you guys think this is a viable path forward. Let me know what you think!
image
image
image

ElTh0r0:
Hi @kishikuun, many thanks for testing and providing feedback! Very much appreciated!

>2. Admin Mode & **MS-SCREENCLIP** - Failing on Win 11:
>I ran Flameshot as Administrator and successfully checked the **MS-SCREENCLIP** box.
>Bug 1 (UI): I couldn't clear the secondary keybinding row, so I changed it to **Shift + Space**. Trying to bind the Windows key in the UI directly results in the Meta+Shift+Shift glitch.

Other shortcuts (other than Win+Shift+S) are independent of the MS ScreenClip option. Whether or not the MS-Screenclip option is checked is irrelevant for this issue. I suspect this is a Windows-related problem. Any shortcuts that are 'hardcoded' in Windows cannot be used in Flameshot. I noticed this during testing as well when trying to register shortcuts that were already bound to Windows features. I don't think there's anything I can do if Windows is forcing its other shortcuts.

>Bug 2 (Functional): Even with the **MS-SCREENCLIP** box successfully checked and running as Admin, pressing Win + Shift + S globally still opens the default **Windows Snipping Tool**. Flameshot does not launch. It seems Windows 11 aggressively ignores the registry override in this scenario.

There is a misunderstanding: When MS-Screenclip is successfully registered, Flameshot does not automatically work with Win+Shift+S. However, it does allow you to change the default Windows screenshot application. Selecting Flameshot as the default in the Windows settings will enable it to work with Win+Shift+S. I have added a tooltip to provide a hint:
grafik
Unfortunately I can only provide a screenshot in German. You need to go to Apps -> Standard Apps -> Screenclip. But if this PR would be merged with the current implementation, the documentation would need an update for sure as well. In addition, I am fully open for suggestions rephrasing the option/tooltip to make it more clear.
grafik

> Moving Forward (A Native Hook Proposal) Since the **MS-SCREENCLIP** Registry method appears unreliable on modern Windows builds (and strictly requires Administrator rights), would the team be open to natively integrating the **WH_KEYBOARD_LL** Win32 API Hook approach directly into Flameshot? By utilizing a low-level keyboard hook (like the one in my standalone C++ daemon), Flameshot can intercept Win+Shift+S and Print Screen at the OS level before Windows Explorer even registers them.
> 
>     * Pros: 100% reliable overriding, completely **bypasses Snipping Tool**, and **Zero Admin Rights** required.
> 
> 
> I would be more than happy to help adapt my C++ logic into a clean Qt/C++ module for the Windows build if you guys think this is a viable path forward. Let me know what you think! 

I am very grateful for your offer, but with this, the shortcut will not be configurable as requested by the users, but will be hardcoded to Win+Shift+S/PrtScr. My intention was to keep it as configurable as possible without hacking into the Windows API.

**Just my personal opinion**: I would try to avoid platform-specific API code as much as possible, because nobody knows when/if MS will change something. I am not interested in integrating more of the API (ideally, I would have liked to get PrntScr working without the 'globalshortcutfilter' and remove that part!). However, the final decision how to proceed with this is not mine to make.

borgmanJeremy:
@ElTh0r0  I agree with keeping it as configurable as possible with limited platform specific code. Thank you!

PR #4563, opened 2026-03-13
Add captureAllMonitors option to skip monitor selection popup

EricBlanquer:
## Summary

PR #4498 changed `flameshot gui` to show a monitor selection popup on multi-monitor setups. While this fixes many DPI/scaling issues, some users miss the ability to capture all monitors at once (see comments in #4498).

This PR adds a `captureAllMonitors` configuration option (default: `false`) that, when enabled, restores the previous behavior — capturing all monitors without the selection popup.

### Changes
- New `captureAllMonitors` boolean option in `ConfigHandler` (default `false`)
- Checkbox in General settings with a **warning tooltip**: *"May cause display issues on multi-monitor setups with mixed DPI scaling"*
- When enabled, `ScreenGrabber::selectMonitorAndCrop` returns the full desktop screenshot
- `CaptureWidget` spans across all screens instead of a single one
- Documented in `flameshot.example.ini`

### What it does NOT change
- Default behavior remains the per-monitor selection popup from #4498
- No changes to the monitor selection UI itself
- No changes to `flameshot screen` or `flameshot full` commands

### Disclaimer
As noted in the tooltip, this option may cause display issues on setups with mixed DPI scaling. This is the same limitation that existed before #4498 and is the reason why per-monitor capture was introduced. Users enable this at their own risk.

Addresses feedback from: @franzitrone, @saalmi098, @CrazyWolf13, @mostafa-abdelbrr in #4498

### DIFF
--- a/flameshot.example.ini
+++ b/flameshot.example.ini
@@ -113,6 +113,10 @@
 ;; Use larger color palette as the default one (bool)
 ;predefinedColorPaletteLarge=false
 ;
+;; Capture all monitors at once without showing the monitor selection popup (bool)
+;; Warning: this may cause display issues on multi-monitor setups with mixed DPI scaling
+;captureAllMonitors=false
+;
 ;; Set JPEG Quality (int in range 0-100)
 ;jpegQuality=75
 ;
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -55,6 +55,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initAntialiasingPinZoom();
     initUndoLimit();
     initInsecurePixelate();
+    initCaptureAllMonitors();
 #ifdef ENABLE_IMGUR
     initCopyAndCloseAfterUpload();
     initUploadWithoutConfirmation();
@@ -105,6 +106,7 @@ void GeneralConf::_updateComponents(bool allowEmptySavePath)
     m_squareMagnifier->setChecked(config.squareMagnifier());
     m_saveLastRegion->setChecked(config.saveLastRegion());
     m_reverseArrow->setChecked(config.reverseArrow());
+    m_captureAllMonitors->setChecked(config.captureAllMonitors());
 
 #if !defined(Q_OS_WIN)
     m_autoCloseIdleDaemon->setChecked(config.autoCloseIdleDaemon());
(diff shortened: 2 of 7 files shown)


borgmanJeremy:
This will not work on most desktops so I do not want to add it back in. It's not just mixed DPI, it also has problems on cosmic, niri, and kde wayland. I understand why some people want it, but its not maintainable and will results in a lot of bug reports.

CrazyWolf13:
@borgmanJeremy Is that something you could consider keeping in with community-based support or some sort of extension?

EricBlanquer:
@borgmanJeremy  I understand the concerns about `captureAllMonitors` and mixed DPI issues. Totally fair.

However, the monitor selection popup introduced in #4498 is not great UX — it adds friction on every single capture for multi-monitor users. Most users expect the screenshot tool to just capture the screen they're working on.

What about a simpler alternative: **capture the screen where the mouse cursor is**, skipping the monitor selection popup entirely? This is what macOS does natively, and it avoids all the mixed DPI / cross-monitor compositing issues since you're always dealing with a single screen.

This could be:
- A config option like `autoSelectMonitor=true` (default `true` to match macOS behavior)
- Or simply the new default behavior, with the popup as a fallback when the option is disabled

The implementation would be straightforward — just call `QCursor::pos()`, find which screen contains that point via `QGuiApplication::screenAt()`, and skip the selection UI.

Would you be open to this approach?

borgmanJeremy:
I already looked into the monitor idea and it's not possible on Wayland .

EricBlanquer:
That's a fair point about Wayland limitations. However, according to Mozilla's Firefox telemetry data, ~93% of Linux desktop users are still on X11 today (source: [Phoronix](https://www.phoronix.com/news/Firefox-Wayland-X11-Stats)). On X11, `QCursor::pos()` + `QGuiApplication::screenAt()` works perfectly to detect which monitor the mouse is on.

Would you consider an X11-only implementation (auto-select monitor under cursor) with a fallback to the current selection popup on Wayland? That way the vast majority of users get a seamless experience, while Wayland users keep the existing behavior until the portal API catches up.

borgmanJeremy:
I need to make a x11 fallback for desktops like i3 that dont support any portal backends. As part of that let me see what the cursor selection on x11 and windows would look like.

PR #4568, opened 2026-03-14
Add QR code scan tool

rockerritesh:
## Summary

- Adds a new toolbar button that decodes QR codes from the selected screenshot area and copies decoded text to clipboard
- Uses ZXing-C++ as an optional dependency, enabled via `-DENABLE_QRCODE=ON`
- Follows the existing `AbstractActionTool` pattern with a 3-second notification label for feedback

## Details

- New files: `src/tools/qrscan/qrscantool.h`, `qrscantool.cpp`, QR icon SVGs (black/white)
- Registration: `capturetool.h` enum, `toolfactory.cpp`, `capturetoolbutton.cpp`
- All QR-related code is guarded by `#ifdef ENABLE_QRCODE` — builds without ZXing are unaffected
- Scans the selected area (or full screenshot if no selection) for QR codes
- Shows "QR code copied to clipboard" or "No QR code found" as a styled label for 3 seconds

## Test plan

- [ ] Build with `-DENABLE_QRCODE=ON` and verify QR scan button appears with icon
- [ ] Capture area containing a QR code, click QR button — verify text is copied to clipboard
- [ ] Capture area without QR code — verify "No QR code found" message appears
- [ ] Build without `-DENABLE_QRCODE` — verify QR button does not appear
- [ ] Verify notification label stays visible for ~3 seconds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,13 @@ if(ENABLE_IMGUR)
   add_compile_definitions(ENABLE_IMGUR)
 endif()
 
+option(ENABLE_QRCODE "Enable QR code scanning via ZXing" OFF)
+
+if(ENABLE_QRCODE)
+  find_package(ZXing REQUIRED)
+  add_compile_definitions(ENABLE_QRCODE)
+endif()
+
 if(DISABLE_UPDATE_CHECKER)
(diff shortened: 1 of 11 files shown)


borgmanJeremy:
Thanks for this, however, I want to keep the existing tool set minimal. This is a great idea for a plugin if we implement a plugin system.

PR #4569, opened 2026-03-14
Hide monitor selection in launcher for single-screen setups

EricBlanquer:
## Summary
- Hide the monitor label and combobox in the capture launcher when only one screen is connected (no need to choose)
- Make the launcher UI more compact: reduced margins, tighter spacing, coordinate fields capped at 70px width, combobox sized to its content


borgmanJeremy:
Thanks!

PR #4583, opened 2026-03-19
Fix HDR overexposure and Legacy Multimonotor Capture Mode

### LINKED ISSUE #3151: Text on becomes very bright and sometimes illegible while using HDR in Windows
### Flameshot Version

12.1.0

### Installation Type

Windows 11

### Description

While using HDR in Windows and taking a screenshot using Flameshot then the text becomes very bright. If the text is on white background then it becomes almost illegible.

Disabling HDR fixes the issue. Re-enabling HDR brings it back.

Although, I noticed for a short instance while fiddling around with the value of option "Opacity of area outside of selection" that the issue was gone. But after a minute or so the issue came back again. I have not been able to reproduce it since.

### Steps to reproduce

1. Enable "Use HDR" in graphical options in Windows 11
2. Take a screenshot

From https://foxholestats.com/:
image

### System Information

1. Windows 11
2. 1 monitor only

Wuerfelhusten:
Summary
Two user-facing features for Windows, both toggleable in Settings → General

Fixes #3151 - Converts HDR/wide-gamut screenshots to sRGB before compositing, preventing washed-out/overexposed captures on displays with HDR enabled. Works correctly in mixed setups (one HDR + one SDR monitor): only screens with a non-sRGB color space are converted; SDR screens are passed through unchanged.

Capture all monitors at once - Restores the pre-#4498 behavior of spanning the capture widget across all connected monitors, allowing selections and annotations to cross screen boundaries. When disabled (default), the monitor-selection UI introduced by #4498  is used instead.

Both options default to off.


borgmanJeremy:
I'm not reverting the multi monitor PR. Please submit just the HDR patch for review.

Wuerfelhusten:
Even as an optional setting? In my version the option for the legacy function is only restored if opted into it. This makes life much easier for people with 3 screens wit all of them having the same resolution and dpi settings

borgmanJeremy:
Correct, I'm trying to simplify the code and get less bug reports on stuff like this. It unifies the code paths.

Wuerfelhusten:
I'll reopen as soon as it is done

saulob:
> I'll reopen as soon as it is done

Hi @Wuerfelhusten , any news on the HDR-only patch?

You mentioned you would reopen it once it was ready, so I was wondering if you made any progress

PR #4587, opened 2026-03-22
Code quality: Consistent includes, remove unused includes

ElTh0r0:
Some code quality improvements. I know this is a "big" change, since almost every file has been modified, but ever since I first started working with the code, I've been confused about the includes. Everything worked, but the way they were used was inconsistent. 😉

I made three commits :

1. The majority of the files is using `#pragma once` as header guard, so I replaced the few `#ifndef [...] #define [...]` guards and added completely missing guards as well.
2. Replaced the very long `target_include_directories(...)` in src/CMakeLists.txt with the following and aligned all includes in all the files accordingly.

```
target_include_directories(flameshot PUBLIC
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/>  # src/
    $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/>
)
```

3. Removed unused includes

### DIFF
--- a/src/tools/launcher/applauncherwidget.h
+++ b/src/tools/launcher/applauncherwidget.h
@@ -4,7 +4,7 @@
 #pragma once
 
 // Must be included before #if def, otherwise Q_OS_WIN is unknown
-#include <QtSystemDetection>
+#include <QSysInfo>
 
 #if defined(Q_OS_WIN)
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Thanks! thats what happens in a long running multi contributor project :)

Issue #3283, opened 2023-07-27
Why we have --delay 500 in .desktop file?


JustZhenya:
### Feature Description

Just switched from custom hotkey to flameshot provided one from .desktop file.
Turns out there is a huge delay between hotkey press and gui appearance.

I checked everything and found this:
https://github.com/flameshot-org/flameshot/blob/a447b3d672ef92acb98be996d58540e36db84e35/data/desktopEntry/package/org.flameshot.Flameshot.desktop#L84

I see it was made intentionally, but why? Do we really need extra input lag? Can you remove it?

Kubuntu 23.04, Plasma 5.27.4, X11, Flameshot 12.1.0-2

JustZhenya:
I noticed this was fixed. Thank you!

PR #4597, opened 2026-03-28
Removing arbitrary delay from desktop actions

borgmanJeremy:
This delay was added a very long time ago to address a unity specific issue. For the upcoming beta I would like to remove it and see if it causes any issues. If not, it will speed up flameshot for users who start it with the desktop file.


borgmanJeremy:
@mmahmoudian  this should address #3283 . If it causes issues during the beta I can add it back in.

PR #4599, opened 2026-03-28
Fix Linux capture overlay offset

tarcisiomiranda:
Fixes a Linux issue where the capture overlay could be shifted downward relative to the real selection area, making the bottom toolbar inaccessible.

This PR keeps the capture window on the intended screen more reliably and avoids Linux fullscreen repositioning that could distort the overlay geometry. It also preserves the existing GNOME/X11 crash workaround.


borgmanJeremy:
Can you help me understand the issue this solves?

PR #4605, opened 2026-03-31
Added dbus capture

borgmanJeremy:
Added the ability to directly initiate a capture with dbus to avoid this workaround: https://github.com/flameshot-org/flameshot/issues/3365#issuecomment-4153276050


mmahmoudian (approved):
lgtm

PR #4608, opened 2026-04-01
fixing signpath

borgmanJeremy:


borgmanJeremy:
In case this comes up again, its important for a reason unknown to me that the artifact.zip be placed inside the build folder.

Issue #4610, opened 2026-04-02
[Bug] Double Scaling on Windows 11: Qt + DWM causing massive UI and shifting pins


silfsy:
### Flameshot Version

Latest Nightly / PR #4498 Build (Commit: 53d4da8)

### Installation Type

Compiled from source

### Operating System type and version

Windows 11

### Description

There is a significant scaling conflict on Windows 11 where the UI elements become massive and the coordinate math for "pins" breaks. This occurs whenever Windows display scaling is set to anything other than 100% (e.g., 125%, 150%). I personally confirmed the issue is highly reproducible and most prominent at 150% scaling.

Because Flameshot is a Qt application, the Qt engine appears to be applying its own scale, but the Windows 11 Desktop Window Manager (DWM) is simultaneously applying its own scaling on top of that. This compounding effect results in:

1. Massive UI elements that do not fit the screen properly.

2. Broken coordinate math: when pinning a screenshot, the pinned image shifts position and appears significantly larger than the captured area.

This issue persists even in the build for PR #4498, which was intended to address multi-monitor behavior. Reverting to version 12.1.0 resolves the issue, suggesting a regression in High-DPI handling in recent versions.

**Temporary Workaround**
Users can run this simplified PowerShell script to launch Flameshot with the correct environment variables. This disables Qt's internal scaling and allows Windows to handle the scaling natively without Qt interference:

```powershell
# 1. Define environment variables for the current process to bypass Qt scaling
$env:QT_AUTO_SCREEN_SCALE_FACTOR = "0"
$env:QT_ENABLE_HIGHDPI_SCALING = "0"
$env:QT_SCREEN_SCALE_FACTORS = ""
$env:QT_SCALE_FACTOR = "1"

# 2. Locate the Flameshot executable
$paths = @(
    "${env:ProgramFiles}\Flameshot\bin\flameshot.exe",
    "${env:ProgramFiles(x86)}\Flameshot\bin\flameshot.exe"
)
$flameshotExe = $paths | Where-Object { Test-Path $_ } | Select-Object -First 1

if ($flameshotExe) {
    # 3. Close existing instances to apply new environment
    Stop-Process -Name flameshot -Force -ErrorAction SilentlyContinue
    
    # 4. Launch Flameshot with fixed scaling
    Start-Process -FilePath $flameshotExe
    Write-Host "Flameshot launched with fixed scaling variables." -ForegroundColor Green
} else {
[...]

PR #4614, opened 2026-04-03
Fix pin position on Windows for scaled screen

ElTh0r0:
Potential fix for #4610
Seems the implementation for macOS and Linux is working on Windows as well! My available test setup is very simple and limited; I would appreciate it if some Windows users could test this and confirm that it is working on their machines with scaling enabled!


borgmanJeremy:
Thank you!

Issue #3463, opened 2024-01-05
macOS windows disappear when Flameshot initiates screenshot


jg3:
### Flameshot Version

Flameshot v12.1.0 (96c2c82e)
Compiled with Qt 5.15.5
darwin: 23.2.0
osx: 14.2

(I installed by downloading a .dmg file from flameshot.org; I uninstalled and reinstalled using the latest .dmg in the course of initial troubleshooting)

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

macOS 14.2.1 (23C71)

### Description

When attempting to take a screenshot all the windows on the desktop disappear when I invoke Flameshot to take a screenshot.  I have been using Flameshot happily for years, I believe this began around my first upgrade to Sonoma.

I also get a message that Flameshot needs permission in system settings for screen recording at every reboot (but that permission is already granted and persists in System Settings, so I'm not sure why that's happening or if it is related)

### Steps to reproduce

0. With flameshot running
1. Take a screenshot (using either the menu option or the keyboard shortcut)
2. Watch all the windows disappear (oh no!  why did that happen?)
3. Take a screenshot anyway (or don't, whatever)
4. Only the desktop/background items which were visible are captured in the screenshot (not surprising

### Screenshots or screen recordings

(gif demo of the issue)
![flameshot-issue-demo-20240105](https://github.com/flameshot-org/flameshot/assets/2963153/46cec83c-3cd2-4cf0-8c00-04fa6ee52351) 

### System Information

MacBook Pro 16-inch 2019 (Intel)
Built-in display (3072 × 1920)
I often use an external display at up to (3840 × 2160), but that does not seem to have any impact on this issue.

pinutsGiesen:
I have the same problem 
Happens both with the built-in display and with an external display 
MacBook Pro 14-inch 2021 (M1)
osx: 14.2

trusiik:
I have the same issue after upgrading to Sonoma. I think it's a weird privacy issue..
<img width="471" alt="image" src="https://github.com/flameshot-org/flameshot/assets/118285510/56b35fa6-7dfe-4f99-9a0c-5d09843cb12b">

pinutsGiesen:
The problem has resolved itself for me.
After the OS update to version OSX 14.3, I had the same error message like @trusiik. 
I stopped flameshot and removed it completely from the autostart.
After a restart, flameshot is now working properly again.

danpilch:
I managed to fix this with these commands:

```bash
[...]

PR #4617, opened 2026-04-04
fix(macos): use CGRequestScreenCaptureAccess instead of grabWindow for permission request

Mitnitsky:
## Problem

On modern macOS (especially 15+), Flameshot re-triggers the Screen Recording permission dialog on every launch. The user grants the permission, but the next time the app starts it prompts again.

## Root cause

The current code uses `QScreen::grabWindow()` as a hack to trigger the TCC permission prompt. This performs an actual screen capture on every launch, which modern macOS interprets as a new capture attempt rather than a no-op — re-triggering the permission dialog even when already granted.

## Fix

Replace the `grabWindow()` hack with the proper CoreGraphics APIs:
- `CGPreflightScreenCaptureAccess()` — checks if permission is already granted (no prompt)
- `CGRequestScreenCaptureAccess()` — shows the system dialog only if not yet granted

These APIs have been available since macOS 10.15 (Catalina) and are the documented way to handle Screen Recording permissions.

### DIFF
--- a/src/core/flameshot.cpp
+++ b/src/core/flameshot.cpp
@@ -7,6 +7,10 @@
 #include "qhotkey.h"
 #endif
 
+#if defined(Q_OS_MACOS)
+#include <CoreGraphics/CoreGraphics.h>
+#endif
+
 #include "config/configresolver.h"
 #include "config/configwindow.h"
 #include "core/qguiappcurrentscreen.h"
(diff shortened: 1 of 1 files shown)


Mitnitsky:
#3463 probably related

borgmanJeremy:
This looks good to me. I want to do some testing here but I plan to merge for the actual v14 release.

borgmanJeremy:
I do not get the screen permission prompt every launch, but I almost always test a from source build. Does the issue only happen when using a .dmg built in CI?

borgmanJeremy (approved):

Issue #2287, opened 2022-01-19
[Mac] - Bug window sliding after capture


vclerissj:
### Flameshot Version

Flameshot v11.0.0 (ad1bf28)
Compiled with Qt 5.15.2
darwin: 21.2.0
osx: 12.1

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

MacOs 12.1

### Description

When I take a screenShot and copy to clipboard, there are a sliding windows

### Steps to reproduce

1. Run the application
2. Take a screen shot
3. copy to clipboard with shortcut (cmd+c)
4. window sliding bug

### Screenshots or screen recordings

https://user-images.githubusercontent.com/18285026/150169330-31984342-42b4-402c-8ff9-1a4abbaf8b0b.mov

### System Information

MacOS 12.1
Monitor : 1920x1080 27" 144Hz

borgmanJeremy:
Im pretty sure this has always been the behavior and is just how MacOS works. Flameshot runs in full screen mode. Whenever your Mac switches off fullscreen mode it plays this animation.

mmahmoudian:
I don't think this is a bug. In MacOS/OSX it has always been so that if you make an application fullscreen it will be automatically moved to a new and dedicated desktop on the right-hand side of the current active desktop. So when the application terminates or goes out of fullscreen, the user will be taken back to the desktop on the left-hand side.

I suggest closing this unless there a way in the API to tell MacOS to skip the animation.

avevlad:
@borgmanJeremy  No! This it's definitely bug! Please, reopen!

mmahmoudian:
@avevlad either we are talking about two different thing (which then you have to explain why this is a bug in your opinion), or this is not Flameshot bug. If you want the animation to go away, read:
https://apple.stackexchange.com/questions/128717/how-to-disable-full-screen-animation-on-os-x-10-9

l4kr:
Can't we do something like windowed fullscreen?

PR #4622, opened 2026-04-06
fix(macos): make fullscreen capture overlay configurable

Mitnitsky:
## Problem

On macOS, `showFullScreen()` triggers a native fullscreen space transition
animation when taking a screenshot, briefly switching the desktop.

## Fix

Replace `showFullScreen()` with `show()` using `FramelessWindowHint`,
`WindowStaysOnTopHint`, and `Qt::Tool` flags to avoid the animation.

Add a `useNativeFullscreen` config option (default: off) in
Settings > General so users who prefer the native fullscreen behavior
can re-enable it.

All changes are guarded with `#if defined(Q_OS_MACOS)`. Non-macOS
platforms are unchanged.

## Testing

- Default (off): capture overlay appears instantly, no desktop animation
- Toggle on: original fullscreen space transition behavior restored
- Setting persists across restarts

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -59,6 +59,9 @@ GeneralConf::GeneralConf(QWidget* parent)
 #if !defined(Q_OS_MACOS)
     initCaptureActiveMonitor();
 #endif
+#if defined(Q_OS_MACOS)
+    initUseNativeFullscreen();
+#endif
 #if defined(Q_OS_LINUX)
     initUseX11LegacyScreenshot();
 #endif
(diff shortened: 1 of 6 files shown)


mmahmoudian:
Fixes #2287

borgmanJeremy:
This one I want to think about more in addition to doing some testing on. I've grown accustomed to the behavior on MacOS and kind of like. It lets me easily swap workspaces between the fullscreen widget and the underlying active windows.

Mitnitsky:
Can create a toggle in the settings so you can opt-in/out for the behavior. Just the animation pf exiting workspace looks strange for me.

mmahmoudian:
I'm in favor of the toggle. When I used to use Flameshot on mac, this animation was catching me off guard, especially because this animation doesn't exists in mac built-in screenshot tool.

Mitnitsky:
Updated with a toggle to opt-in for fullscreen

borgmanJeremy (approved):

PR #4628, opened 2026-04-07
feature(macos): show dock icon when config window is open

Mitnitsky:
## Summary

On macOS with `LSUIElement=true`, Flameshot runs as an agent app with no dock icon. This makes it impossible to Cmd+Tab to open windows or see the app in the dock.

This PR adds a dock icon that appears while any user-facing window is visible and disappears when the last one is closed or hidden.

## Supported windows

- Settings (ConfigWindow)
- About (InfoWindow)
- Capture Launcher (CaptureLauncher)
- Upload History (UploadHistory)

## Implementation

- `Flameshot::showDockIcon(QWidget*)` — public one-liner for any window. Call after `show()`.
- Uses `QWindow::visibilityChanged` signal to track window visibility.
- Per-window dynamic property (`_visibleInDock`) prevents double-counting on minimize/maximize transitions.
- Reference counter toggles `NSApplicationActivationPolicy` between Regular (dock icon) and Accessory (no dock icon).
- Adding a new window requires just one line: `showDockIcon(myWidget)`.

## Testing

- Open/close each window type — dock icon appears/disappears correctly
- Multiple windows open — dock icon stays until last one closes
- CaptureLauncher hide (start capture) + close settings — dock icon disappears
- Minimize/restore — dock icon persists correctly

### DIFF
--- a/src/core/flameshot.cpp
+++ b/src/core/flameshot.cpp
@@ -7,6 +7,40 @@
 #include "qhotkey.h"
 #endif
 
+#if defined(Q_OS_MACOS)
+#include <QWindow>
+#include <objc/message.h>
+
+namespace {
+
+constexpr long NSApplicationActivationPolicyRegular = 0;
+constexpr long NSApplicationActivationPolicyAccessory = 1;
+
+void setActivationPolicy(long policy)
+{
+    auto sharedApp = reinterpret_cast<id (*)(id, SEL)>(objc_msgSend);
+    auto setPolicy = reinterpret_cast<void (*)(id, SEL, long)>(objc_msgSend);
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
Thanks for the very informative link explaining the problem and fix. However this only fixes the config window and not any other window that is spawned. Is there a way to abstract this so other widgets can add a one-liner and get this feature?

Mitnitsky:
> Thanks for the very informative link explaining the problem and fix. However this only fixes the config window and not any other window that is spawned. Is there a way to abstract this so other widgets can add a one-liner and get this feature?

I've added those changes :)

borgmanJeremy (approved):

PR #4629, opened 2026-04-07
fix(macos): fix clipboard copy failing from tray and GUI

### LINKED ISSUE #4630: [MacOS-ARM64]: Image is not saved in clipboard
### Flameshot Version

Flameshot v14.0.0 (b89ebc3)
Compiled with Qt 6.11.0
darwin: 25.3.0
macos: 26.3.1

### Installation Type

macOS Tahoe 26.3.1 (a) - ARM64

### Description

When taking a screenshot on Apple Silicon Macs, saving to the file works as expected. However, saving to the clipboard freezes the program for several seconds, after which the following line is printed in Terminal (if ran from there): `kdsingleapplication.localsocket: Message to primary timed out`.

The "output" is also inconsistent: the clipboard either remains empty, or is populated by an "empty" image of the same dimensions as the screenshot.

### Steps to reproduce

1. Install Flameshot by downloading `Flameshot-14.0.rc1-artifact-macos-arm64.dmg`.
2. Run the app.
3. Take a screenshot by pressing the button in the tray or via the shortcut.
4. Try to save the screenshot to clipboard by a shortcut or via clicking the button in the interactive menu.

https://github.com/user-attachments/assets/fcca8928-bc50-4080-8275-b581f9eee9b8

This is how it looks on user's end. Clipboard still contains the text, rather than the supposed screenshot.

### System Information

macOS Tahoe 26.3.1 (a) | ARM64
MacBook Pro (Apple M1 Pro)
Built-in Display (3024x1964)

Mitnitsky:
## Problem

Copying screenshots to clipboard on macOS was broken in two ways:

1. **3-second delay + clipboard empty**: `copyToClipboard()` created a
   `KDSingleApplication` instance and called `isPrimaryInstance()`, which
   always returned `false` because the daemon already holds the socket.
   This triggered IPC message delivery instead of using the in-process
   daemon directly. The IPC timed out after ~3 seconds and clipboard was
   never set.

2. **Lazy clipboard data lost on exit**: `setData("image/png")` puts a
   lazy reference on the macOS pasteboard — if the process exits before
   paste, data is lost ("Cannot keep promise"). The separate
   `saveJpegToClipboardMacOS()` workaround used osascript with a temp
   file and incorrectly declared JPEG data as PNG UTI (`«class PNGf»`).

## Fix

- Remove the `isPrimaryInstance()` check on macOS — use `instance()` to
  check if the daemon singleton exists in-process, matching non-macOS
  behavior. IPC fallback is preserved for true secondary processes (CLI).
- Add macOS branch in `saveToClipboardMime()` using `setImageData(QImage)`
  for native pasteboard persistence (`public.tiff`) plus `setData()` for
  exact format availability (PNG or JPEG).
- Delete `saveJpegToClipboardMacOS()` entirely — the fixed
  `saveToClipboardMime()` now handles both formats correctly.

All changes are guarded with `#if defined(Q_OS_MACOS)`. The Linux, Windows,
and Wayland clipboard paths are unchanged.

## Testing

- Tray menu Ctrl+C: clipboard set instantly (was 3s delay + empty)
- CLI `flameshot full --clipboard`: works via IPC fallback
- Both JPEG and PNG modes verified with `osascript -e 'clipboard info'

closes #4630

### DIFF
--- a/src/core/flameshotdaemon.cpp
+++ b/src/core/flameshotdaemon.cpp
@@ -141,12 +141,7 @@ void FlameshotDaemon::createPin(const QPixmap& capture, QRect geometry)
 
 void FlameshotDaemon::copyToClipboard(const QPixmap& capture)
 {
-#if defined(Q_OS_MACOS) && defined(USE_KDSINGLEAPPLICATION)
-    auto kdsa = KDSingleApplication(QStringLiteral("org.flameshot.Flameshot"));
-    if (kdsa.isPrimaryInstance() && instance()) {
-#else
     if (instance()) {
-#endif
         instance()->attachScreenshotToClipboard(capture);
         return;
     }
@@ -156,9 +151,7 @@ void FlameshotDaemon::copyToClipboard(const QPixmap& capture)
 
 #if defined(USE_KDSINGLEAPPLICATION) &&                                        \
   (defined(Q_OS_MACOS) || defined(Q_OS_WIN))
-#if defined(Q_OS_WIN)
     auto kdsa = KDSingleApplication(QStringLiteral("org.flameshot.Flameshot"));
-#endif
     stream << QStringLiteral("attachScreenshotToClipboard") << capture;
     kdsa.sendMessage(data);
 #else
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
This worked well and can be merged after addressing clang-format. The bug was introduced here. I am not sure why I didn't catch it in original testing: https://github.com/flameshot-org/flameshot/pull/4473

Mitnitsky:
Addressed the clang-format, should pass now

borgmanJeremy (approved):

PR #4640, opened 2026-04-14
fix(windows): restore spanning multi-monitor capture; fix geometry, compositor, and UI placement

Galygious:
## Summary

On Windows multi-monitor setups the capture overlay had regressed into a broken single-monitor-ish flow: the canvas could extend off-screen, multi-monitor layouts were misrendered (vertical offsets and gaps between monitors were silently flattened), the tool buttons landed off-screen on any monitor that wasn't the "first" one, and region-selection drag performance was pathological on large virtual desktops.

This PR restores the original spanning virtual-desktop capture model on Windows and fixes the underlying bugs cleanly. Tested end-to-end on a 6-monitor mixed-DPI Windows setup.

### Root causes fixed

1. **Compositor used a non-spatial layout.** `windowsScreenshot` placed each screen by "sum widths of predecessors" / "sum heights of predecessors", which silently discarded vertical offsets and gaps between monitors. Replaced with a single `primary->grabWindow(0, vdesk.x, vdesk.y, vdesk.w, vdesk.h)` call, letting the Windows DWM composite the virtual desktop itself. Per-monitor DPI, negative coordinates, gaps, and monitor ordering are all handled by the OS — no manual math.

2. **Overlay window was sized from the screenshot pixmap** rather than from the virtual desktop geometry, so any rounding slop or compositor drift leaked into window size (off-screen growth). Now derived from `QGuiApplication::primaryScreen()->virtualGeometry()`.

3. **`ButtonHandler` got a single "monitor at (0,0)" region** for the spanning overlay, so tool buttons computed relative to that rect landed off-screen for selections on any other monitor. Each physical monitor is now registered with its widget-local rect: `screen.geometry().translated(-vdesk.topLeft())`.

4. **`OverlayMessage`** (help and error text) was anchored to the whole-virtual-desktop area, stretching across all monitors. Anchored to the primary monitor's widget-local region.

5. **`paintEvent`** blitted the full screenshot pixmap unconditionally on every invalidation (`Q_UNUSED(paintEvent)` + `drawPixmap(0, 0, m_context.screenshot)`), producing pathological slowdowns during selection drag on large virtual desktops. Now clipped to `paintEvent->rect()` — only the background blit is narrowed; dim overlay and tool overlays are untouched.

`cropToMonitor` Windows branch (used by the explicit per-monitor CLI path e.g. `flameshot screen -n N`) was updated to match: derives the canvas DPR from the pixmap itself (`fullScreenshot.width() / vdesk.width()`) so it stays consistent with whatever Qt tagged the grab at, and rescales to `targetDpr` when they differ.

### Files changed

- `src/utils/screengrabber.cpp` — `windowsScreenshot` rewritten to delegate compositing to Windows; `cropToMonitor` Windows branch uses pixmap-derived DPR.
- `src/widgets/capture/capturewidget.cpp` — Windows branch uses `virtualGeometry()`; `ButtonHandler` regions registered per-monitor; `OverlayMessage` pinned to primary; `paintEvent` clips the background blit.

### Non-Windows platforms

Untouched. All platform-specific changes are inside `#if defined(Q_OS_WIN)` blocks; the Linux and macOS paths are exactly as before.

## Test plan

- [x] Build on Windows with MSVC 2022 + Qt 6.7.3
- [x] Real-hardware test on a 6-monitor mixed-DPI Windows setup: overlay spans all displays; selection reaches every edge on every monitor; tool buttons render on the correct monitor; drag is fluid; monitor layout (including vertical offsets and gaps) renders spatially correctly
- [ ] Reviewer sanity check on a simpler 2-monitor Windows setup
- [ ] Reviewer sanity check that Linux/macOS behavior is unchanged
- [ ] `flameshot screen -n N` still produces the correct per-monitor crop on Windows

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -248,7 +248,8 @@ QPixmap ScreenGrabber::grabEntireDesktop(bool& ok, int preSelectedMonitor)
     screenshot = windowsScreenshot(wid);
 #endif
 
-    // If monitor was pre-selected skip UI and crop directly
+    // Explicit per-monitor capture path (e.g. `flameshot screen -n N`,
+    // DBus capture-screen): crop the spanning composite down to that monitor.
     if (preSelectedMonitor >= 0) {
         const QList<QScreen*> screens = QGuiApplication::screens();
         if (preSelectedMonitor < screens.size()) {
@@ -257,7 +258,15 @@ QPixmap ScreenGrabber::grabEntireDesktop(bool& ok, int preSelectedMonitor)
         }
     }
 
+#if defined(Q_OS_WIN)
+    // Windows keeps its unified virtual-desktop capture model: the
+    // interactive overlay draws from the full spanning composite. The
+    // monitor-picker / single-monitor-crop path used on other platforms is
+    // intentionally bypassed here.
+    return screenshot;
+#else
     return selectMonitorAndCrop(screenshot, ok);
+#endif
 }
 
 QPixmap ScreenGrabber::grabFullDesktop(bool& ok)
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
Please read the beta telease notes. The new behavior is intentional.

PR #4651, opened 2026-04-20
Fix aspect ratio in selection widget when pressing CTRL

ElTh0r0:
Another approach to fix aspect ratio in selection widget when pressing CTRL (fix for #3436).

This PR gets the aspect ratio only once when the left mouse button is pressed. Before, the aspect ratio was permanently calculated while the mouse was moved, but QWidget:geometry only returns a QRect with integer precision, so we constantly lost aspect ratio precision while the mouse was moved.

There is another old PR #3439, which was never finalized. It may be closed if this one is merged.

### DIFF
--- a/src/widgets/capture/selectionwidget.cpp
+++ b/src/widgets/capture/selectionwidget.cpp
@@ -20,6 +20,7 @@ SelectionWidget::SelectionWidget(QColor c, QWidget* parent)
   , m_color(std::move(c))
   , m_activeSide(NO_SIDE)
   , m_ignoreMouse(false)
+  , m_aspectRatio(1)
 {
     // prevents this widget from consuming CaptureToolButton mouse events
     setAttribute(Qt::WA_TransparentForMouseEvents);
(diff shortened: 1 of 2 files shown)


mmahmoudian (approved):

mmahmoudian:
Nice fix. Also great to catch the division by zero. I initially didn't think about that.

As a general comment and not specific to this PR, I was puzzled why even changing the variable name and prepend `m_`. I think in general we need some more commenting in the codebase to explain some of these naming conventions.

ElTh0r0:
Thank you for the review!

> As a general comment and not specific to this PR, I was puzzled why even changing the variable name and prepend `m_`. I think in general we need some more commenting in the codebase to explain some of these naming conventions.

I moved the aspect ratio from a local function variable to a class member variable. The `m_` prefix is a common naming convention in C++ projects and most / all (?) Flameshot classes are using this convention, so I added the prefix too, even if this rule is not written anywhere.
I am open for a separate discussion about this topic!

andrewanswer:
[ElTh0r0](https://github.com/ElTh0r0) Thank you for this patch, waiting for releasing that!

ElTh0r0:
Hi @borgmanJeremy 
once you have time, please review if this fix can be merged 😊

borgmanJeremy:
Thanks for the ping!  Lost track of this one.

PR #4655, opened 2026-04-21
Add persistent history for dismissed pins

### LINKED ISSUE #360: Save history of screenshots taken
Adds capability to flameshot hold history of screenshots taken is awesome to retrieve old screenshots. Another software that do that with a good UX is the [Budgie Screenshot Widget](https://github.com/cybre/budgie-screenshot-applet) as you can see below:

image

manueltarouca:
Closes #360.

Dismissing a pin was a one-way door. This opens it.

Every close path (Esc, double-click, close button, programmatic close) now funnels through a new `dismissed()` signal from `PinWidget::closeEvent`, writes the full-resolution pixmap plus metadata (geometry, screen name, zoom, opacity, created and dismissed timestamps) to a JSON-indexed store, and makes the entry recoverable from a new Pin History dialog (virtualized `QListView` with lazy thumbnails), a Recent pins tray submenu (top 5, rebuilt on `aboutToShow`), and an explicit `restore()` that routes through `FlameshotDaemon::attachPin` — now public and taking optional zoom and opacity so restored pins reproduce the prior visual state rather than snapping to defaults.

Retention runs on three triggers (startup, every dismissal, dialog open) against both max-entries and max-age, against recorded event timestamps rather than wall-clock deltas, so the retention math tolerates system clock changes.

## What lands

- `src/utils/pinhistory{entry,store,manager}.{h,cpp}` — POD, persistent store (`QSaveFile` atomic writes, rebuild-from-scan that quarantines unknown files under `recovered/` on manifest corruption), and the manager owned by `FlameshotDaemon`.
- `src/config/pinhistoryconf.{h,cpp}` — new `Pin History` configuration tab (enable toggle, max entries 1–1000, max age with preset intervals, storage path label, disk usage, reveal, clear).
- `src/widgets/pinhistory{dialog,model}.{h,cpp}` — the dialog with context menu (Restore / Copy / Save As / Reveal / Delete) and toolbar (Clear all, Sort). Uniform item sizes on the `QListView` give free virtualization.
- Modified: `PinWidget` (new signal + `closeEvent` override + constructor gains optional `zoom`/`opacity` with backward-compatible defaults), `FlameshotDaemon` (`attachPin` public + optional zoom/opacity + manager ownership), `TrayIcon` (`Pin history…` action + `Recent pins` submenu populated on `aboutToShow`), `ConfigHandler` (three new flat keys in `[General]`).

## Verified locally (macOS 15, native dev build)

- [x] AC-1 pin → dismiss → shows in history and Recent pins
- [x] AC-2 persists across app restart
- [x] AC-4 double-click restores at stored geometry
- [x] AC-5 dismissed restore creates a new entry, original untouched
- [x] AC-6 copy-to-clipboard is pixel-exact
- [x] AC-8 / AC-9 delete one / clear all (with confirmation)
- [x] AC-12 disable / re-enable preserves existing entries
- [x] AC-19 disabled feature ≡ current master behavior

## Not yet verified (flagged per spec Appendix A.5)

- [ ] AC-17 Linux X11, Linux Wayland, Windows (author is single-platform)
- [ ] AC-10 max-entries enforcement (scriptable, not run)
- [ ] AC-11 max-age purge (needs a time-freeze harness)
- [ ] AC-14 / AC-15 stress and large-store performance

None are merge blockers. Happy to defer to reviewers with different platforms, or to a hardening pass once maintainers are comfortable with the design.

## Design decisions that deserve a second look

- **Storage path** mirrors `src/utils/history.cpp` (`$XDG_CACHE_HOME/flameshot/pin-history/` on Linux and macOS, `%APPDATA%/flameshot/pin-history/` on Windows). Spec §DM-1 argues for `AppDataLocation` instead, because OS cache cleaners could wipe user-controlled retention. Happy to switch if `history.cpp`'s cache placement is a legacy choice rather than a convention worth extending.
- **Config namespace** — new keys (`pinHistoryEnabled`, `pinHistoryMaxEntries`, `pinHistoryMaxAgeDays`) sit flat in `[General]` next to `uploadHistoryMax` and `historyConfirmationToDelete`. Spec §FR-23 wanted a `[PinHistory]` section; flat won because no existing feature carves its own section.
- **Threading** — writes are synchronous. The project has no existing threading primitives, and a full-res PNG encode on SSD sits well under perceptual thresholds on the dismissal path. `QtConcurrent::run` is the obvious next step if multi-monitor grabs start to show hitches.

## Deferred (spec §12 extensions, additive follow-ups)

FR-14 date-range filter, quicklook-on-hover preview, bulk save-as-ZIP, and a configurable hotkey slot for the dialog. None required for MVP; all additive on top of this landing.

## Translations

All user-visible strings wrapped in `tr()` or `QT_TRANSLATE_NOOP`. `lupdate` not run as part of this PR — recommend folding this into the next Weblate sync.

### DIFF
--- a/src/config/CMakeLists.txt
+++ b/src/config/CMakeLists.txt
@@ -11,6 +11,8 @@ target_sources(
           extendedslider.cpp
           filenameeditor.cpp
           generalconf.cpp
+          pinhistoryconf.h
+          pinhistoryconf.cpp
           setshortcutwidget.cpp
           shortcutswidget.cpp
           strftimechooserwidget.cpp
--- a/src/config/configwindow.cpp
+++ b/src/config/configwindow.cpp
@@ -5,6 +5,7 @@
 #include "config/configresolver.h"
 #include "config/filenameeditor.h"
 #include "config/generalconf.h"
+#include "config/pinhistoryconf.h"
 #include "config/shortcutswidget.h"
 #include "config/visualseditor.h"
 #include "utils/colorutils.h"
@@ -81,6 +82,15 @@ ConfigWindow::ConfigWindow(QWidget* parent)
                         QIcon(modifier + "name_edition.svg"),
                         tr("Filename Editor"));
 
+    // pin history
+    m_pinHistoryConf = new PinHistoryConf();
+    m_pinHistoryConfTab = new QWidget();
+    auto* pinHistoryLayout = new QVBoxLayout(m_pinHistoryConfTab);
+    m_pinHistoryConfTab->setLayout(pinHistoryLayout);
+    pinHistoryLayout->addWidget(m_pinHistoryConf);
+    m_tabWidget->addTab(
+      m_pinHistoryConfTab, QIcon(modifier + "config.svg"), tr("Pin History"));
+
     // shortcuts
     m_shortcuts = new ShortcutsWidget();
     m_shortcutsTab = new QWidget();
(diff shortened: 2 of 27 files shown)


Gocql022:
not effective in fedora43@gnome49
录屏 2026-05-17 01-32-13

borgmanJeremy:
Please discuss large PR's before submitting so everyone can agree on design / scope details. . Given that this PR is very large and reported to not work I am going to close this.

PR #4661, opened 2026-04-24
Update KDSingleApplication to v1.2.1

ElTh0r0:
KDSingleApplication received a little bug fix release. I am not aware that we have been affected by the bug, but I think updating is still a good idea!

See https://github.com/KDAB/KDSingleApplication/releases/tag/v1.2.1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,7 +124,7 @@ if(USE_KDSINGLEAPPLICATION)
       FetchContent_Declare(
         kdsingleApplication
         GIT_REPOSITORY https://github.com/KDAB/KDSingleApplication.git
-        GIT_TAG v1.2.0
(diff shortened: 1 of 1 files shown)


mmahmoudian (approved):

mmahmoudian:
Should we merge this before v14 release or after? The version change is a minor one and logically there should not be any breaking changes. I also agree with @ElTh0r0 that updating is a good idea in general.

mmahmoudian (approved):

ElTh0r0:
I think there is no urgency to merge this before v14 release. I have no 100% overview of the open issues, but I think we didn't receive any crash reports which may be related to this bug fix.
The risk of merging should be low (on my system I didn't face an issue), but who knows 😉 I'd vote for merging after the v14 release.

PR #4662, opened 2026-04-24
feat: configurable arrow key step size for selection widget

### LINKED ISSUE #3570: Allow taking interactive screenshots using only keyboard (`flameshot gui`)
### Feature Description

I'm a keyboard enjoyer, I don't like using mouse/touchpad at all, if I can do same this using only keyboard.

AnveshJarabani:
## Summary

- Adds two new config keys (`arrowMoveStep` and `arrowResizeStep`) that control how many pixels each arrow key press moves/resizes the selection
- Both exposed in the GUI settings (General tab) as spinboxes with range 1-200
- Default value of `1` preserves existing behavior — fully backwards compatible

## Motivation

The selection widget currently uses hardcoded 1px steps for arrow key movement and resizing. On modern high-resolution displays, this makes keyboard-driven region selection impractical — you'd need hundreds of key presses to move a selection across the screen.

This addresses the keyboard-only workflow requested in #3570. While it doesn't add full keyboard-only region creation, it makes the arrow key adjustment significantly more usable by letting users configure their preferred step size.

## Changes

| File | Change |
|------|--------|
| `src/utils/confighandler.cpp` | Register `arrowMoveStep` and `arrowResizeStep` as `BoundedInt(1, 200, 1)` |
| `src/utils/confighandler.h` | Add getter/setter macros |
| `src/widgets/capture/selectionwidget.h` | Add `m_moveStep` and `m_resizeStep` members |
| `src/widgets/capture/selectionwidget.cpp` | Load step sizes from config, use in all 12 move/resize/symResize methods |
| `src/config/generalconf.h` | Declare spinbox widgets and slots |
| `src/config/generalconf.cpp` | Add spinbox UI in General tab with tooltips |
| `flameshot.example.ini` | Document new keys |

## Usage

In `flameshot.ini`:
```ini
[General]
arrowMoveStep=10
arrowResizeStep=5
```

Or via the GUI: Settings → General → "Arrow move step (px)" / "Arrow resize step (px)"

## Test plan

- [ ] Default values (1) preserve existing 1px behavior
- [ ] Setting `arrowMoveStep=10` makes arrow keys move selection by 10px
- [ ] Setting `arrowResizeStep=20` makes Shift+arrow resize by 20px
- [ ] GUI spinboxes read and write values correctly
- [ ] Values persist across restarts
- [ ] Boundary clamping works (selection doesn't go off-screen)
- [ ] Minimum selection size (1x1) is still enforced

Closes #3570

### DIFF
--- a/flameshot.example.ini
+++ b/flameshot.example.ini
@@ -122,6 +122,12 @@
 ;; Set JPEG Quality (int in range 0-100)
 ;jpegQuality=75
 ;
+;; Arrow key step size for moving the selection (int in range 1-200)
+;arrowMoveStep=1
+;
+;; Arrow key step size for resizing the selection (int in range 1-200)
+;arrowResizeStep=1
+;
 ;; Shortcut Settings for all tools
 ;[Shortcuts]
 ;TYPE_ARROW=A
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -53,6 +53,8 @@ GeneralConf::GeneralConf(QWidget* parent)
     initCopyPathAfterSave();
     initAntialiasingPinZoom();
     initUndoLimit();
+    initArrowMoveStep();
+    initArrowResizeStep();
     initInsecurePixelate();
 #if !defined(Q_OS_MACOS)
     initCaptureActiveMonitor();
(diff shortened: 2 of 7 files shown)


mmahmoudian:
@borgmanJeremy I think this PR is conceptually in-line with #4796.

Issue #4463, opened 2026-01-02
DOC: Screenshot capture failure with Flameshot on Ubuntu 24.04 LTS (GNOME/Wayland)


JustinFntn:
## Environment

- **OS:** Ubuntu 24.04.3 LTS (Noble Numbat)  
- **Desktop:** GNOME Shell 46.0  
- **Display server:** Wayland  
- **Installation method:** `sudo apt install flameshot`  
- **Flameshot version:** `v12.1.0` (`Debian 12.1.0-2build2`)  

## Problem

On this setup, trying to take a screenshot with Flameshot results in the following error:

```text
flameshot: error: Unable to capture screen
```

This happens when launching from the tray icon or desktop launcher.  
After debugging, the issue appears to be related to screenshot permissions handled by `xdg-desktop-portal` on GNOME/Wayland.

Flameshot itself works correctly once the screenshot permissions are properly granted to the application.

## Solution: Grant screenshot permissions via D-Bus

The following sequence fixes the problem permanently on Ubuntu 24.04 + GNOME 46 + Wayland.

> ⚠️ Run these commands as your regular user (no `sudo`), in a terminal inside your graphical session.

### 1. Reset existing screenshot permissions

```bash
dbus-send --session --print-reply=literal \
  --dest=org.freedesktop.impl.portal.PermissionStore \
  /org/freedesktop/impl/portal/PermissionStore \
  org.freedesktop.impl.portal.PermissionStore.DeletePermission \
  string:'screenshot' string:'screenshot' string:''
```

This clears any previous (possibly broken or denied) screenshot permission entries.

### 2. Explicitly grant screenshot permission to Flameshot

```bash
dbus-send --session --print-reply=literal \
  --dest=org.freedesktop.impl.portal.PermissionStore \
  /org/freedesktop/impl/portal/PermissionStore \
  org.freedesktop.impl.portal.PermissionStore.SetPermission \
  string:'screenshot' boolean:true string:'screenshot' \
  string:'org.flameshot.Flameshot' array:string:'yes'
```

- `org.flameshot.Flameshot` is the application ID from the desktop file.
- The `array:string:'yes'` value grants the screenshot permission.

### 3. Verify that the permission is stored

```bash
dbus-send --session --print-reply=literal \
  --dest=org.freedesktop.impl.portal.PermissionStore \
  /org/freedesktop/impl/portal/PermissionStore \
  org.freedesktop.impl.portal.PermissionStore.Lookup \
[...]

Issue #3700, opened 2024-08-27
Flameshot not working on ubuntu 24.04 LTS


aman2talk:
### Version 
Flameshot v12.1.0 (Debian 12.1.0-2build2)
Compiled with Qt 5.15.12

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...), Snap from Snapcraft

### Operating System type and version

ubuntu: 24.04 LTS

### Description

The app doesn't get installed from snap store. if you try to install using the apt, It gets installed but when you start the app and try to capture the screenshot it says "Unable to capture the screen". 

### Screenshots or screen recordings

![image](https://github.com/user-attachments/assets/f778c139-b53b-451d-a9ea-4eae598542ab)

### System Information

linux: 6.8.0-41-generic

Kev-Prime:
+1 this having the same issues on arch most recent updoot nuked flameshot.

aman2talk:
I installed the fresh OS ubuntu 24.04 LTS with all the available updates. But it does not launch. But I am able to trigger the app via command line 'flameshot gui'. I even tried it compiling from source but same issue.

aman2talk:
Hi Everyone, I found the solution for the above issue. Actually the problem is, By default **Ubuntu 24.04** is using the **Wayland** as **GDM (GNOME Display Manager)** and **flameshot and upwork time tracker** require **Xorg** to function properly. You have to use the Xorg instead. To Switch from Wayland to Xorg follow the steps below.

### 1. Edit GDM(GNOME Display Manager) Configuration
a) Open the GDM configuration file using a text editor with root privileges
`sudo nano /etc/gdm3/custom.conf`

b) Uncomment or add the following line under the **[daemon]** section
`WaylandEnable=false`

### 2. Restart GDM:
Restart the GDM display manager to apply the changes:
`sudo systemctl restart gdm3`
Alternatively, you can reboot your system.

### 3. Verify the Change
To confirm you're now using Xorg, open a terminal and run
`echo $XDG_SESSION_TYPE`
This should output **x11** if you're successfully running on Xorg.

Now restart the flameshot app and try to take the screenshot. It will work flawlessly :)

Kev-Prime:
I just confirmed it works on both my X11 environment and wayland. 

Falmeshot supports both. 

If one isn't work certainly feel free to try the other of course.

jonathanhirschman:
[...]

PR #4664, opened 2026-04-26
screengrabber: pass non-empty parent_window to xdg-desktop-portal

artefaktor93:
## Bug

On GNOME 46 Wayland, daemon-mode capture (tray icon click, custom keybinding, or `gdbus call --session --dest org.flameshot.Flameshot --object-path /org/flameshot/Flameshot --method org.flameshot.Flameshot.captureScreen ...`) fails. `journalctl --user -u xdg-desktop-portal-gnome` shows:

```
xdg-desktop-portal-gnome[NNNN]: Failed to associate portal window with parent window ''
```

and Flameshot logs `Не удалось захватить экран` / `Unable to capture the screen`. Only `flameshot gui` invoked from a child of an already-visible window works, because xdg-desktop-portal-gnome on >= 46 strictly rejects requests where the `parent_window` argument of `org.freedesktop.portal.Screenshot.Screenshot()` is an empty string.

`ScreenGrabber::freeDesktopPortal()` in `src/utils/screengrabber.cpp` was hard-coding `""` as that parameter, which is the root cause.

## Fix

Allocate an offscreen `QWidget` with `Qt::WA_DontShowOnScreen` so a `winId` is generated without flashing a window. On X11 sessions pass `x11:0x<hex_winId>` (a real parent handle following the xdg-desktop-portal spec). On Wayland sessions, where we can't easily produce a real `wl_surface` exported handle from this code path, pass `wayland:` — non-empty, prefixed with the documented platform tag, and accepted by xdg-desktop-portal-gnome / xdg-desktop-portal-kde as a no-real-parent fallback. The portal spec explicitly allows the prefix-only form for callers that have no usable parent.

The change is contained to a single function, adds no dependencies, and preserves existing behavior on portals that previously accepted empty strings (they accept the prefixed forms too).

## Testing

- Ubuntu 24.04 base / Zorin OS 17 Pro, GNOME 46 Wayland
- NVIDIA proprietary 590.48.01
- Qt 6.4.2
- xdg-desktop-portal 1.18, xdg-desktop-portal-gnome 46.2
- Built from this branch with `cmake -B build && cmake --build build -j && sudo cmake --install build`
- Verified that `gdbus call ... captureScreen` triggers the selection overlay and writes `Снимок скопирован в буфер обмена` (screenshot copied to clipboard) to the journal
- Confirmed stable across two full logout/login cycles
- Tray-icon "Take Screenshot" entry now also works in the same session

## Refs

- Refs: #4663 (direct duplicate — same root cause, same error string)
- Refs: #4463, #3700, #3365 (GNOME Wayland capture failure reports likely caused by the same empty `parent_window`)

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -126,8 +126,8 @@ void ScreenGrabber::freeDesktopPortal(bool& ok, QPixmap& res)
     if (QGuiApplication::platformName() == QLatin1String("wayland")) {
         parentWindow = QStringLiteral("wayland:");
     } else {
-        parentWindow = QStringLiteral("x11:0x%1")
-                         .arg(parentDummy.winId(), 0, 16);
+        parentWindow =
+          QStringLiteral("x11:0x%1").arg(parentDummy.winId(), 0, 16);
     }
 
     screenshotInterface.call(
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -111,9 +111,28 @@ void ScreenGrabber::freeDesktopPortal(bool& ok, QPixmap& res)
     });
     timeout.start();
 
+    // Build a non-empty parent_window handle. xdg-desktop-portal-gnome
+    // (>= 46) rejects an empty string with "Failed to associate portal
+    // window with parent window ''" and the Screenshot request fails.
+    // On X11 we pass a real x11:<hex> handle from an offscreen QWidget.
+    // On Wayland we fall back to an empty string inside a wayland: prefix
+    // (xdg-desktop-portal-gnome treats unknown handles as no-parent and
+    // proceeds, instead of rejecting outright).
+    QString parentWindow;
+    QWidget parentDummy;
+    parentDummy.setAttribute(Qt::WA_DontShowOnScreen, true);
+    parentDummy.resize(1, 1);
+    parentDummy.show();
+    if (QGuiApplication::platformName() == QLatin1String("wayland")) {
+        parentWindow = QStringLiteral("wayland:");
+    } else {
+        parentWindow = QStringLiteral("x11:0x%1")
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Was the bug replicated on master before applying this patch?

taisph:
FWIW I've tested the build artifact in this PR [^1] and from the main branch [^2]. Both seems to work from both command line (flameshot gui) and print screen shortcut. The v14.0.rc1 release fails and earlier releases fail with "Unable to capture screen".

Tested on Ubuntu 24.04.4 under Wayland.

[^1]: flameshot-v14.0.rc1+git15.7f35b4e0-artifact-ubuntu-24.04-amd64/flameshot-14.0.0-1.ubuntu-24.04.amd64.deb
[^2]: flameshot-v14.0.rc1+git16.1534a895-artifact-ubuntu-24.04-amd64/flameshot-14.0.0-1.ubuntu-24.04.amd64.deb

borgmanJeremy:
@taisph  thank you, I will fix the formatting issue and mege.

AndyM48:
I think this bug has stopped flameshot from displaying on my system as well
`Linux slimbook 7.0.12-artix1-1 #1 SMP PREEMPT_DYNAMIC Wed, 10 Jun 2026 16:57:23 +0000 x86_64 GNU/Linux`
There may need to be a new release(tag) for the repos to be updated with the patches?

mmahmoudian:
@AndyM48 
> There may need to be a new release(tag) for the repos to be updated with the patches?

The v14.0.0 tag is already there, but we have not yet made a release page for it or updated it on Flathub.

AndyM48:
I had 14.0.0 installed and it did not work. I had to downgrade to the previous version for it to work again. 

I thought it might be this bug.

mmahmoudian:
@AndyM48 can you open a bug report for v14.0.0?

PR #4669, opened 2026-05-03
Avoid GNOME Wayland screenshot flash and monitor picker with optional Shell bridge

kanak-buet19:
## Summary

  Adds an optional GNOME Shell screenshot bridge for GNOME Wayland.

  The bridge lets Flameshot capture through GNOME Shell with screenshot flash disabled, then falls back to the existing portal path when
  the bridge is unavailable.

  ## What changed

  - Add optional GNOME Shell extension under `contrib/gnome-shell-extension`
  - Add a Flameshot GNOME Wayland backend that calls `org.flameshot.ShellIntegration`
  - Preserve existing `xdg-desktop-portal` fallback behavior
  - Return pointer coordinates from the Shell bridge so Flameshot can auto-select the active monitor on multi-monitor setups
  - Add install and verification docs for the GNOME Shell bridge
  - Document the optional GNOME Wayland integration in `README.md`

  ## Testing

  Tested locally on GNOME Shell 50.1 Wayland:

  - Built successfully with CMake
  - Installed and enabled the GNOME Shell bridge
  - Verified bridge D-Bus method with `gdbus introspect`
  - Verified direct bridge screenshot call returns success
  - Verified `flameshot gui` opens without GNOME portal flash
  - Verified monitor picker is skipped by using the pointer monitor
  - Verified no unsafe-mode notification after switching to Shell screenshot sender allowlist

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,14 @@ Now every time you press <kbd>Prt Sc</kbd> it will start Flameshot GUI instead o
 
 - Experimental Gnome Wayland and Plasma Wayland support.
 
+- GNOME Wayland normally captures through `xdg-desktop-portal`, which may show
+  a GNOME Shell flash. Flameshot includes an optional GNOME Shell bridge under
+  [`contrib/gnome-shell-extension`](contrib/gnome-shell-extension/README.md).
+  When enabled on supported GNOME Shell versions, Flameshot uses that bridge to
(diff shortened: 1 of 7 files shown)


borgmanJeremy:
Hello,

Thanks for this, but I do not want to rely on an extension for core functionality.

kanak-buet19:
No worries.

Issue #4679, opened 2026-05-09
Pin window gets minimized when clicking elsewhere


Nanmonk:
### Flameshot Version

13.3.0

### Installation Type

Package manager (pacman)

### Operating System type and version

Arch Linux

### Description

After pinning an image to the desktop, clicking on any other window causes the pin window to automatically minimize to the taskbar.
The taskbar flameshot icon then shows multiple stacked windows.

This defeats the purpose of pinning — the pin window should remain visible on screen even when it loses focus, like a sticky note.

### Steps to reproduce

1. Run `flameshot gui`
2. Select a region
3. Click the pin tool to pin the image to the desktop
4. Click on any other window

### System Information

- OS: Arch Linux (kernel 7.0.3-arch1-2)
- Desktop Environment: KDE Plasma
- Window Manager: KWin
- Display protocol: Wayland

mmahmoudian:
Is it minimized or does it go below other windows?

Nanmonk:
> Is it minimized or does it go below other windows?

  It goes below other windows. The pin window is still on the desktop
  but gets covered by other windows when they gain focus.


  After further investigation on KDE Plasma + Wayland, I found two separate issues:

  1. Pin window goes below other windows

  Qt::WindowStaysOnTopHint is silently ignored on Wayland because the standard xdg-shell protocol has no "keep above" state. The compositor controls window stacking.

  2. Pin window always appears at the center of the screen

  On Wayland, applications cannot set the position of xdg-shell toplevel windows. move() calls are ignored by the compositor. This is by design in the Wayland protocol.

  ---
  Workarounds for KDE users

  Option A — KWin Window Rules (no scripting required):
  System Settings → Window Management → Window Rules → Add rule matching title flameshot-pin, force "Keep above other windows" and "Skip taskbar".
  This fixes issues 1 (and taskbar visibility), but not positioning.

  Option B — KWin Script (fixes all three issues):
  A small KWin script can listen for windowAdded, set keepAbove/skipTaskbar, and reposition the window by reading the target coordinates encoded in the window title (flameshot-pin:x:y). This requires flameshot to encode the target
   position in the window title.

  ---
  Proper fix

  The clean application-level solution would be to use the zwlr_layer_shell_v1 Wayland protocol for the pin window. Layer surfaces support explicit positioning and can be placed above other windows. KWin supports this protocol.
  However, this would require significant refactoring of PinWidget.

mmahmoudian:
[...]

PR #4682, opened 2026-05-11
scripts: add KWin script workaround for pin window on KDE Plasma Wayland

Nanmonk:
## Problem

  On KDE Plasma + Wayland, the pin window has two issues:

  1. It sinks below other windows when focus changes — `Qt::WindowStaysOnTopHint` is silently ignored on Wayland because the xdg-shell protocol has no "keep above" state.
  2. It appears in the taskbar and can be minimized from there.

  Related: #4679

  ## Solution

  A small optional KWin script that listens for `windowAdded` and sets `keepAbove`/`skipTaskbar` for any window titled `flameshot-pin`.

  ## Scope

  - Only affects KDE Plasma + Wayland users
  - All other platforms are completely unaffected
  - Two new files under `scripts/kwin/`, no changes to the main codebase

  ## Installation

  ```bash
  mkdir -p ~/.local/share/kwin/scripts/flameshot-pin-fix/contents/code
  cp scripts/kwin/flameshot-pin-fix/metadata.json \
     ~/.local/share/kwin/scripts/flameshot-pin-fix/
  cp scripts/kwin/flameshot-pin-fix/contents/code/main.js \
     ~/.local/share/kwin/scripts/flameshot-pin-fix/contents/code/
  ```

  Then enable it in **System Settings → Window Management → KWin Scripts**,
  and re-login or run `qdbus org.kde.KWin /KWin reconfigure`.

  To verify you are on Wayland: `echo $XDG_SESSION_TYPE` should output `wayland`.

  ## Notes

  The proper fix would be to use the `zwlr_layer_shell_v1` protocol in `PinWidget`,
  but that requires significant refactoring. This script serves as a lightweight
  workaround in the meantime.

  Tested on KDE Plasma 6, Wayland, Arch Linux.

### DIFF
--- a/scripts/kwin/flameshot-pin-fix/contents/code/main.js
+++ b/scripts/kwin/flameshot-pin-fix/contents/code/main.js
@@ -1,0 +1,7 @@
+workspace.windowAdded.connect(function(window) {
+    if (!window.caption.startsWith("flameshot-pin")) {
+        return;
+    }
+    window.keepAbove = true;
+    window.skipTaskbar = true;
+});
--- a/scripts/kwin/flameshot-pin-fix/metadata.json
+++ b/scripts/kwin/flameshot-pin-fix/metadata.json
@@ -1,0 +1,9 @@
+{
+    "KPlugin": {
+        "Description": "Keep flameshot pin windows always on top and off the taskbar",
+        "Id": "flameshot-pin-fix",
+        "License": "GPL-3.0",
+        "Name": "Flameshot Pin Fix",
+        "Version": "1.0"
+    }
+}


borgmanJeremy:
I would prefer to not distribute this with Flameshot but curious for @mmahmoudian  thoughts.

mmahmoudian:
@borgmanJeremy 
I agree that this is not useful for everyone. Perhaps we can do one of the following
1. Put this in a misc folder in the main repo, that is excluded in all CIs. Other similar files and patches can go there
2. Create a dedicated github repo and have one folder per window manager with good documentation.
3. Considering that these are two short scripts, it can be put in a webpage on the website.

PR #4687, opened 2026-05-15
Allow filename templates to create subdirectories

### LINKED ISSUE #510: Allow sub directory formatting based on existing file name formatting
The current behavior when i try to set the file save destination including sub directories based on current time gives a weird math related fraction formatting:

Assuming it's now April 23th, 2019, 20:22 - I would like to save my screenshot in the following path:
`~/Pictures/Screenshots/19/04/23/20:22:00.png`
The equivalent formatting is `%y/%m/%d/%T`

Instead i get the fractional format when using slash signs between two date parts as you can see in the following screenshot:
My attempt

It would be great if slashes (/) could instead mark sub directories.

popsiclelmlm:
## Summary
- preserve literal `/` characters in filename templates as path separators while still sanitizing slashes produced by strftime specifiers
- create parent directories for generated filename-template paths when the configured destination is an existing directory

Fixes #510.

## Testing
- `git diff --check`
- `cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug` (fails locally: Qt6 development files are not installed, so `Qt6Config.cmake` is unavailable)

### DIFF
--- a/src/utils/filenamehandler.cpp
+++ b/src/utils/filenamehandler.cpp
@@ -7,6 +7,7 @@
 #include "utils/strfparse.h"
 
 #include <QDir>
+#include <QFileInfo>
 #include <exception>
 #include <locale>
 
@@ -41,12 +42,15 @@ QString FileNameHandler::parseFilename(const QString& name)
         res.chop(1);
     }
 
+    const QString pathSeparatorToken = QStringLiteral("\x1f");
+    res.replace(QLatin1String("/"), pathSeparatorToken);
     res =
-      QString::fromStdString(strfparse::format_time_string(name.toStdString()));
+      QString::fromStdString(strfparse::format_time_string(res.toStdString()));
 
     // add the parsed pattern in a correct format for the filesystem
     res = res.replace(QLatin1String("/"), QStringLiteral("⁄"))
-            .replace(QLatin1String(":"), QLatin1String("-"));
+            .replace(QLatin1String(":"), QLatin1String("-"))
+            .replace(pathSeparatorToken, QLatin1String("/"));
     return res;
 }
 
@@ -79,6 +83,7 @@ QString FileNameHandler::properScreenshotPath(QString path,
     if (info.isDir()) {
         // path is a directory => generate filename from configured pattern
         path = QDir(QDir(path).absolutePath() + "/" + parsedPattern()).path();
+        QDir().mkpath(QFileInfo(path).absolutePath());
     } else {
         // path points to a file => strip it of its suffix for now
         path = QDir(info.dir().absolutePath() + "/" + info.completeBaseName())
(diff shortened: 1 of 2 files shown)


ElTh0r0:
> fails locally: Qt6 development files are not installed, so Qt6Config.cmake is unavailable

Thank you for trying fix the mentioned issue, but after reading above statement it seems you didn't test your proposal yourself - and after testing it myself, I must say, that it doesn't properly work 😑

Currently the following is created with the file pattern '%m/%D/%y'
image

For proper fix it must be additionally considered, that the save path is automatically updated with the (new) subfolders as well - this must be prevented:


this causes multiple unwanted subfolders after taking more than one screenshot
image

ElTh0r0:
Personally, I like the idea of creating date-based subfolders, but the solution needs to be rethought.

In my opinion, it should not be allowed to use '/' and '\\' in file names at all and the workaround with this special character replacement should be completly removed.
https://github.com/flameshot-org/flameshot/blob/95032bd2b752c711860ed06b9f0225243a4b6cbd/src/utils/filenamehandler.cpp#L48

Just my two cents - waiting for further feedback.

popsiclelmlm:
Thanks for testing and for the detailed screenshots. I pushed a follow-up that keeps the configured save directory unchanged when the user accepts the generated filename, so date-based subfolders from the filename pattern should not be written back into the save path and nested on later screenshots.

I also left the configured path untouched when the save path is fixed. I could only run `git diff --check` locally; the existing build directory here does not contain a generated Makefile/Ninja file, so I could not run a local Qt build in this workspace.

ElTh0r0:
> so I could not run a local Qt build in this workspace.

Thank you for trying to take one part of my considerations into account (my first remark is still broken). However, I won't continue with the testing if you make code changes without testing them yourself first! Please run local tests yourself before pushing further commits.

popsiclelmlm:
Thanks for the follow-up. I set up the local Qt6 build environment and verified this before pushing another change.

The new commit fixes the remaining %m/%D/%y parsing case: when the configured pattern contains an explicit path separator, slashes produced by strftime expansions are also treated as path separators. Patterns without explicit separators keep the previous behavior of replacing strftime slashes with the fraction slash character.

Local checks run:
- cmake -S . -B build-local -GNinja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=/opt/homebrew/opt/qt
- cmake --build build-local --target flameshot -j2
- HOME=$(mktemp -d) build-local/src/flameshot.app/Contents/MacOS/flameshot config --filename %m/%D/%y -> Parsed pattern example: 05/05/15/26/26
- HOME=$(mktemp -d) build-local/src/flameshot.app/Contents/MacOS/flameshot config --filename %D_%H -> keeps the slash replacement behavior, e.g. 05⁄15⁄26_18
- git diff --check

Issue #256, opened 2018-06-21
Allow notification to be dragged and dropped


magnus-gross:
Screenshot apps like spectacle allow the notification to be dragged and dropped resulting in the captured image being dropped into another app. It would be cool if flameshot would support this too. :)

If you don't know what I mean, see https://dot.kde.org/2017/01/31/plasma-59-kicks-2017-style below "Be even more productive"

lupoDharkael:
Interesting concept, Flameshot used the general freedesktop's notification interface but I think this could be easily implemented.

PR #4691, opened 2026-05-18
Open saved screenshot file when notification is clicked

dshcherb:
On GNOME (Wayland), clicking Flameshot's screenshot notification did nothing. The notification was sent via D-Bus with empty actions and only a KDE-specific x-kde-urls hint, which GNOME Shell ignores.

The change introduces:

- Click-to-open logic:  added a `default` action to the D-Bus notification when a file path is attached (per the [freedesktop notification spec](https://specifications.freedesktop.org/notification/1.3/basic-design.html), this makes the notification body clickable). When the user clicks it, the notification daemon emits ActionInvoked, which a persistent actionHandler() singleton receives and calls QDesktopServices::openUrl() to open the file. The notification ID-to-path mapping is stored in a static map. We suppress the two non-actionable AbstractLogger::info() notifications (`GNOME Wayland detected...` and `Capture saved to clipboard`) by changing them to qDebug(). These sent competing notifications without a default action - if the user clicked one before the actionable `Capture saved as <path>` notification appeared, the click did nothing.

- Daemon delegation (Flatpak on GNOME Wayland): The Flatpak D-Bus proxy routes ActionInvoked only to the sandbox whose unique bus ID called Notify(). Each flatpak run invocation creates a separate proxy instance, so when the CLI sends the notification and exits, the daemon never receives the click signal.
This is fixed by having the CLI suppress its own `Capture saved as...` notification and sending the save path to the daemon via a new showSaveNotification D-Bus method. The daemon sends the notification itself, receives the click, and opens the file. The CLI exits immediately, allowing rapid successive screenshots.

Platform impact:
- macOS/Windows: no change (guarded by platform #if)
- Linux X11: minor, notification now includes default action
- Linux Wayland/KDE: improvement, click-to-open alongside x-kde-urls
- Linux Wayland/GNOME: fixed, click now opens the saved file

### DIFF
--- a/src/core/flameshotdbusadapter.cpp
+++ b/src/core/flameshotdbusadapter.cpp
@@ -4,6 +4,7 @@
 #include "flameshotdbusadapter.h"
 #include "core/flameshot.h"
 #include "core/flameshotdaemon.h"
+#include "utils/systemnotification.h"
 
 FlameshotDBusAdapter::FlameshotDBusAdapter(QObject* parent)
   : QDBusAbstractAdaptor(parent)
@@ -30,4 +31,19 @@ void FlameshotDBusAdapter::attachTextToClipboard(const QString& text,
 void FlameshotDBusAdapter::attachPin(const QByteArray& data)
 {
     FlameshotDaemon::instance()->attachPin(data);
+}
+
+void FlameshotDBusAdapter::registerNotificationPath(uint id,
+                                                      const QString& path)
+{
+    SystemNotification::registerNotificationPath(id, path);
+}
+
+void FlameshotDBusAdapter::showSaveNotification(const QString& path)
+{
+    SystemNotification sysNotif;
+    sysNotif.sendMessage(
+      QObject::tr("Capture saved as ") + path,
+      QObject::tr("Flameshot Info"),
+      path);
 }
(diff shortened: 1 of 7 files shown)


creativeheadz:
Hi @dshcherb, I built nearly the same thing independently last week (persistent action handler, notification id to path map, daemon delegation via a new D-Bus method) before finding this PR, so I can confirm the approach holds up beyond Flatpak/GNOME. On X11/Cinnamon the same lifetime problem appears for a simpler reason: `flameshot gui` exits as soon as the capture is saved, so the CLI process can never receive `ActionInvoked`. Delegating to the daemon fixes it there too.

One addition worth folding in on top of your `default` action: a named action that *reveals* the file via `org.freedesktop.FileManager1.ShowItems` instead of opening it. That is the desktop-agnostic equivalent of the KDE-only `x-kde-urls` hint already sitting in `systemnotification.cpp` (from #256), and it makes the saved capture draggable on GNOME/Cinnamon/XFCE as well. The two coexist cleanly: actions `["default", "", "flameshot-open-folder", "Open Folder"]`, body click opens the image, button reveals it. I have it working here: roughly 80 lines plus a small `FileManagerUtils::revealFile()` helper with fallbacks for Windows/macOS and for desktops with no `FileManager1` implementation.

This branch is 74 commits behind master and conflicts in one file. Happy to rebase it and add the reveal action as a second commit if that is useful, or to open it separately on top once this lands, whichever you prefer.

PR #4698, opened 2026-05-21
Increment/decrement circle counter on mouse wheel

milnadar:
1. Enhance the circle counter functionality by allowing to manually edit the counter pressing keyboard CTRL key and scrolling the wheel
2. Cap counter to the 1..999 rage (Negative numbers don't quite make sense to me. 3 digits is the biggest number that actually fit the counter) Subject for discussion?

### DIFF
--- a/src/tools/capturetool.h
+++ b/src/tools/capturetool.h
@@ -105,7 +105,12 @@ public:
         return {};
     };
     virtual QRect boundingRect() const = 0;
-    virtual bool handleMouseWheelEvent(int delta, bool adjustmentButtonPressed, CaptureContext& context) { return false; }
+    virtual bool handleMouseWheelEvent(int delta,
+                                       bool adjustmentButtonPressed,
+                                       CaptureContext& context)
+    {
+        return false;
+    }
 
     // The icon of the tool.
     // inEditor is true when the icon is requested inside the editor
--- a/src/tools/capturetool.h
+++ b/src/tools/capturetool.h
@@ -105,6 +105,7 @@ public:
         return {};
     };
     virtual QRect boundingRect() const = 0;
+    virtual bool handleMouseWheelEvent(int delta, bool adjustmentButtonPressed, CaptureContext& context) { return false; }
 
     // The icon of the tool.
     // inEditor is true when the icon is requested inside the editor
--- a/src/tools/circlecount/circlecounttool.cpp
+++ b/src/tools/circlecount/circlecounttool.cpp
@@ -180,17 +180,37 @@ void CircleCountTool::paintMousePreview(QPainter& painter,
 
     // Thickness for pen is *2 to range from radius to diameter to match the
     // ellipse draw function
-    auto orig_pen = painter.pen();
-    auto orig_opacity = painter.opacity();
+    auto const previewWidth = (size() + THICKNESS_OFFSET) * 2;
+    painter.save();
     painter.setOpacity(0.35);
     painter.setPen(QPen(context.color,
-                        (size() + THICKNESS_OFFSET) * 2,
+                        previewWidth,
                         Qt::SolidLine,
                         Qt::RoundCap));
     painter.drawLine(context.mousePos,
                      { context.mousePos.x() + 1, context.mousePos.y() + 1 });
-    painter.setOpacity(orig_opacity);
-    painter.setPen(orig_pen);
+
+
+    auto font = painter.font();
+    font.setPixelSize(previewWidth/2);
+    font.setBold(true);
+    painter.setFont(font);
+
+    painter.setPen(QPen("white"));
+    painter.drawText(QRect{ context.mousePos.x() - previewWidth/2,
+                            context.mousePos.y() - previewWidth/2,
+                            previewWidth,
+                            previewWidth  }, Qt::AlignCenter, QString::number(context.circleCount));
+
+    painter.restore();
+}
+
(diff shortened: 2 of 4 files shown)


mmahmoudian:
This would address #1214

mmahmoudian:
@milnadar can I ask why this was closed?

milnadar:
> @milnadar can I ask why this was closed?

Hi @mmahmoudian , how are you? Sorry, I didn't put the reason for closing. The #1214 that you'd mentioned had this proposed [PR](https://github.com/flameshot-org/flameshot/pull/2162) which is esentially the same thing that I did and it was rejected so I thought no reason to keep my PR hanging there. Hope that answers your question. Have a nice day

mmahmoudian:
I think what you are referring to is #2162. I'm curious to know @borgmanJeremy opinion on this.


@milnadar Regardless of the ultimate direction of this PR, I'm thankful for your contribution.

milnadar:
> I think what you are referring to is #2162. I'm curious to know @borgmanJeremy opinion on this.

Right, this is the one I was referring to. I would be glad to reopen the pr or create a new one with different logic because I personally DO need that feature. A side question please: the refactoring of the Circle counter implementation, who would be the person to discuss it with?

borgmanJeremy:
@mmahmoudian I think if we add this it should work as follows:

1. Circles start at one
2. Numbers increment automatically when a circle is placed
3. If users decrement the number, they can decrement to N-1...to (ie if the last number the placed was a 4, then can decrement from 3,2,1)
4. After the user decrements the number, it should start counting up again.

I didn't test the PR's yet so I don't know how they compare to this. Does that make sense in principal? I'm open to feedback on this.

mmahmoudian:
@borgmanJeremy 
My personal use-case for changing the number is when I'm making a tutorial and I have used numbers 1 through N in another screenshot, and I need to start from N+1. For me, decrement is not as useful as increment. For now, I have to click N times outside of the capture region. This is tedious, but usually works. The only time it doesn't work is when I'm capturing the entire monitor, which means there is no place outside of the capture region to place 1..N

Also sometimes I have to skip a number, because the user for example has to do something else in that step, like checking their email for 2FA or ... 

So for me, practically it is useful if I can start from an arbitrary positive integer, then click click... and then Ctrl+scroll up to increment to jump a number, and again click click... . In other words, I've never encountered the need to decrement.

One question we need to answer is: should be allow the same number to appear in more than one bubble? Preventing duplicates can help avoiding counting mistakes. Although because the user has to hold Ctrl and scroll, making mistake is not easy.

@milnadar
> the refactoring of the Circle counter implementation, who would be the person to discuss it with?

I believe @borgmanJeremy is the best person to discuss this with.

borgmanJeremy:
@mmahmoudian ah okay I can see how that would be useful, so I think maybe allowing scroll up and scroll down makes sense.

For me duplicates is okay because I might number things in color #2, and then restart with color #2.

mmahmoudian:
> For me duplicates is okay because I might number things in color #2, and then restart with color #2.

That makes sense. I didn't consider colors.

@milnadar what do you think? As the person making the PR, is there anything you would like to add/change? What are your use-cases and preferences?

milnadar:
> @borgmanJeremy My personal use-case for changing the number is when I'm making a tutorial and I have used numbers 1 through N in another screenshot, and I need to start from N+1. For me, decrement is not as useful as increment. For now, I have to click N times outside of the capture region. This is tedious, but usually works. The only time it doesn't work is when I'm capturing the entire monitor, which means there is no place outside of the capture region to place 1..N
> 
> Also sometimes I have to skip a number, because the user for example has to do something else in that step, like checking their email for 2FA or ...
> 
> So for me, practically it is useful if I can start from an arbitrary positive integer, then click click... and then Ctrl+scroll up to increment to jump a number, and again click click... . In other words, I've never encountered the need to decrement.
> 
> One question we need to answer is: should be allow the same number to appear in more than one bubble? Preventing duplicates can help avoiding counting mistakes. Although because the user has to hold Ctrl and scroll, making mistake is not easy.
> 
> @milnadar
> 
> > the refactoring of the Circle counter implementation, who would be the person to discuss it with?
> 
> I believe @borgmanJeremy is the best person to discuss this with.

Duplacates are the actual reason I wanted to add this feature. Sometimes I need to tie two different areas on the screenshot so I usually do that with numbers.

I don't have any additional requests for this one except I would like to open up a discussion about encapsulating the `CaptureContext::circleCount` variable inside of the Circle counter plugin. Right now the plugins internal functonality is spilled to the general code (Referring here to methods like: `CaptureWidget::restoreCircleCountState`, undo/redo, `CaptureWidget::StartDrawObjectTool`and the `CaptureContext` itself)

borgmanJeremy:
This generally looks good. I like that the preview now includes the number.  I also tested that scroll on a touch pad properly increments. One issue to resolve:

1. Using the scroll wheel hard caps the limit to 999, but if you place another bubble it lets you go past 1000. I kind of agree with capping at 999, but then we should also cap it to 999 when clicking.

milnadar:
> This generally looks good. I like that the preview now includes the number. I also tested that scroll on a touch pad properly increments. One issue to resolve:
> 
>     1. Using the scroll wheel hard caps the limit to 999, but if you place another bubble it lets you go past 1000. I kind of agree with capping at 999, but then we should also cap it to 999 when clicking.

Thank you for the review. I will address the capping issue

borgmanJeremy:
Sorry for the delay I was busy with the v14 build. I am ready to merge this once the CI / clang-format stuff passes.

PR #4701, opened 2026-05-26
update the releasing checklist to include Linux ARM CI

mmahmoudian:


borgmanJeremy:
This is probably not needed now that we merged the native arm CI runners.

PR #4726, opened 2026-06-11
add SECURITY.md

mmahmoudian:
We have recently received some vulnerability reports via email. So far only one of them was an actual vulnerability (based on my calculation, with CVSS score of 1.0, which is benign). @borgmanJeremy took the initiative and enabled the security section of this repo, and I put together a security policy after going through the following sources:
- https://nvd.nist.gov/vuln-metrics/cvss
- https://curl.se/dev/vuln-disclosure.html
- https://guix.gnu.org/en/security/
- https://github.com/greenshot/greenshot/blob/main/SECURITY.md

Turned out writing a security policy is not as easy as I thought, I wrote the text, and I also discussed the wordings and content with Google Gemini in case I have missed to mention something or if the tone has set the reporting bar too high. Luckily I had most of it covered and missed only one point that I then added. Ultimately, the language was corrected using Grammarly.

I don't expect this to be merged as-is, as perhaps it needs some more modifications. I'm open for suggestions from both Devs (especially @borgmanJeremy) and the community. So let's keep this open for about a week before merging.

I here tag few more people that I believe should check this document:
@panpuchkov  @lupoDharkael  @holazt  @hosiet  @Martin-Eckleben  @veracioux  @FelixJochems

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -29,7 +29,7 @@ To help us triage and resolve the issue efficiently, your report must include th
 To protect our volunteers' time, the following types of reports are considered out of scope unless they present a novel, unique threat vector:
 
 - Vulnerabilities that require prior root/administrator access or an already compromised operating system
-  - this is because if root is compromised, the user have bigger problem on their hand the bad actor can do much worse than what is remotely feasible with a screenshot tool
+  - this is because if root is compromised, the user have bigger problem on their hand and the bad actor can do much worse than what is remotely feasible with a screenshot tool
 - Theoretical issues or raw outputs from automated security scanners without a verified, functional Proof of Concept.
 
 ## Response Process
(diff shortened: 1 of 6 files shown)


SECURITY.md
```diff
@@ -0,0 +1,60 @@
...
+
+**Optional Information**:
+
+* Potential Solution: Suggestions, patches, or code fixes to address the issue are highly appreciated.
+
+# Out-of-Scope Vulnerabilities
+To protect our volunteers' time, the following types of reports are considered out of scope unless they present a novel, unique threat vector:
+
+- Vulnerabilities that require prior root/administrator access or an already compromised operating system
+  - this is because if root is compromised, the user have bigger problem on their hand the bad actor can do much worse than what is remotely feasible with a screenshot tool
```
borgmanJeremy: - this is because if root is compromised, the users have bigger problems on their hand and the bad actor can do much worse than what is remotely feasible with a screenshot tool

borgmanJeremy (approved):
minor wording correction otherwise this looks like a good starting point

Issue #4731, opened 2026-06-12
Allow manually specifying monitor to capture


pol-rivero:
### Feature Description

First of, thank you for the new multi-monitor support in v14! As someone who recently started using mixed scaling, this is a huge improvement <3

I'd like to request an option (CLI flag) in `flameshot gui` to specify which monitor number to use for screen capture, thus skipping the new monitor selection screen. This would be equivalent to the `-n` / `--number` flag that already exists in `flameshot screen`.

## Rationale:
Even if Wayland doesn't allow querying which monitor is currently selected (i.e. which monitor contains the cursor), some DEs such as Hyprland do allow it. Adding this option would allow users to create a wrapper that obtains the selected monitor and passes it to Flameshot, removing the need for the intermediate popup and improving the overall UX.

borgmanJeremy:
Hello,

I did update the screen command exactly for this use case in #4544

pol-rivero:
@borgmanJeremy Yes, I'm aware of that (please see my original comment). What I'm requesting is for the same flag to be added in `flameshot gui` as well.

Please let me know if you need any further information.

borgmanJeremy:
Why does it need to be added to gui?

pol-rivero:
Because in v14 it now shows a screen selector popup, instead of directly showing the screen capture gui

borgmanJeremy:
But if you call it with the screen command it bypasses that.

pol-rivero:
Apologies, maybe I'm misunderstanding how Flameshot works.
I'd like to show the UI where I can crop, draw lines, annotate, etc. as shown in the project's readme (below is the gif from the readme):
![gif](https://raw.githubusercontent.com/flameshot-org/flameshot/master/data/img/preview/animatedUsage.gif)

I thought the only way to show that UI was using `flameshow gui` (which now shows an additional "screen selection" popup). If I run `flameshow screen`, it just captures the entire screen instead of showing that UI.

borgmanJeremy:
Ohhh thank you for clarifying. I added a --interactive for the screen command to support that (I think that's what it is) I'm afk now

pol-rivero:
Thank you very much for clarifying! I don't know how I missed it... 🤦

For anyone searching, the flag is `--edit`:
```sh
flameshot screen --number 0 --edit
```

Sorry for the noise... 😅

borgmanJeremy:
If you'd like to contribute I bet the readme needs updated 😊

[...]

PR #4734, opened 2026-06-12
Document hyprland usage

pol-rivero:
As promised in #4731, I'm adding the documentation I wished I had when I was configuring Flameshot to work nicely with Hyprland.
I've hidden the code block inside a `<details>` element to avoid taking too much space in the readme.

I also considered documenting the `flameshot screen --number N --edit` more prominently in the "Example commands" section but ended up deciding against it because the "Capture Active Monitor" setting already exists and should cover most cases.
The Wayland DEs where it doesn't work probably require more tinkering in addition to `--edit`, that's why I added a full Hyprland config.

Please let me know if you disagree and would like me to add a `flameshot screen --number N --edit` line to the "Example commands" section too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@
 
 - [Features](#features)
 - [Usage](#usage)
+  - [Usage on Windows](#usage-on-windows)
+  - [Usage on Hyprland](#usage-on-hyprland)
   - [CLI configuration](#cli-configuration)
   - [Config file](#config-file)
 - [Keyboard Shortcuts](#keyboard-shortcuts)
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@
 - [Features](#features)
 - [Usage](#usage)
   - [Usage on Windows](#usage-on-windows)
-  - [Usage on Hyprland](#usage-on-hyprland)
+  - [Usage on Hyprland / Sway / wlroots](#usage-on-hyprland--sway--wlroots)
   - [CLI configuration](#cli-configuration)
   - [Config file](#config-file)
 - [Keyboard Shortcuts](#keyboard-shortcuts)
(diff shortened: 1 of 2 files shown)


mmahmoudian:
@pol-rivero 

Thank you for the PR.

> let me know if you disagree and would like me to add a `flameshot screen --number N --edit` line to the "Example commands" section too.

I don't believe it is necessary. Hyprland is geared towards tinkerers. This is a good start for users to get a minimum viable setup and then they can build around it of they feel like it. I also believe documentation should be brief to avoid boredom and to avoid getting overwhelming to read and follow. This PR is in a good level of details imho.

Would you be interested in adding the same content to our website?

https://github.com/flameshot-org/flameshot-org.github.io

You can create a page called "Setup in a Window Manager" or something like that. Perhaps later I can add Niri, Sway, River, and etc. to the list.

I can add it myself, but since you already wrote the text in markdown, it should be very low-effort work and still get the credit for your contribution to documentation.

mmahmoudian:
On second thought, perhaps README is not the best place for this. I suggest appending it to this markdown file:

https://github.com/flameshot-org/flameshot/blob/master/docs/Sway%20and%20wlroots%20support.md

pol-rivero:
@mmahmoudian Sure thing, I'll move it there.
Should I also add a link to the markdown file in the README? Right now it seems a bit hard to find (at least, I had no idea it existed).

mmahmoudian:
The documentation is for users, if you as a user believe adding the link would improve the documentation and help others users, please add it. 😊

Thanks for you kind contributions

pol-rivero:
@mmahmoudian Done, please check if everything looks okay :)

mmahmoudian (approved):

mmahmoudian:
Thank you. This is a valuable contribution.

Issue #4725, opened 2026-06-11
[Bug Flameshot RC-3]:  Wayland/Niri Stretched monitor preview and clipboard copy not working


SoyDiego:
### Flameshot Version
Flameshot: v14.0.0 (f3e81d26), compiled with Qt 6.11.1

### Installation Type

Compiled from source

### Operating System type and version

Manjaro Linux

### Description

Hi, I was trying to use RC3 to see if it's possible to take screenshots in Wayland.
You did a lot of great things and thanks for the effort and the hard work!
Right now I can select the monitor where I want to take the screenshot, but I'm not sure if it's Niri-specific — when selecting the monitor, everything appears stretched. You can see the following image as an example:

<img width="3373" height="1133" alt="Image" src="https://github.com/user-attachments/assets/955e0a6d-f8d4-4161-97f6-7db97bf7088b" />

After that, the arrows, circles, and everything else work fine, but when pressing Ctrl+C to copy, or clicking the copy button, it doesn't copy to the clipboard.

I'm not complaining, just reporting the bug in case you haven't seen it.
Thanks again!

### System Information

```
OS: Manjaro Linux
Kernel: 6.18.33-1-MANJARO
Compositor: Niri 26.04 (8ed0da4)
GPU: Intel Raptor Lake-S UHD Graphics + NVIDIA GeForce RTX 4060 Max-Q (hybrid)
Session: Wayland (WAYLAND_DISPLAY=wayland-1, XDG_CURRENT_DESKTOP=niri)
```

**XDG Desktop Portals**
```
xdg-desktop-portal 1.20.4-1
xdg-desktop-portal-wlr 0.8.2-1
xdg-desktop-portal-gnome 50.0-1
xdg-desktop-portal-gtk 1.15.3-1
xdg-desktop-portal-kde 6.6.5-1
```

**portals.conf**
```ini
[preferred]
default=wlr
org.freedesktop.impl.portal.Screenshot=wlr
org.freedesktop.impl.portal.ScreenCast=wlr
```

Creative0708:
I can reproduce the stretched screen on Arch Linux (Flameshot v14.0.0, Compiled with Qt 6.11.1, extra/flameshot) with niri 26.04 (8ed0da4). Copying works fine on my end, though -- I think the screen stretching and clipboard not working are separate bugs.

SoyDiego:
Fixed with #4735

PR #4735, opened 2026-06-13
Fix capture when logical min/max coordinates are not zero

rs-pro0:
The issue might occur when you have a monitor with negative coordinates, or when all monitors start at coordinates larger than 0x0 (for me, it was 3840x0).

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -489,8 +489,8 @@ QPixmap ScreenGrabber::cropToMonitor(const QPixmap& fullScreenshot,
     qreal targetDpr = targetScreen->devicePixelRatio();
 
     // Calculate total logical dimensions and minimum coordinates
-    int minX = 0, minY = 0;
-    int maxX = 0, maxY = 0;
+    int minX = INT_MAX, minY = INT_MAX;
+    int maxX = INT_MIN, maxY = INT_MIN;
 
     for (QScreen* screen : screens) {
         QRect geo = screen->geometry();


borgmanJeremy:
Can you explain the issue this is resolving?

rs-pro0:
> Can you explain the issue this is resolving?

I have my main display at 3840x0 and secondary one at 0x0, because my secondary display is disconnected, my minX should be 3840, but it's still 0. Here is how the screenshot gui looks on the current master branch.
image
And here on my branch.
image

mmahmoudian:
The following explanation is copied from #4743 which is solving the same issue the same way. I'm adding the PR's description to add to the context:

Posted by @Creative0708

> Unfixed code:
> 
> https://github.com/flameshot-org/flameshot/blob/c47870ac95cdce9e63c19198b1f1c0bf57504cf6/src/utils/screengrabber.cpp#L491-L501
> 
> Because the variables are initialised to 0, `minX/Y` can't go above 0 and `maxX/Y` can't go below 0, even if the monitor geometries do, resulting in improper scaling in compositors that don't place monitors at the origin (e.g. niri).
> 
> This PR changes the variables to `INT_MIN`/`INT_MAX` so the calculation works properly.
> 
> Fixes #4725

Creative0708:
Oops, somehow missed this PR when looking for duplicates. Nice work

SoyDiego:
This PR fix #4725 (my issue). I have tested in NIRI and works perfectly!!
Thanks

SoyDiego (approved):

borgmanJeremy (approved):

mmahmoudian:
@rs-pro0 and @Creative0708 thank you both for your kind contributions. It is valuable to have contributors like you in this project.

Issue #4391, opened 2025-11-13
Dual-Monitor Capture Breaks When Primary Monitor Is Right-Side Display (Black Screen + Misaligned Overlay) on Windows 11


Tupu4545:
### Flameshot Version

13.3.0 

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Windows 11 Pro 24H2 Build: 26100.4351

### Description

When using Flameshot on a dual-monitor setup on Windows 11, the capture overlay becomes completely broken if the primary monitor is the right-side display. Pressing Print Screen causes:

The primary monitor to turn black

The captured image to teleport to the left monitor (cropped/misaligned)

The selection overlay to appear on the wrong monitor

Flameshot to read the wrong framebuffer entirely

This happens only when the monitors are arranged [2][1], with monitor 1 set as the primary on the right.

If monitor 2 is set as primary, the issue disappears.
If both monitors are perfectly aligned horizontally with monitor 2 as primary — everything works normally.

If I move monitor 2 even one pixel down or change the primary → Flameshot breaks again.

This actually happened randomly and wasn't a issue before

### Steps to reproduce

Connect two 1080p monitors

Arrange them in Windows Display Settings as:

[ 2 ][ 1 ]

with 1 as primary display (on the right)

Move monitor 2 slightly down by a few pixels OR place them misaligned vertically

Launch Flameshot

Press Print Screen

Observe:

Primary monitor goes black

Flameshot overlay appears on the wrong monitor

Capture area is shifted

Cropped version of monitor 1 appears on monitor 2

What I already tried (no fix)

Fully removed NVIDIA drivers using DDU in Safe Mode

Reinstalled latest NVIDIA driver with clean install

Reinstalled Flameshot

Deleted config directories

Reset Windows display configuration

Disabled FancyZones (the issue started happening after using FancyZones or driver updated?) eitherway uninstalled both

Physically swapped HDMI ports

Tested with only one monitor connected (works)

Tested various primary monitor assignments

Tested latest GitHub Actions build

Rebooted dozens of times

None of these resolved the problem.

Expected behavior

Flameshot should correctly capture from the primary monitor regardless of physical arrangement or horizontal/vertical positioning. The overlay should not shift or black out the screen.

### Screenshots or screen recordings

full screenshot via win+ prt src: <img width="3840" height="1229" alt="Image" src="https://github.com/user-attachments/assets/2149bbb6-91d7-4315-919a-82a608415bbe" /> 

[...]

Issue #4385, opened 2025-11-12
Cannot capture on multiple monitors after upgrading to Windows 11


uelkue:
### Flameshot Version

Flameshot v13.3.0 (da913d7f)
Compiled with Qt 6.9.2
winnt: 10.0.22631
windows: 11

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Windows 11 23H2

### Description

Once Flameshot is launched to take a screenshot, it squeezes the content displayed on my primary monitor into a zone between the two monitors and the content from the secondary monitor vanishes, thus cannot be captured. Besides, the content becomes harder to see, due to getting squished into a smaller area and over the bezels of the monitors.

I have already tried fully uninstalling and then re-installing Flameshot, updated display drivers, disconnected and reconnected the display cables but none helped.

With Windows 10, everything was working as expected.

### Steps to reproduce

1. Launch Flameshot on a multi-monitor setup.
2. Notice the second monitor content disappearing and primary monitor content getting scaled down, leaving black borders.

### Screenshots or screen recordings

<img width="3640" height="1920" alt="Image" src="https://github.com/user-attachments/assets/ad7dc1b1-00b6-4dfd-b091-12b8467960d5" />

The screenshot above shows the combined display area and the awkward shape available to Flameshot while taking a screenshot (Notice the black borders).

Content from task bar blurred due to privacy reasons.

### System Information

<img width="661" height="272" alt="Image" src="https://github.com/user-attachments/assets/f0eeb890-e696-45b2-a10c-c050244083f1" />

Monitor 1 is disabled (laptop screen), only 2 and 3 are active. Monitor 3 is set as the primary display.
Monitor 2: 1080x1920 @ 60 Hz, 100% Scaling
Monitor 3: 2560x1440 @ 75 Hz, 125% Scaling

Windows 11, Version 23H2

madralf:
Same issue but also happened with earlier versions. Actually for me Flameshot was never able to deal with multiple monitors with different scaling settings.

gMicroge:
The same issue occurs when using v13.3.0. It was absent in v13.1.0

uelkue:
@gMicroge I had 13.1.0 as I upgraded to Win11. The same issue was present there as well, at least on my end. I then clean installed 13.3.0 to see if it solved the issue.
Although I don't want to exclude the possibility that the Windows upgrade might have messed something up in Flameshot.

christopher-buss:
[...]

Issue #4362, opened 2025-11-03
Black screen during capture and incorrect screen scaling


danieldemoraisgurgel:
### Flameshot Version

Version 13.3.0 (da913d7f)

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Windows 11 25H2

### Description

 - When taking a screenshot, the main laptop screen goes completely black and the secondary screen (external monitor) becomes misaligned.

Windows 11 25H2, Full HD (1980x1080p

![Image](https://github.com/user-attachments/assets/94bd7e52-89d0-443d-8cbd-4da03700f61a)

x) screens.

### Screenshots or screen recordings

![Image](https://github.com/user-attachments/assets/94bd7e52-89d0-443d-8cbd-4da03700f61a)

### System Information

![Image](https://github.com/user-attachments/assets/94bd7e52-89d0-443d-8cbd-4da03700f61a)

pedroalves5770:
I have the same issue! 
rolled back to 13.1.0 and it works

diggidong:
I have the same issue on Arch Linux 6.17.6 and flameshot v13.3.0.

alecclyde:
Same issues, Triple monitor setup (smaller one on top, two large on bottom). Windows 11. Screens seem to move up and left significantly, bottom 2 monitors are completely blank.

rsaleri:
Similar issue, triple monitor setup, this is what happens when I try to take a screenshot with 13.3.0 (sorry for the weird "censoring") on Pop_OS 22.04
![Image](https://github.com/user-attachments/assets/1c042d3e-1d0c-449e-87d4-65157200f446)

zloylinux:
It's strange that the developer isn't in a hurry to fix the issue in their app. I ran into the same problem.

eduardohf:
Hi guys, I am facing the same issue with Windows 11, laptop monitor and a second external monitor. the laptop screen goes completely black.

eduardohf:
I rolled back to 13.1.0 and it did not work for me, but when I rolled back to 13.0.1, it works...

l1mey112:
Same issue

intero-jbm:
Same issue and meanwhile unusable :( Roll back did also not work, tried new version again, but no success.

zloylinux:
> Same issue and meanwhile unusable :( Roll back did also not work, tried new version again, but no success.

[...]

Issue #4152, opened 2025-08-13
Problem with screenshot when having 2 monitors (Windows 11 24h2)


franciswrpereira:
### Flameshot Version

Flameshot Version
Flameshot v13.0.1 (48f8d1fd)
Compiled with Qt 6.8.1
winnt: 10.0.26100

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Windows 11 24H2

### Description

Problem with screen capture when having 2 monitors, the main one being on the right and the secondary one being on the left, causes the area to be captured to be shifted to the left and cuts off a good part of the full screen for screen capture.

### Steps to reproduce

1. Start capturing the screen under the conditions I mentioned in the description

### Screenshots or screen recordings

<img width="5759" height="2158" alt="Image" src="https://github.com/user-attachments/assets/0c6699e3-a7bd-4f45-acea-96e543395c4c" />

### System Information

1. Windows 11 24H2
2. The 2 monitors are exactly the same size and 4K resolution at 150% scale
3. Nvidia Driver: 580.97

borgmanJeremy:
We just merged #4127 to fix this but have not made a new release yet.

@mmahmoudian are you on board with another quick point release to get that fix out?

mmahmoudian:
@borgmanJeremy definitely. And also let's have the Flatpak out as well in Flathub as it is still 12.1.0

israelmaestro:
The same happens to me, but in the vertical orientation!

Using Windows 11 Home Edition, with two displays (laptop's and HDMI), where display 2 is configured to be above display 1, resolution 1920x1080 on both displays, when calling Flameshot for screen capture, the entire image on both displays is displaced upwards. It was not occurring in the previous version.

SoyDiego:
I think I have the same problem in Manjaro KDE. It was working perfectly and now doesn't work with 2 monitor (one vertical and the other in normal position) Here you have an example when I'm pressing print screen in my horizontal monitor
<img width="3441" height="1432" alt="Image" src="https://github.com/user-attachments/assets/26c13df6-4519-4807-a9e3-d9d2994c501f" />

Black screen on the bottom :( It's like is reducing the display or something like this.

dmolin:
I can confirm this happening also in my Linux Manjaro with AwesomeWS with multiple monitors (I've 3 monitors) after my recent update yesterday.

vmvartiainen:
[...]

PR #4736, opened 2026-06-13
Fix Windows multi-monitor composite screenshot layout

jacksonzacarias:
Place each grabWindow pixmap using the virtual desktop logical union origin scaled by per-screen DPR, and match cropToMonitor to the same coordinates. Staggered monitor layouts (e.g. different Y, negative X) no longer misalign the composite or crop.

See docs/Windows-multi-monitor-fix.md for root cause and manual test matrix.

### DIFF
--- a/docs/Windows-multi-monitor-fix.md
+++ b/docs/Windows-multi-monitor-fix.md
@@ -1,0 +1,76 @@
+# Windows: multi-monitor capture and overlay alignment
+
+This note summarizes **what was wrong** on Windows with heterogeneous multi-monitor layouts and **what to change** so capture, compositing, and cropping stay consistent with the virtual desktop.
+
+## Symptom
+
+On Windows 11 with multiple displays (e.g. negative `X`, different heights, non-zero secondary `Y`), users could see:
+
+- black regions in the composite screenshot;
+- selection / overlay offset from the real image;
+- cropped or "swapped" monitor content.
+
+The behaviour often changed when switching which display was primary.
+
+## Root cause
+
+`ScreenGrabber::windowsScreenshot()` placed each per-screen `grabWindow()` pixmap on a canvas using **only**:
+
+- summed widths of monitors fully to the **left** of the current screen, and
+- summed heights of monitors fully **above** the current screen.
+
+That approximates a **striped grid**, not the true **2D virtual desktop** layout. It breaks when monitors are staggered (e.g. left monitor has a different `Y` than the right one): the composite buffer no longer matches the union of `QScreen::geometry()` in pixel space.
+
+`cropToMonitor()` on Windows reused the same "sum left / sum above" logic, so the crop rectangle did not match the composite either.
+
+## Intended behaviour (reference)
+
+Treat the desktop as one rectangle: the union of all screen geometries in **logical** coordinates, with origin at the **minimum** `(x, y)`. Each monitor's pixmap is pasted at:
+
+```text
+physicalOffset = (screen.geometry().topLeft() - virtualTopLeft) * screen.devicePixelRatio()
+```
+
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
What is this trying to fix?

jacksonzacarias:
This PR aims to improve Windows multi-monitor capture when displays are arranged in layouts that are not simple aligned grids (negative X coordinates, different monitor heights, or staggered Y positions).

During investigation, I reproduced the issue on Windows 11 with a monitor positioned at `X=-1920` and `Y=51` relative to the primary display. In that setup, the composite screenshot, overlay, and crop calculations could become inconsistent with the actual virtual desktop geometry, resulting in black regions, shifted selections, or incorrect cropping.

The proposed change aligns both compositing and cropping with the union of all `QScreen::geometry()` rectangles, using a single virtual desktop coordinate space and DPR-aware offsets.

While investigating, I also noticed several reports that appear to show similar symptoms:

* #4152 (capture shifted/cropped with a monitor positioned to the left of the primary display)
* #4391 (black screen and misaligned overlay)
* #4362 (black primary display and secondary monitor misalignment)
* #4385 (duplicate report mentioning disappearing monitor content and black borders)

I'm not claiming all of these issues share the exact same root cause, but they seem consistent with geometry/compositing mismatches in asymmetric multi-monitor layouts.

Additional details, rationale, and test cases are documented in:

`docs/Windows-multi-monitor-fix.md`

borgmanJeremy:
can you post more details and pictures for the configuration that resulted in the problem and what it looked like?

borgmanJeremy:
every issue referenced is already closed....

jacksonzacarias:
Thanks for asking for more details.

We reproduced the issue on two different physical PCs running Windows 11 25H2 and verified the behaviour before and after the proposed patch.

### Test environments

| System | Edition | Build |
|----------|----------|----------|
| System 1 | Windows 11 Pro | 26200.8655 |
| System 2 | Windows 11 Home | 26200.8457 |

### Configuration that triggers the issue

The problem appears in non-trivial multi-monitor layouts, for example:

DISPLAY1
1920x1080
X=-1920
Y=51

DISPLAY2
1920x1200
X=0
Y=0
Primary=true
image


Important characteristics:

- secondary monitor positioned at a negative X coordinate;
- different monitor heights/resolutions;
- non-aligned Y position (staggered layout).

### OS version

(INSERT IMAGE 1 HERE - winver screenshot)

### Desktop before starting Flameshot

Both displays working normally before entering capture mode.

image

### Before the proposed patch

Flameshot 13.3 on Windows 11.

Observed behaviour:

- laptop display becomes almost completely black;
- overlay appears only on the external monitor;
- captured background does not match the actual desktop layout;
- monitor content may appear cropped or misplaced depending on monitor arrangement and primary display selection.

The photo was taken with a phone because the issue affects the capture overlay itself.

deskto02
desktop01-problema


### After the proposed patch

Same workflow using a patched build.

The important difference is that the composite image, monitor placement, overlay positioning and crop calculations now match the actual virtual desktop layout.

The darkened displays shown in monitor-selection mode are expected Flameshot behaviour; this PR does not modify that UI.

comnossasolucaov14

### Technical summary

During investigation, it appeared that the Windows compositing and cropping logic could become inconsistent with the actual virtual desktop geometry when monitor coordinates are asymmetric (negative X, different heights, staggered Y positions, and potentially mixed DPI setups).

The proposed patch aligns both windowsScreenshot() and cropToMonitor() with the logical union of all QScreen::geometry() rectangles, using a single virtual desktop coordinate space and per-screen DPR-aware placement.

I'm not claiming this is the same root cause as any previously closed issue. I only noticed similar symptoms while investigating and focused on the behaviour reproduced on the systems above.

I can also provide monitor geometry logs and additional test results if that would be useful.

borgmanJeremy:
Please test with the v14 beta not v13.x

PR #4738, opened 2026-06-14
Let the GUI colors be choosen via hex values

IndolikarQuantifier:
This will add functionality to set color values from hex input.

image

image

### DIFF
--- a/src/config/uicoloreditor.cpp
+++ b/src/config/uicoloreditor.cpp
@@ -79,6 +79,8 @@ void UIcolorEditor::changeInputColor(const QString& hexColor) {
         QColor color(hexColor);
         if(color.isValid()) {
             m_colorWheel->setColor(color);
+            updateLocalColor(color);
+            updateUIcolor();
         }
     }
 }
--- a/src/config/uicoloreditor.cpp
+++ b/src/config/uicoloreditor.cpp
@@ -12,6 +12,7 @@
 #include <QMap>
 #include <QSpacerItem>
 #include <QVBoxLayout>
+#include <QLineEdit>
 
 UIcolorEditor::UIcolorEditor(QWidget* parent)
   : QWidget(parent)
@@ -24,6 +25,7 @@ UIcolorEditor::UIcolorEditor(QWidget* parent)
     m_hLayout->addItem(new QSpacerItem(space, space, QSizePolicy::Expanding));
     m_vLayout->setAlignment(Qt::AlignVCenter);
 
+    initHexColorInput();
     initButtons();
     initColorWheel();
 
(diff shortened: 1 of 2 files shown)


mmahmoudian:
@IndolikarQuantifier 
Looks good to me, but please add a text above the textbox to clarify what the textfield is representing. Not all users are tech savvy  enough to know what a hexadecimal color code is. Also you are allowing alpha to be set. Please reflect that in the text too, so that users know they also have the option.

IndolikarQuantifier:
> @IndolikarQuantifier Looks good to me, but please add a text above the textbox to clarify what the textfield is representing. Not all users are tech savvy enough to know what a hexadecimal color code is. Also you are allowing alpha to be set. Please reflect that in the text too, so that users know they also have the option.

I have added label for hex text box. 
Also the hex values represents RGB and it doesn't contain alpha.

src/config/uicoloreditor.cpp
```diff
@@ -68,6 +70,17 @@ void UIcolorEditor::updateLocalColor(const QColor c)
         m_contrastColor = c;
     }
     m_lastButtonPressed->setColor(c);
+    m_hexColorEdit->setText(c.name());
+}
+
+void UIcolorEditor::changeInputColor(const QString& hexColor) {
+
+    if(hexColor.contains(QRegularExpression("^#(?:[0-9A-Fa-f]{6}|[0-9A-Fa-f]{8})$"))) {
```
ElTh0r0: > Also the hex values represents RGB and it doesn't contain alpha.

Your regular expression suggests the opposite! You are explicitly allowing a hex value with either 6 OR 8 characters which would be #RRGGBB or #AARRGGBB (A = alpha) 😉

IndolikarQuantifier: Oh that's right I have updated the regex so that it will accept only RGB values

ElTh0r0: I noticed an issue during my test: If I insert a (valid) hex value into the textbox, the color is not used/applied! For me it is still using the previous color. I think you have to connect to slot `UIcolorEditor::updateUIcolor` to apply the input.

IndolikarQuantifier: resolved this issue

ElTh0r0 (approved):

IndolikarQuantifier:
Hi @mmahmoudian
who's review is required to merge this PR.

mmahmoudian:
@IndolikarQuantifier 
> who's review is required to merge this PR.

Although @ElTh0r0 kindly approved the PR and I also reviewed it, I believe we still need to hear @borgmanJeremy opinion on the matter. Thanks for your patience.

borgmanJeremy (approved):
thanks!

borgmanJeremy:
Once the clang-format failures are addressed we can merge. Thanks for your patience.

IndolikarQuantifier:
> Once the clang-format failures are addressed we can merge. Thanks for your patience.

Thansk for review
I have made changes to pass clang-format check

IndolikarQuantifier:
can you please merge this PR @mmahmoudian @borgmanJeremy

borgmanJeremy:
Thank you!

PR #4760, opened 2026-06-19
improve config text

mmahmoudian:
The "Capture active monitor" is exclusive to X11 (as the tooltop also indicates), so here I add minimum possible number of words to clarify that.

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -924,7 +924,7 @@ void GeneralConf::setInsecurePixelate(bool checked)
 void GeneralConf::initCaptureActiveMonitor()
 {
     m_captureActiveMonitor = new QCheckBox(
-      tr("Capture active monitor (skip monitor selection)"), this);
+      tr("Capture active monitor in X11 (skip monitor selection)"), this);
     m_captureActiveMonitor->setToolTip(
       tr("Automatically capture the monitor where the cursor is located "
          "instead of showing the monitor selection dialog. "


src/config/generalconf.cpp
```diff
@@ -924,7 +924,7 @@ void GeneralConf::setInsecurePixelate(bool checked)
 void GeneralConf::initCaptureActiveMonitor()
 {
     m_captureActiveMonitor = new QCheckBox(
-      tr("Capture active monitor (skip monitor selection)"), this);
+      tr("Capture active monitor in X11 (skip monitor selection)"), this);
```
borgmanJeremy, line 927: or Windows (i think)

mmahmoudian: How about macOS? If it works on x11, Windows and macOS, I can write "excluding Wayland"

borgmanJeremy: i double checked, its not available on macos

mmahmoudian:
@ElTh0r0 I don't have access to a Window machine to test this. Does the "Capture active monitor" option in the config work on Windows? If it does, I can create a PR and add Windows to the text as well.

ElTh0r0:
> @ElTh0r0 I don't have access to a Window machine to test this. Does the "Capture active monitor" option in the config work on Windows? If it does, I can create a PR and add Windows to the text as well.

yes, "Capture active monitor" works on Windows, too!

PR #4777, opened 2026-06-25
ci: update linux distro build matrix

holazt:
- add debian 13 amd64/arm64/armhf, drop debian 12
- add fedora 43/44 amd64
- add ubuntu 26.04 amd64/arm64, drop ubuntu 22.04
- add opensuse leap 16.0 amd64

### DIFF
--- a/.github/workflows/Linux-pack.yml
+++ b/.github/workflows/Linux-pack.yml
@@ -190,12 +190,6 @@ jobs:
               symbol: 44,
               arch: x86_64
             }
-          - {
-              name: 'opensuse-leap-16.0',
-              os: opensuse-leap,
-              symbol: '16.0',
-              arch: x86_64
-            }
     steps:
       - name: Checkout Source code
         if: github.event_name == 'push'
--- a/.github/workflows/Linux-pack.yml
+++ b/.github/workflows/Linux-pack.yml
@@ -37,16 +37,9 @@ jobs:
       matrix:
         dist:
           - {
-              name: debian-12,
+              name: debian-13,
               os: debian,
-              symbol: bookworm,
-              arch: amd64,
-              runs-on: ubuntu-latest
-            }
-          - {
-              name: ubuntu-22.04,
-              os: ubuntu,
-              symbol: jammy,
+              symbol: trixie,
               arch: amd64,
               runs-on: ubuntu-latest
             }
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
Id like a one in, one out in place on the builds so we are not continuously adding more builds to the CI and release process. Could you please remove the older versions from the build pipeline?

holazt:
> Id like a one in, one out in place on the builds so we are not continuously adding more builds to the CI and release process. Could you please remove the older versions from the build pipeline?

ok, I have removed debian 12 and ubuntu 22.04.

jankatins:
Fedora 44 package does not work:

```text
Failed to resolve transaction: 
Problem: conflicting requests
  - nothing provides qt6-svg(x86-64) needed by flameshot-14.0.0-1.fc44.x86_64 from @commandline
 ```

**Comment by jankatins** [comment]
This needs to update the spec file to change the hardcoded dependency for fedora:

https://github.com/flameshot-org/flameshot/blob/6a635d2d5d88190dab8152bcebb4a85796a55ab3/packaging/rpm/flameshot.spec#L44-L48  

Given that the fedora version before did have a dependency for `Requires: qt6-qtsvg >= 6.2.4` and I couldn't find any package with the `qt6-svg` name in fedora in the last few fedora version, this looks like a copy and paste error in https://github.com/flameshot-org/flameshot/commit/46e4a7daf5c94d1512512d88705ff05cf79dea29#diff-00583ee3e4e32bbf90e6a95c528d2f3ecea49e07f995538e37bbb765e1985e54

-> if that dependency should be there, it should simply be named `qt6-qtsvg` and maybe get the version back

```diff
 %if 0%{?suse_version} 
- Requires: qt6-svg 
+ Requires: qt6-qtsvg >= 6.2.4
 %else 
 Requires: qt6-svg%{?_isa} 
 %endif 
```

jankatins:
I guess this might also be an inspiration: the [official fedora spec file](https://src.fedoraproject.org/rpms/flameshot/blob/rawhide/f/flameshot.spec) uses `Requires: qt6-qtsvg%{?_isa}`

holazt:
removed opensuse-leap-16.0 support because there is no official `qt6-svg` package available for openSUSE Leap 16.0, https://software.opensuse.org/package/qt6-svg

borgmanJeremy (approved):

borgmanJeremy:
Thank you!

PR #4778, opened 2026-06-25
Remove deprecated msg and added error message

borgmanJeremy:
I removed the "deprecated" note on the X11 screenshot method. It's clear that many users will need this method for a while. 

Added a clear error message for what to do when the portal times out.

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -970,7 +970,7 @@ void GeneralConf::useNativeFullscreenChanged(bool checked)
 void GeneralConf::initUseX11LegacyScreenshot()
 {
     m_useX11LegacyScreenshot =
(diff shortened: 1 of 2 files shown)


mmahmoudian:
Considering that many users Don't use CLI to trigger Flameshot, I'd suggest:
1. Adding an argument that is Enabled by default for auto-detecting Wayland/X11
2. If X11 was deceted, automatically enable the X11 legacy mode
3. Communicate this messages via notifications as well

What do you think? It is possible, and if it is, is it feasible?

borgmanJeremy:
@mmahmoudian The reason I dont automatically switch is because some X11 managers work fine with the portal (gnome and cinnamon for sure). The message should have come up via notification, it did when I tested it here.

I tested it by setting the timeout to 1mS temporarily.

Issue #4784, opened 2026-06-29
Package for Fedora 44


jankatins:
### Feature Description

Unfortunately, the Fedora 42 packages does not install on Fedora 44:

```shell
λ  sudo dnf install ./flameshot-v14.0+git0.da6121bd-artifact-fedora-42-x86_64/flameshot-14.0.0-1.fc42.x86_64.rpm
Place your right index finger on the fingerprint reader
Updating and loading repositories:
...
Repositories loaded.
Failed to resolve the transaction:
Problem: conflicting requests
  - nothing provides qt6-svg(x86-64) needed by flameshot-14.0.0-1.fc42.x86_64 from @commandline
You can try to add to command line:
  --skip-broken to skip uninstallable packages
```

mmahmoudian:
I can confirm this:

```console
❯  sudo dnf install /home/mehrad/Downloads/flameshot-v14.0+git0.da6121bd-artifact-fedora-42-x86_64/flameshot-14.0.0-1.fc42.x86_64.rpm
[sudo] password for mehrad:
Updating and loading repositories:
 Fedora 44 - x86_64 - Updates                                                   100% |   6.0 KiB/s |  15.7 KiB |  00m03s
Repositories loaded.
Failed to resolve the transaction:
Problem: conflicting requests
  - nothing provides qt6-svg(x86-64) needed by flameshot-14.0.0-1.fc42.x86_64 from @commandline
You can try to add to command line:
  --skip-broken to skip uninstallable packages
```

I also checked and I have the `qt6-qtsvg` and `qt6-qtsvg-devel` installed. E.g.:

```console
❯  sudo dnf install qt6-qtsvg
[sudo] password for mehrad:
Updating and loading repositories:
Repositories loaded.
Package "qt6-qtsvg-6.11.1-1.fc44.x86_64" is already installed.

Nothing to do.
```

mmahmoudian:
I tried to bump the Fedora version in CI, hoping that it would solve the issue, assuming it comes from package name change, but it didn't solve it.

jankatins:
Yeah , the dependency is hardcoded in the spec file: 

```spec
%if 0%{?suse_version}
Requires: qt6-svg
%else
Requires: qt6-svg%{?_isa}
%endif
```

borgmanJeremy:
I believe this has been merged in #4777

PR #4785, opened 2026-06-29
bump fedora version in CI

mmahmoudian:
I tried to see if this would solve #4784, but it didn't. Alas, I bumped the two fedora versions our CI build for to the two last versions.

### DIFF
--- a/.github/workflows/Linux-pack.yml
+++ b/.github/workflows/Linux-pack.yml
@@ -179,15 +179,15 @@ jobs:
       matrix:
         dist:
           - {
-              name: fedora-41,
+              name: fedora-43,
               os: fedora,
-              symbol: 41,
+              symbol: 43,
               arch: x86_64
             }
           - {
-              name: fedora-42,
+              name: fedora-44,
               os: fedora,
-              symbol: 42,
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Does #4777  work?

jankatins:
4777 also doesn't work. See https://github.com/flameshot-org/flameshot/pull/4777#issuecomment-4841650953 for a possible fix... Looks like a copy and paste error and no one has tested the fedora package after the last change to the spec file :-/

mmahmoudian:
I close this in favor of #4777.

@jankatins 
> Looks like a copy and paste error and no one has tested the fedora package after the last change to the spec file :-/

I believe that's where the community comes in.

Flameshot devs in general have 2 options:
1. only package those that we have a feasible mean to test, and ignore other formats and distros (meaning we have VMs or machines to test and we have time on our hand to check them one by one)
2. package things to the best of our ability, and try to resolve the issues if/when bugs are reported

Neither would immediately solve the problem you are facing, but I'd argue that the 2nd approach gets you much closer to the solution, as the path is already paved and needs some improvement/modification to get it into a working state. Right?

Issue #4341, opened 2025-10-29
Screenshots on macOS look desaturated compared to the native screenshot tool (missing/incorrect color profile)


jankarres:
### Flameshot Version

v13.3.0 (88c951e)

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...), Using the ready-made package from Github Releases

### Operating System type and version

macOS 15.7.1 (Darwin 24.6.0)

## Summary

On macOS, screenshots saved with Flameshot appear less saturated than screenshots created with the native macOS screenshot tool. Side-by-side comparisons show Flameshot PNGs are **untagged** (no embedded ICC profile), while the native tool embeds the active display profile (wide-gamut). On color-managed systems/apps this yields washed-out colors.

## Expected Behavior
- Flameshot should assign/convert to a chosen output colorspace and **embed** the matching ICC profile (PNG `iCCP`) so color-managed apps render the screenshot with the same appearance as on screen.
- Visual parity with the native macOS screenshot.
 
## Actual Behavior
- Flameshot PNGs contain **no embedded ICC profile** (`profileName: <nil>`).
- On wide-gamut displays, untagged images are treated as sRGB, leading to visibly **desaturated** output compared to the native screenshot, which is tagged with the active display profile (`LG HDR 5K` in this case).
    

## Technical Analysis / Hypothesis
- macOS display pipeline often renders in a wide-gamut space (e.g., Display-P3 or an HDR-capable profile).
- Flameshot’s macOS path (Qt) appears to:
    1. read framebuffer/window pixels without a colorspace conversion, and
    2. save the PNG **without** embedding an ICC profile.
- The native tool either preserves the display profile or converts and embeds an appropriate profile, avoiding a gamut/appearance mismatch.

### Steps to reproduce

1. On a P3/HDR-capable Mac display, open vivid UI/web content.
2. Capture the same region twice:
    - with **macOS** (`⇧⌘4` / Screenshot.app),
    - with **Flameshot**.
        
3. Inspect both PNGs in Preview (`Tools → Show Inspector → More Info → Color Profile`) and via Terminal.

<img width="2982" height="1604" alt="Image" src="https://github.com/user-attachments/assets/caa4c3d5-4dd4-4dff-963e-4d2a8a6e184d" />
<img width="1491" height="805" alt="Image" src="https://github.com/user-attachments/assets/75979d07-e39f-4e15-a43f-5f270cb4f970" />

[...]

Issue #4615, opened 2026-04-04
Screenshot colors become faded when capturing Edge browser content with HDR enabled on Windows


r4496618:
### Flameshot Version

Environment:

Flameshot v13.3.0 (da913d7f)

Compiled with Qt 6.9.2

Windows 10.0.19045 (Windows 10)

### Installation Type

Flatpak from Github

### Operating System type and version

Windows 10.0.19045 (Windows 10)

### Description

When capturing web content from Microsoft Edge browser, the resulting screenshot has noticeably faded/washed-out colors (see attached screenshot 1).

I've traced the issue to the Windows HDR setting. When HDR is turned off, screenshots from Edge are captured correctly with normal colors. However, I prefer to keep HDR enabled and would rather not disable it as a workaround.

<img width="1286" height="821" alt="Image" src="https://github.com/user-attachments/assets/245171d8-6bc7-4e74-a91b-be294ab095a0" />

Is there any configuration or setting within Flameshot that could resolve this issue without disabling HDR? I would greatly appreciate any guidance.

Thank you very much for your help. This problem has been bothering me for a long time.

### Steps to reproduce

When capturing web content from Microsoft Edge browser, the resulting screenshot has noticeably faded/washed-out colors (see attached screenshot 1).

I've traced the issue to the Windows HDR setting. When HDR is turned off, screenshots from Edge are captured correctly with normal colors. However, I prefer to keep HDR enabled and would rather not disable it as a workaround.

### System Information

<img width="498" height="692" alt="Image" src="https://github.com/user-attachments/assets/47358935-50f5-40df-bca3-d48a5eaeca1a" />

tristan11232021-design:
Yup Same problem. it seams random though. I took one screenshot that turned out fine then when I switched windows my HDR must have kicked on because my screen suddenly got bright and since then it won't do it without blowing out the color. the weird thing is that when I use screen snip it locks the screen everything except edge which blows out and everything is over exposed.

mmahmoudian:
@tristan11232021-design 
> the weird thing is that when I use screen snip it locks the screen everything except edge which blows out and everything is over exposed.

Do youean that the Windows Snipping Tool also somewhat suffers from the same issue except for Edge browser?

PR #4787, opened 2026-06-30
Embed the display ICC profile in captures on macOS (#4341)

jankarres:
_Disclaimer: I'm not a C++/Qt developer. I put this together with AI assistance and tried hard to match the patterns and style already in the codebase, but it needs a review from someone who knows Qt well._

### Problem 

On a wide-gamut/HDR display the captured pixels are in the display's color space, but Flameshot saved the PNG with no ICC profile attached. Without that tag, other apps fall back to assuming sRGB, so the screenshot looks noticeably desaturated next to the one macOS' built-in tool makes — that one embeds the active display profile (`LG HDR 5K`, in my case).

The reason it happens: captures travel through the app as a `QPixmap`, which can't hold a `QColorSpace`, and every place that writes the image writes from that pixmap. So the display profile was never asked for and never written out.

### Fix

The display's color profile is now read once, at capture time, and travels alongside the capture until it's written:

- A small `ColorProfileProvider` asks CoreGraphics for the active display's profile (`CGDisplayCopyColorSpace` → ICC → `QColorSpace`). If that doesn't give us something usable it falls back to Display P3 for wide-gamut screens, and otherwise just leaves the image untagged like before. On non-macOS platforms it returns nothing, so those builds behave exactly as they do today.
- `ScreenGrabber` grabs the profile when it takes the screenshot and hands it on; it rides along on `CaptureContext` since the pixmap itself can't carry it.
- From there it's passed through `exportCapture` to each thing that writes an image, saving to a file, `--raw`, the clipboard, pins (including a pinned image's own copy/save), and the "Open With" temp file. The parameter defaults to empty, so nothing else in the codebase had to change. The actual tagging is just `QImage::setColorSpace`, which labels the pixels without recoloring them — which is what we want, because they're already in the display's space.

### Why macOS only

That's the only platform I could actually reproduce the bug on, so it's the only one I felt comfortable fixing. From what I can tell the same thing can affect wide-gamut displays on Windows and on Linux (X11/Wayland) as well, but I had no way to test there, so I deliberately left those paths untouched. I designed the provider so adding the other platforms later is mostly a matter of filling in `forScreen()` for each — the reasoning, trade-offs and a full test plan are in the write-up attached to #4341.

### Testing

There weren't any C++ tests in the repo yet, so I added a small QtTest suite (`tests/unit/test_colorprofile.cpp`) and hooked it into the CTest step that CI already runs. It's headless and platform-neutral: it covers the tagging helper, the fallback logic, and a PNG round-trip that proves the profile actually ends up embedded (plus a baseline showing an untagged image stays untagged).

I also checked it for real on my `LG HDR 5K` monitor: before, a saved capture had no profile; now `sips` reports `profile: LG HDR 5K`, same as the native screenshot tool.

### Known gaps

- Qt only embeds the profile when the `QColorSpace` is valid (matrix-based RGB); unusual display profiles fall back to Display P3.
- Imgur uploads still go out untagged for now. It's behind the `ENABLE_IMGUR` option (off by default), so I couldn't compile or test that path, and it runs through its own separate code rather than `exportCapture`.
- Windows/X11/Wayland are intentionally not addressed here.

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -2,6 +2,7 @@
 #include "screengrabber.h"
 #include "core/qguiappcurrentscreen.h"
 #include "utils/abstractlogger.h"
+#include "utils/colorprofileprovider.h"
 #include "utils/confighandler.h"
 #include "utils/monitorpreview.h"
 #include "utils/systemnotification.h"
@@ -242,6 +243,7 @@ QPixmap ScreenGrabber::grabEntireDesktop(bool& ok, int preSelectedMonitor)
     screenshot = currentScreen->grabWindow(
       wid, geom.x(), geom.y(), geom.width(), geom.height());
     screenshot.setDevicePixelRatio(currentScreen->devicePixelRatio());
+    m_colorSpace = ColorProfileProvider::forScreen(currentScreen);
     return screenshot;
 
 #elif defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
@@ -285,7 +287,14 @@ QPixmap ScreenGrabber::grabFullDesktop(bool& ok)
     QPixmap screenshot;
 
 #if defined(Q_OS_MACOS)
-    // On macOS, composite all screens into a single pixmap.
+    // On macOS, composite all screens into a single pixmap. Screens may have
+    // different color profiles, but the composite can carry only one. Convert
+    // each screen into a common target space (the primary display's) before
+    // compositing, then tag the result with that target so every region is
+    // correct — not just the primary one.
+    const QColorSpace target =
+      ColorProfileProvider::forScreen(QGuiApplication::primaryScreen());
+    m_colorSpace = target;
     const QList<QScreen*> screens = QGuiApplication::screens();
     QRect totalGeom;
     for (QScreen* s : screens) {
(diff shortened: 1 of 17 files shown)


borgmanJeremy:
@jankarres  Did you test the case where multiple monitors have different profiles? Could you please remove the unit tests? I don't think they will be long term useful or maintained so I don't want to add them.

If you try adding linux + windows it may solve: #4615 and #3151

jankarres:
Thanks for taking a look, and good questions.

**Multiple monitors with different profiles:** No, I didn't test that (as I have just one monitor), and to be honest it's a case I hadn't even considered until you brought it up. You're right that it matters. Originally the full desktop capture just tagged the whole stitched image with the primary display's profile, which is wrong for any region that comes from a second monitor with a different profile. I've now changed it so each screen gets converted into a common target space (the primary display's) before it's composited, and the final image is tagged with that. Screens that already use the same profile are left untouched, so there's no unnecessary reprocessing.

The honest caveat is that I only have a single monitor here, so I could confirm the single display path still works (a saved `flameshot full` still comes out tagged as `LG HDR 5K`, matching the native tool), but I couldn't try the mixed profile case myself. Do you have a setup with two differently profiled monitors where you could test it?

**Unit tests:** Removed, as requested.

**Linux and Windows (#4615 and #3151):** For what it's worth, from reading both they actually look like Windows HDR issues rather than the wide gamut tagging this PR is about. With HDR on, Windows works in an extended range (scRGB) and shows SDR content at an adjustable white level, and capturing that into a plain 8 bit sRGB image without tone mapping is what makes it come out washed out (#4615) or too bright (#3151), which also fits "turning HDR off fixes it". So an embedded color profile probably won't solve those on its own; they'd need proper HDR to SDR tone mapping. So I don't think this PR is the right place to fix the issue.

src/utils/screengrabber.cpp
borgmanJeremy, line 288: I can't quite put my finger on it, and its hard to be sure without the ability to test this, but this feels wrong. I think if there are two different monitors with different color profiles and you were to composite them like this, one would not look correct.

Issue #4312, opened 2025-10-26
Misleading "Use last region" in the configuration panel?


jpggithub:
### Flameshot Version

Flameshot v12.1.0 (Debian 12.1.0+ds-3)
Compiled with Qt 5.15.17
linux: 6.16.12+deb14+1-amd64
debian: testing

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Debian testing

### Description

Perhaps I miss a thing but I think that "Use last region" in the configuration panel can be misleading in some cases.

When it is checked and I launch "flameshot screen", it does not take a screenshot of the screen but only of the last region.

If it seems more intuitive to anyone, "flameshot screen" should be take a screenshot of the entire screen, even if "Use last region" is checked.

At least, I think that this counter-intuitive behavior  (IMHO 😉) could be indicated in the documentation.

### Steps to reproduce

1. run `flameshot gui`
2. check "Use last region"
3. make other things that using flameshot
4. launch `flameshot screen`

flameshot takes a screen shot of the region not the entire screen as expected by the command line.

### System Information

Debian GNU/Linux Testing
Flameshot v12.1.0 (Debian 12.1.0+ds-3)
Compiled with Qt 5.15.17
linux: 6.16.12+deb14+1-amd64
debian: testing

borgmanJeremy:
@mmahmoudian  I think I agree that if the screen subcommand is used, "Use last region" should be ignored.

mmahmoudian:
@borgmanJeremy 
I agree with you both. The only subcommand that should take into account the "use last region" is the `gui`. 

@jpggithub thanks for reporting this

serebryakov02:
I can reproduce this on macOS 26.3 (25D125), arm64, with a local build from commit 550d4fae.

I would like to work on this; if that is okay, could you please assign it to me?

borgmanJeremy:
Thanks!

serebryakov02:
@borgmanJeremy This issue can be closed.

PR #4789, opened 2026-06-30
Fix screen capture ignoring GUI last-region setting

serebryakov02:
Addresses #4312.

This PR makes **Use last region for GUI mode** apply only to `flameshot gui`.

Before this change, the last region was applied inside `CaptureRequest` for graphical requests. Because of that, `flameshot screen` could reuse the previous GUI region instead of capturing the full screen.

Now the last-region setting is handled only in the `gui` command path. The `screen` command ignores it, as discussed in the issue.

While testing the original reproduction steps on macOS, I also found that `flameshot screen` without `--number` was aborting. The screenshot was grabbed, but the selected monitor was not stored, so the command failed afterward. This PR stores the current monitor index in the macOS screen path so the original `flameshot screen` command works.

Regression risk should be low:
- The behavior matches the maintainer direction that only `gui` should use the last-region setting.
- The macOS monitor fix only affects macOS.
- Linux and Windows screen capture paths are not changed by the macOS-specific fix.

Tested on macOS 26.3 arm64:
- `gui` reuses the last region when the setting is enabled.
- `screen` ignores the last GUI region and captures the full monitor.
- `screen` without `--number` no longer aborts.

### DIFF
--- a/src/core/capturerequest.cpp
+++ b/src/core/capturerequest.cpp
@@ -2,12 +2,7 @@
 // SPDX-FileCopyrightText: 2017-2019 Alejandro Sirgo Rica & Contributors
 
 #include "capturerequest.h"
-#include "config/cacheutils.h"
-#include "utils/confighandler.h"
 
-#include <QApplication>
-#include <QClipboard>
-#include <QDateTime>
 #include <stdexcept>
 #include <utility>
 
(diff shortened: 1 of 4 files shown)


borgmanJeremy:
This breaks using the last region during my testing (linux, KDE, wayland).

serebryakov02:
@borgmanJeremy Thank you, I found the issue. I had moved `saveLastRegion()` handling out of `CaptureRequest`, but only restored it in the CLI `gui` path. That missed other graphical capture entry points such as tray/DBus/shortcut captures.

I pushed a follow-up in https://github.com/flameshot-org/flameshot/pull/4789/changes/3e5fe397f1f943ec8e734ddea58a5d3ddf20b171 that applies the saved-region default centrally in `Flameshot::gui(...)` when no explicit initial selection is already set. `screen` still uses `SCREEN_MODE`, so it should not inherit the saved region.

Local checks passed: clang-format dry run, `git diff --check`, and `cmake --build build --target flameshot`. I cannot test KDE/Wayland locally, so confirmation on that setup would be appreciated.

borgmanJeremy:
Thank you!

Issue #4754, opened 2026-06-19
Flameshot 14.0 final macOS DMG for Intel is Apple Silicon-only


HackintoshHD:
### Flameshot Version

Flameshot 14.0 final, macOS Intel (x86_64) DMG

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

macOS Sequoia 15.7.7 Intel

### Description

First of all, I would like to thank all developers of Flameshot for this incredible tool - your work is highly appreciated!

**Version:** Flameshot 14.0 final, macOS Intel (x86_64) DMG

Despite its 'Intel' label, the released macOS DMG for Intel is Apple Silicon-only and will not execute on an Intel Mac, even if launched directly via shell ('bad CPU type').

This could be considered a regression to the Flameshot 14.0.rc3 release, where the macOS DMG for Intel at least contained executable Intel code (cf. https://github.com/flameshot-org/flameshot/issues/4683).

### Steps to reproduce

1. Download the Flameshot 14.0 macOS DMG for Intel
2. Open the macOS DMG and install the application (Gatekeeper allowance necessary)
3. Try to launch the application normally and via shell

### System Information

15" MacBook Pro w/ Intel Core i7 running macOS Sequoia 15.7.7 Intel

tessus:
I wanted to open an issue, but it seems someone beat me to it. 

<img width="299" height="83" alt="Image" src="https://github.com/user-attachments/assets/ef14401a-6580-4baf-9bc3-573e22533258" />

```
/Volumes/Flameshot/Flameshot.app/Contents/MacOS$ file flameshot
flameshot: Mach-O 64-bit executable arm64
```

matidev:
@HackintoshHD @tessus - on the [README of v14](https://github.com/flameshot-org/flameshot/releases/tag/v14.0.0), they already mentioned about this issue

> MacOS intel build is no longer included. The CI was producing binaries users reported as not working in the beta and I have no ability to troubleshoot it

borgmanJeremy:
I attempted to fix it at the last minute, but it appears it did not work. This issue needs to be solved by the community, if someone can fix the CI and test the build I'm happy to merge the fix.

HackintoshHD:
Thank you @borgmanJeremy, @matidev and @tessus for your reactions!

[...]

PR #4795, opened 2026-07-01
Fixing the currently broken macOS packaging for x64 (Intel)

HackintoshHD:
Fixing the currently broken macOS packaging for x64 (Intel) with a deployment target of macOS Sonoma 14.

### DIFF
--- a/.github/workflows/MacOS-pack.yml
+++ b/.github/workflows/MacOS-pack.yml
@@ -5,41 +5,42 @@ on:
     branches:
       - master
     paths-ignore:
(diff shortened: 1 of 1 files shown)


HackintoshHD:
This should hopefully fix https://github.com/flameshot-org/flameshot/issues/4754.

borgmanJeremy:
Thank you, I will include this in a 14.1 release which also should fix some other issues.

borgmanJeremy:
I tested this on an m1 air and it works there too.

PR #4796, opened 2026-07-01
Add side panel widget to manually edit geometry

### LINKED ISSUE #4703: Lock current screenshot dimensions label in place, display it while resizing the frame and allow user to input dimensions from keyboard.
### Feature Description

Its really inconvenient that screenshot selection dimensions aren't at a reasonable set location and there is no option to enable it being static. Its really frustrating at times, when you need to, for example, set screenshot frame over to be exactly 500x500, and not only you can't manually type it in from keyboard, you can't see the dimensions while resizing to get exactly a 500x500 frame. It would definitely benefit the app if those features were included.

milnadar:
An attempt to resolve https://github.com/flameshot-org/flameshot/issues/4703

This PR adds a "Grab area" widget in the side panel that allows for manual geometry edits

image

### DIFF
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -1271,6 +1271,13 @@ void CaptureWidget::initPanel()
             this,
             &CaptureWidget::setDrawColor);
     connect(m_sidePanel,
+            &SidePanelWidget::geometryChangeRequested,
+            this,
+            [this](QRect const &geometry){
+                if(m_selection)
+                    m_selection->setGeometry(geometry);
+            });
+    connect(m_sidePanel,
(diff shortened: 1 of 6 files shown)


borgmanJeremy:
@mmahmoudian I am okay with this feature but wanted your feedback before reviewing the code / merging.

mmahmoudian:
I haven't yet read the code, so last part of my comment might be already implemented 😅 sorry for that.

I think it is a nice feature, but perhaps not something that everyone use frequently considering the CLI `--region` and launvher. I wonder if it is possible to have it in a foldable "drawer" (don't know the C++ lingo for this GUI element) or only show it when no tool is selected or the move tool is selected.

Issue #2499, opened 2022-03-22
Esc undo all my input when I select text tool


haosdent:
### Flameshot Version

Flameshot v11.0.0 (ad1bf28)
Compiled with Qt 5.15.2
darwin: 21.2.0
osx: 12.1

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

MacOS

### Description

When I try to add some text in the screenshot, suppose I am tying 

"hello", then I type Esc key.

Then all my input would become lost. What my expectations should be "hello" is added into the screenshot and exit the text adding mode.

Is there anyway I could work around this?

### Steps to reproduce

1. Screenshot
2. Click "add text" tool
3. Type "hello"
4. Type Esc key.

### System Information

Mac OS 12.1 (21C52)

mmahmoudian:
I'm not sure if this should be considered as a bug. Afterall, the <kbd>Esc</kbd> is typycally used to cancel the current operation. When using the Text tool, one can use <kbd>Ctrl</kbd>+<kbd>Enter</kbd> (in case of Mac probably <kbd>⌘</kbd>+<kbd>Return</kbd>) or left-clicking outside of the text box to confirm the text. This behavior (<kbd>Ctrl</kbd>+<kbd>Enter</kbd>) is also very commonly used in various software.

All these said, I acknowledge that in Inkscape, LibreOffice Draw, and Gimp the <kbd>Esc</kbd> will not destroy the text and they just jump out of insert mode while keeping the textbox selected and ready to me moved by mouse or keyboard.

I suggest we consider this as an Enhancement.

haosdent:
> All these said, I acknowledge that in Inkscape, LibreOffice Draw, and Gimp the Esc will not destroy the text and they just jump out of insert mode while keeping the textbox selected and ready to me moved by mouse or keyboard.

Yes, @mmahmoudian 
And when draw line in flameshot, type `Esc` would remove the line, this is wired as well.

mmahmoudian:
> And when draw line in flameshot, type <kbd>Esc</kbd> would remove the line, this is wired as well.

This part I strongly disagree. When drawing a line/shape/arrow the user is actively holding down the left-click and pressing <kbd>Esc</kbd> completely make sense to cancel the current operation. If one wants to keep the drawing they should just let go of the left-click instead of pressing <kbd>Esc</kbd>.

[...]

PR #4804, opened 2026-07-06
Commit text on Esc in text tool

serebryakov02:
This addresses #2499.

The fix is scoped to the Text tool: when the active text editor receives `Esc`, it consumes the shortcut and requests the same commit path used when clicking outside the text box. This keeps the typed text and exits editing mode without changing the existing `Esc` behavior for in-progress drawing operations like lines, shapes, or arrows.

### DIFF
--- a/src/tools/capturetool.h
+++ b/src/tools/capturetool.h
@@ -78,7 +78,9 @@ public:
         // increase tool size for all tools
         REQ_INCREASE_TOOL_SIZE,
         // decrease tool size for all tools
-        REQ_DECREASE_TOOL_SIZE
+        REQ_DECREASE_TOOL_SIZE,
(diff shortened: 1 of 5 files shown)


serebryakov02:
I noticed this is the second time overall that I have run into issues with failing CI checks while contributing here. In this case, the failure was caused by clang-format expectations that I missed locally before pushing.

I will take this as a reminder to review the project’s CI checks more carefully and align my local workflow with them, especially formatting and linting, so I can avoid repeating the same mistake in future updates.

The formatting issue has been fixed and the branch has been updated, so this should be ready for review now.

mmahmoudian (approved):
LGTM

borgmanJeremy:
Thanks!

Issue #4721, opened 2026-06-10
Allow screen selection with keyboard


frulio:
### Feature Description

When I take a screenshot in a dual monitor configuration, a window with the 2 monitors appears with the text "Monitor XXX - Click to select." I would appreciate it if I could move with the arrows to highlight which one to use and enter to take the screenshot instead of having to use the mouse.

Granted, you have to use the mouse if you want to take a screenshot of a region, but at least it is the last step.

mmahmoudian:
I agree with this suggestion. Maybe also showing some small number on the top left corner of the small image of each monitor so that the user can also press the number and select the monitor.

borgmanJeremy:
I also agree. I would say cap it at 9 monitors starting with 1 and ending with 9?

theheckwr:
Additionally to selecting a screen via number (which should cover most setups) I would suggest to select via tab or the arrow keys and confirm via Enter / Space.

serebryakov02:
Hello! I would like to work on this issue. If no one else is currently working on this, could you please assign it to me?

borgmanJeremy:
@serebryakov02 Sure that would be great thanks!

serebryakov02:
This issue has been addressed and can be closed.

PR #4811, opened 2026-07-08
Add keyboard controls to monitor selector

serebryakov02:
Addresses #4721.

I implemented keyboard support for the multi-monitor selection dialog:

- Added visible number badges for monitors 1-9.
- Added direct monitor selection with number keys 1-9.
- Added keyboard navigation with Tab and arrow keys.
- Added confirmation with Enter / Space.
- Kept Escape cancellation and mouse-click selection behavior.

I would appreciate extra testing and feedback on Linux and Windows, since the normal monitor selection dialog is not shown on macOS in the current code path. To test the behavior locally on macOS, I temporarily bypassed the macOS-specific early return so I could exercise the selector UI, but that workaround is not included in this PR. The submitted change keeps the existing macOS behavior intact and only updates the shared monitor selector logic used where the selector is normally active.

### DIFF
--- a/src/utils/monitorpreview.cpp
+++ b/src/utils/monitorpreview.cpp
@@ -16,38 +16,48 @@ MonitorPreview::MonitorPreview(int monitorIndex,
                                QWidget* parent)
   : QWidget(parent)
   , m_monitorIndex(monitorIndex)
+  , m_selected(false)
+  , m_mouseHovered(false)
+  , m_imageLabel(nullptr)
+  , m_keyLabel(nullptr)
+  , m_textLabel(nullptr)
 {
     QVBoxLayout* layout = new QVBoxLayout(this);
     layout->setContentsMargins(10, 10, 10, 10);
     layout->setSpacing(10);
 
-    QLabel* imageLabel = new QLabel(this);
-    imageLabel->setAlignment(Qt::AlignCenter);
-    imageLabel->setPixmap(thumbnail);
-    imageLabel->setStyleSheet(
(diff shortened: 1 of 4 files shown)


mmahmoudian:
Thanks. I quite like the number, tab, and arrow navigation mix.

One change which was added that was not in your PR comment was the addition of setting alpha values for the text. I don't know why someone wants to do that, but well, it is there if they want to change it.

serebryakov02:
@mmahmoudian Thanks for pointing that out.

To clarify, I did not intend to add a new user-facing text opacity option. The previous code already used alpha values in the text label background: `rgba(..., 200)` for the normal state and `rgba(..., 220)` for hover. I moved that duplicated styling into `updateStyle()` so the same visual treatment can be reused for both mouse hover and keyboard highlight.

So the text alpha behavior itself is not new; it is mostly a small refactor. The new visual part is that keyboard-highlighted monitors now use the same highlighted style as mouse hover.

serebryakov02:
@mmahmoudian Please let me know if you would prefer this to be changed or reverted.

mmahmoudian:
I see. Thanks for clarification. I misread the code initially. Now i see it in line 97

```cpp
    const int textAlpha = highlighted ? 220 : 200;
```

Sorry for my wrong comment 👍

CaptaiNiveau:
I've tried this on Hyprland, was a breeze thanks to nix :D
Anyway, it worked as expected! 
I've changed my binds and all to use the build output. If anything crops up, I'll let you know.

borgmanJeremy (approved):
Thank you!! This works great.

PR #4822, opened 2026-07-15
fix: pin button hidden when -p flag used

luyanfeng:
When CLI flags like -p (save path) are specified, initButtons() unconditionally removed TYPE_PIN (and other) buttons from the editor UI, making pin unavailable.

Fix: only remove buttons whose corresponding task flag is already set in the request, instead of blanket-removing all action buttons.


luyanfeng (commented):
.

Issue #4733, opened 2026-06-12
flameshot 14.0.0-1 not work


als-creator:
### Flameshot Version

flameshot on terminal:
qt.qpa.theme.gnome: dbus reply error: [ "org.freedesktop.DBus.Error.NameHasNoOwner" ] "Could not activate remote peer 'org.freedesktop.portal.Desktop': startup job failed"
qt.qpa.theme.gnome: dbus reply error: [ "org.freedesktop.DBus.Error.NameHasNoOwner" ] "Could not activate remote peer 'org.freedesktop.portal.Desktop': startup job failed"

hot key on command "flameshot gui" don't work

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

archlinux

### Description

hot key on command "flameshot gui" don't work

### System Information

OS: ArchLinux x86_64
Kernel: Linux 7.0.11-zen1-1-zen
DE: Xfce4 4.20
Shell: fish 4.7.1
Terminal: guake
Display (MT-E27AF1M): 1920x1080 in 27", 75 Hz [External]
CPU: AMD Ryzen 5 5500GT (12) @ 4.46 GHz - 52.2°C
GPU: AMD Radeon Graphics (7) @ 1.90 GHz - 48.0°C (654.71 MiB / 3.96 GiB, 16%) []
Vulkan: 1.4.348 - radv [Mesa 26.1.2-arch1.1]
OpenGL: 4.6 (Compatibility Profile) Mesa 26.1.2-arch1.1
Board: B450M-A PRO MAX (MS-7C52) (1.0)

glektarssza:
Also seeing this. I'm running Xfce4 as well.

More complete system spec dump:

```
OS: Arch Linux x86_64
Host: ROG Strix G713RC_G713RC (1.0)
Kernel: Linux 7.0.11-arch1-1
Uptime: 1 h 54 m
Processes: 476
Packages: 2180 (pacman)
Shell: zsh 5.9.1

DE: Xfce4 4.20
WM: Xfwm4 (X11)

CPU: AMD Ryzen 7 6800HS (16) @ 4.79 GHz
GPU 1: NVIDIA GeForce RTX 3050 Mobile [Discrete]
GPU 2: AMD Radeon 680M [Integrated]
Memory: [ ■--------- ] 5.84 GiB / 62.06 GiB (9%)
[...]

Issue #4639, opened 2026-04-12
Screenshot portal timed out after 30 seconds


UltraBlackLinux:
### Flameshot Version

Flameshot v14.0.0 (016b9197)
Compiled with Qt 6.10.2

### Installation Type

User repository (AUR)

### Operating System type and version

6.19.9-zen1-1-zen, Endeavour OS, Hyprland 0.54.2

### Description

Getting the `Screenshot portal timed out after 30 seconds` error on hyprland after building flameshot from source. The release works as expected.

I had a look at existing issues which pointed me to installing `xdg-desktop-portal-wlr`, but that unfortunately didn't change anything.

### Steps to reproduce

`flameshot gui`

### System Information

Hyprland 0.54.2 (59f9f26)
Wayland

```
System:
  Host: lolcat Kernel: 6.19.9-zen1-1-zen arch: x86_64 bits: 64
  Desktop: Hyprland v: 0.54.2 Distro: EndeavourOS
Graphics:
  Device-1: Advanced Micro Devices [AMD/ATI] Navi 22 [Radeon RX 6700/6700
    XT/6750 XT / 6800M/6850M XT] driver: amdgpu v: kernel
  Device-2: Logitech Webcam C270 driver: snd-usb-audio,uvcvideo type: USB
  Display: wayland server: Xwayland v: 24.1.9 compositor: Hyprland v: 0.54.2
    driver: X: loaded: amdgpu unloaded: modesetting,radeon dri: radeonsi
    gpu: amdgpu resolution: 1: 2560x1440~144Hz 2: 1920x1080~60Hz
  API: EGL v: 1.5 drivers: kms_swrast,radeonsi,swrast
    platforms: gbm,wayland,x11,surfaceless,device
  API: OpenGL v: 4.6 compat-v: 4.5 vendor: amd mesa v: 26.0.3-arch1.1
    renderer: AMD Radeon RX 6700 XT (radeonsi navi22 ACO DRM 3.64
    6.19.9-zen1-1-zen)
  API: Vulkan v: 1.4.341 drivers: radv surfaces: N/A
  Info: Tools: api: clinfo, eglinfo, glxinfo, vulkaninfo
    gpu: corectrl,radeontop x11: xprop,xrandr
```

```
$ ls /usr/share/xdg-desktop-portal/*.conf
[...]

PR #4826, opened 2026-07-17
Fail fast when the screenshot portal cannot take the capture

jmcarthur:
Improves the error reporting behind the timeouts in #4639.

Since v14, every capture on Linux goes through the org.freedesktop.portal.Screenshot portal. Flameshot never checks the reply of the Screenshot D-Bus call. On systems where no portal backend provides that interface (misconfigured portals on Wayland, or X11 window managers, since xdg-desktop-portal-gtk 1.14 dropped its screenshot backend), the call fails within milliseconds, but flameshot keeps waiting for a Response signal that can never arrive and then reports a misleading "Screenshot portal timed out" error.

This change checks the reply and fails immediately with the real D-Bus error instead of a fake timeout. It does not change which systems can take screenshots. It makes the failures diagnosable, which is what the reports in #4639 mostly lack. The causes there include missing portal backends and an upstream Hyprland grim hang that really does time out.

Tested on EndeavourOS with i3 (X11, xdg-desktop-portal-gtk only, so no Screenshot interface on the bus). Before the change, `flameshot full` hung for 15 seconds and then failed. With it, the command fails in under a second and prints the real error, "No such interface org.freedesktop.portal.Screenshot". The useX11LegacyScreenshot path still captures normally. Systems with a working portal are unaffected.

A few notes for review:

- The unixScreenshot() helper folds two identical call sites into one. I think it belongs here, because both call sites need the new error handling. If you want a minimal diff for a point release, I'm happy to split it.
- The portal spec says clients should listen on the request handle from the reply, since portals before 0.9 return a different path than the token predicts. I left that out on purpose. Portals that old predate 2017, and the re-subscription is racy anyway.
- The new error string is one more burden for translators. If you would rather reword it, no objection.
- I have a follow-up branch that falls back to the native X11 grab when the portal is unavailable. I plan to propose it separately, since it touches the Legacy X11 policy discussed in #4733 and should not hold this diagnostic fix hostage.

Claude Code (an AI tool) wrote this patch. I directed the work and read the diff, but I don't know this codebase well and my C++ is rusty, so my review was shallow. Review accordingly. I did test it on my machine, which hits this bug. Master hangs for 15 seconds and fails, and this branch reports the real error in under a second.

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -29,6 +29,7 @@
 #if !(defined(Q_OS_MACOS) || defined(Q_OS_WIN))
 #include "request.h"
 #include <QDBusInterface>
+#include <QDBusMessage>
 #include <QDBusReply>
 #include <QDir>
 #include <QUrl>
@@ -49,7 +50,9 @@ ScreenGrabber::ScreenGrabber(QObject* parent)
     QImageReader::setAllocationLimit(1024);
 }
 
-void ScreenGrabber::freeDesktopPortal(bool& ok, QPixmap& res)
+ScreenGrabber::PortalStatus ScreenGrabber::freeDesktopPortal(
+  QPixmap& res,
+  QString& errorDetail)
 {
 
 #if !(defined(Q_OS_MACOS) || defined(Q_OS_WIN))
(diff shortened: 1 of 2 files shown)


borgmanJeremy:
This conceptually looks fine. I have only reviewed it but not tested yet.

PR #4828, opened 2026-07-18
feat: add Tesseract OCR text recognition tool {English Support}

SniperRavan:
This Pull Request introduces a new Text Recognition (OCR) action tool  
  to Flameshot, allowing users to select an area of their screen, run OCR, 
  copy the extracted text to the clipboard, and send a desktop notification
  with the results.                                                        
                                                                           
    At this initial stage, the tool is optimized and configured for        
  **English language recognition** (`eng`).


mmahmoudian:
Checkout this link:

https://github.com/flameshot-org/flameshot/pulls?q=is%3Apr+ocr

SniperRavan:
> Checkout this link:
> 
> https://github.com/flameshot-org/flameshot/pulls?q=is%3Apr+ocr

Thanks for sharing the link. I wasn't entirely sure what you wanted me to take away from it, so I thought I'd explain why I opened this PR.

I'm fairly new to Linux, having recently switched from Windows. One of the few features I genuinely miss is built-in OCR support in screenshot tools. After trying Flameshot, I really liked it—it has become part of my daily workflow because it's lightweight, intuitive, and does its job well.

The one thing I consistently find myself missing is OCR. Right now, if I want to extract text from a screenshot, my workflow is something like:

1. Take the screenshot with Flameshot.
2. Open it in Google Lens (or another OCR service).
3. Copy the extracted text.
4. Paste it wherever I need it.

It's a surprisingly time-consuming process for something I do frequently.

That's what motivated me to implement this feature. My approach performs OCR locally, so the image never has to leave the user's device, which is an important privacy advantage over cloud-based alternatives. I know the recognition may not be perfect in every case, but for most everyday use it's accurate enough and significantly more convenient than switching between multiple applications.

If you shared the link because there have already been previous OCR proposals, I understand. I just wanted to explain the motivation behind my contribution. From a new user's perspective, OCR feels like a natural feature for a screenshot utility, and I believe it would improve the overall user experience if implemented in a way that remains optional and doesn't add unnecessary complexity for users who don't need it.

I'd be interested to hear the maintainers' thoughts on why previous OCR proposals weren't merged, as I'd like to better understand the project's direction.

mmahmoudian:
> I'm fairly new to Linux, having recently switched from Windows.

Welcome to the Linux world.


> I'd be interested to hear the maintainers' thoughts on why previous OCR proposals weren't merged, as I'd like to better understand the project's direction.

At least so far the consensus among devs have been that OCR is out of the scope of a screenshot tool. It is something that definitely useful, and therefore so far the decision have been to keep it as an idea for plugin. Two of the reasons for this are:
1. It would add tesseract as a hard dependency of Flameshot 
2. The OCR will be called only after the screenshot is taken. So a simple Linux pipe in terminal handles it.:

```sh
flameshot gui --raw | tesseract -l fin stdin stdout
```

I wrote this blog post 5 years ago in which I'm using the code above:

https://mehrad.ai/posts/20210702-extracting-payment-info-in-rasterized-invoice/

Alas, I'm not against the idea, but we should consider if it worth increasing the runtime dependencies for everyone in every OS.

What do you think about having it as a plugin?

SniperRavan:
> > I'm fairly new to Linux, having recently switched from Windows.
> 
> Welcome to the Linux world.
> 
> > I'd be interested to hear the maintainers' thoughts on why previous OCR proposals weren't merged, as I'd like to better understand the project's direction.
> 
> At least so far the consensus among devs have been that OCR is out of the scope of a screenshot tool. It is something that definitely useful, and therefore so far the decision have been to keep it as an idea for plugin. Two of the reasons for this are:
> 
> 1. It would add tesseract as a hard dependency of Flameshot
> 2. The OCR will be called only after the screenshot is taken. So a simple Linux pipe in terminal handles it.:
> 
> ```shell
> flameshot gui --raw | tesseract -l fin stdin stdout
> ```
> 
> I wrote this blog post 5 years ago in which I'm using the code above:
> 
> https://mehrad.ai/posts/20210702-extracting-payment-info-in-rasterized-invoice/
> 
> Alas, I'm not against the idea, but we should consider if it worth increasing the runtime dependencies for everyone in every OS.
> 
> What do you think about having it as a plugin?

Thanks for the explanation—it makes a lot more sense now. I understand the concern about adding Tesseract as a mandatory dependency and keeping Flameshot focused on its core purpose.

I like the idea of making OCR optional. If a plugin architecture is something the project is interested in, I think OCR would be a great candidate for it, since users who need it could install it without affecting everyone else.

My main motivation came from a user experience perspective. As someone new to Linux, OCR was one of the first features I looked for in a screenshot tool, and I found myself switching between multiple applications to accomplish something that felt like a natural extension of taking a screenshot. That's why I explored implementing it.

I'd be happy to explore a plugin-based approach if that's the direction the project prefers. Could you point me to any existing plugin architecture or design discussions for Flameshot, or would this be something that would need to be designed first?

mmahmoudian:
> As someone new to Linux, OCR was one of the first features I looked for in a screenshot tool, and I found myself switching between multiple applications to accomplish something that

As a solution, you can try what I proposed in my blog post, as it uses Flameshot and pipe the raw image to tesseracts. You already have both installed and all you need to do is to create a keyboard shortcut in your Desktop Environment (e.g., Shift+PrintScreen).

> I'd be happy to explore a plugin-based approach if that's the direction the project prefers. Could you point me to any existing plugin architecture or design discussions for Flameshot, or would this be something that would need to be designed first?

We have a minimal plugin setup as a PR: #3661

And we have the following Plugin System design in mind, but not yet implemented:

SniperRavan:
> > As someone new to Linux, OCR was one of the first features I looked for in a screenshot tool, and I found myself switching between multiple applications to accomplish something that
> 
> As a solution, you can try what I proposed in my blog post, as it uses Flameshot and pipe the raw image to tesseracts. You already have both installed and all you need to do is to create a keyboard shortcut in your Desktop Environment (e.g., Shift+PrintScreen).
> 
> > I'd be happy to explore a plugin-based approach if that's the direction the project prefers. Could you point me to any existing plugin architecture or design discussions for Flameshot, or would this be something that would need to be designed first?
> 
> We have a minimal plugin setup as a PR: #3661
> 
> And we have the following Plugin System design in mind, but not yet implemented:
> 
> #2529

Thanks for pointing me to the plugin RFC and the implementation PR. After reading through them, I understand the long-term direction much better.

Since the plugin framework is still under development, would you recommend waiting until the API stabilizes before adapting OCR into a plugin, or is there a part of the plugin system where additional contributors would be helpful right now?

I'm interested in contributing, and if OCR is better suited as a plugin rather than a core feature, I'm happy to move in that direction once the necessary infrastructure is available.

mmahmoudian:
Thanks for looking into the RFC and being willing to contribute to the software you use.

I will defer to @borgmanJeremy on this matter.

borgmanJeremy:
I would reccomend waiting. I haven't seen (or created) and API interface I am happy with. I've taken a couple of stabs and reviewed others proposals and none are quite right.

SniperRavan:
> I would reccomend waiting. I haven't seen (or created) and API interface I am happy with. I've taken a couple of stabs and reviewed others proposals and none are quite right.

Thanks for the clarification. That makes sense.

I'll wait until the plugin API has a more stable design before trying to adapt OCR into a plugin. In the meantime, I'll keep improving my OCR implementation locally and follow the progress of the plugin framework.

If there's an opportunity to help with the plugin infrastructure or to use OCR as a reference plugin once the API is ready, I'd be happy to contribute.

PR #4831, opened 2026-07-19
added a no ai policy

ghost:


borgmanJeremy:
We do not have a no AI policy.

Issue #3062, opened 2023-01-19
Bundle AVIF and JPEG XL support into Windows installer


novomesk:
### Feature Description

Hello,

would you agree to include Qt plug-ins for AVIF and JPEG XL into Windows installer of Flameshot? So Windows users would have the possibility to save into these formats directly.

I am author of those plug-ins and I build Windows versions of the plug-ins on AppVeyor using Microsoft Visual Studio 2019 and Qt 5.15.2 . I think Flameshot for Windows is built using the same.

I did an experiment:

1. I took latest Flameshot Artifacts from [AppVeyor](https://ci.appveyor.com/project/borgmanJeremy/flameshot/builds/45737640/artifacts) and installed it.
2. I added [qavif.dll](https://github.com/novomesk/qt-avif-image-plugin/releases/tag/v0.5.1) and [qjpegxl.dll](https://github.com/novomesk/qt-jpegxl-image-plugin/releases/tag/v0.4.2) into `C:\Program Files\Flameshot\bin\imageformats\` folder.
3. Created folders `data\mime\packages\` under `C:\Program Files\Flameshot\bin\` and added **freedesktop.org.xml** there (complete path is `C:\Program Files\Flameshot\bin\data\mime\packages\freedesktop.org.xml`)
4. Flameshot is able to save AVIF and JPEG XL (JXL) images on Windows!

freedesktop.org.xml comes from _shared-mime-info_  2.2 package and contains definition of `image/avif` and `image/jxl` MIME types.

mmahmoudian:
@veracioux @hosiet @ZetaoYang What do you think?

hosiet:
I don't have any opinion on Windows installer. I will try to look into support on Linux side.

novomesk:
BTW support for Linux is possible also via `kimageformats` package which contains similar plug-ins (if relevant dependencies are available).

PR #4847, opened 2026-07-27
Add AVIF and JPEG XL plugins on Windows

novomesk:
This is enough to enable AVIF and JXL on Windows ( https://github.com/flameshot-org/flameshot/issues/3062 ).

The MIME types are already included in Qt.

The `qavif6.dll` and `qjpegxl6.dll` were also built on AppVeyor (image Visual Studio 2022 using the same Qt version `C:\Qt\6.9.3\msvc2022_64` as current version of Flameshot on Windows), they work together.

I think the downloading binary DLLs is the most realistic solution, because AppVeyor has 1 hour timeout and it is not enough to built all dependencies in a single run.

If you start building Flameshot using Qt 6.11.1 (for example) in the future, the plugins should continue to work. But of course I can re-compile/update at any time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,14 @@ if(WIN32)
   # Include all dynamically linked runtime libraries such as MSVCRxxx.dll
   include(InstallRequiredSystemLibraries)
 
+  option(DOWNLOAD_EXTERNAL_QT_PLUGINS "Download DLL plugins for AVIF and JPEG XL" ON)
+  if(DOWNLOAD_EXTERNAL_QT_PLUGINS)
+    file(DOWNLOAD https://github.com/novomesk/qt-avif-image-plugin/releases/download/v0.10.3/qavif6.dll ${CMAKE_BINARY_DIR}/imageformats/qavif6.dll)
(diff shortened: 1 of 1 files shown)


mmahmoudian:
@novomesk it is always nice to see a familiar account. :)

Thanks for the PR. I wonder if there is a way to download and compile rather than fetching a binary blob.

novomesk:
> I wonder if there is a way to download and compile rather than fetching a binary blob.

AVIF plugin needs libavif and at least libaom, which should be built using NASM (otherwise encoding is very slow).

JXL plugin needs libjxl (contains google highway and brotli).

You can build them from source, but on AppVeyor there isn't enough time for all of them at the same time.

Flameshot alone takes 28-30 minutes.

AVIF plugin and its dependencies 12-15 minutes.

JPEG-XL plugin 8-16 minutes.

As I wrote, the build job would likely hit 1 hour timeout.

borgmanJeremy:
Thanks for the contribution but we will not ship these binaries.

novomesk:
Would it be acceptable to include AVIF and JXL plugins from https://invent.kde.org/frameworks/kimageformats on Windows?

Kimageformats are used by many people on Linux and Flameshot on Flathub also use it.

The DLLs and the needed libraries can be built in your repo where you will have everything under your control. I am willing to help to set up the scripts.

PR #4858, opened 2026-08-03
Update Japanese translation

coolvitto:
Compatible with v14.0

### DIFF
--- a/data/translations/Internationalization_ja.ts
+++ b/data/translations/Internationalization_ja.ts
@@ -29,12 +29,12 @@
     <message>
         <location filename="../../src/tools/accept/accepttool.cpp" line="31"/>
         <source>Accept</source>
-        <translation>承認</translation>
+        <translation>確定</translation>
     </message>
     <message>
         <location filename="../../src/tools/accept/accepttool.cpp" line="41"/>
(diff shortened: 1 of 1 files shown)


mmahmoudian:
Thank you for you contribution, but as explained in our documentation, we do translation through the following weblate:

https://hosted.weblate.org/engage/flameshot/

Would it be possible for you to transfer your valuable contribution to the weblate?

coolvitto:
Understood.
I will translate it using Weblate.

There are missing items in the Japanese translation on Weblate, so please update it.

coolvitto:
What I mean to say is that the Japanese translation has fewer characters than the original text.
480<655

sshot-1
sshot-2

mmahmoudian:
> What I mean to say is that the Japanese translation has fewer characters than the original text.

Do you man there are fewer phrases to be translated in Japanese?

coolvitto:
Judging by the figures on Weblate, it seems rather low to me.
However, the number of strings naturally varies depending on the language, so I might just be imagining things.

mmahmoudian:
Thanks. I will take a closer look at this.

Thank you for your willingness to contribute to this project. Your contributions are very valuable.

PR #4866, opened 2026-08-09
Use $APPIMAGE for autostart entry when running as AppImage

someone-ua:
When Flameshot runs as an AppImage, "flameshot" is not on PATH and QCoreApplication::applicationFilePath() points at the temporary squashfs mount, which disappears once the AppImage exits. Use the $APPIMAGE environment variable as the Exec path when set, falling back to "flameshot" otherwise, so the generated autostart .desktop file keeps working after reboot.

### DIFF
--- a/src/utils/confighandler.cpp
+++ b/src/utils/confighandler.cpp
@@ -319,9 +319,20 @@ void ConfigHandler::setStartupLaunch(const bool start)
     QFile file(path + "Flameshot.desktop");
     if (start) {
         if (file.open(QIODevice::WriteOnly)) {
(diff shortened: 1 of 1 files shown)


mmahmoudian:
I understand the code logic and it make sense [1], but I'm not sure if it works. Have you tested this yourself? 



1. https://docs.appimage.org/packaging-guide/environment-variables.html

someone-ua:
Yes, I have tested it. It works. There is now actual path to .AppImage in .desktop file so autostart works correctly.

PR #4867, opened 2026-08-09
perf: insert button list in sorted order instead of re-sorting

dionjoshualobo:
This PR optimizes button list ordering in `ButtonListView::updateActiveButtons` by inserting newly checked buttons directly at the correct priority position instead of appending and re-sorting the entire list.

Replaced:
  - `append` + `std::sort(...)`
With:
  - `std::lower_bound(...)` + `insert(...)`

What does the change do?

- Avoids re-sorting the full list on every insert.
- Keeps button order stable by `CaptureToolButton::getPriorityByButton(...)`.
- Completes the existing TODO about removing external sorts.


borgmanJeremy:
Thank you!

Issue #265, opened 2018-07-01
Flameshot does not handle 4K well


GXGOW:
**Describe the bug**

Whenever I try to take a screenshot with Flameshot, it only shows a quarter of the screen. More specific, the top left quarter of the screen.
![screenshot from 2018-07-01 11-14-48](https://user-images.githubusercontent.com/14166923/42132851-0a850e6c-7d20-11e8-9144-17a6941d02f0.png)

**To Reproduce**

* Open Flameshot
* Click 'Take Screenshot' from the tray icon

**Expected behavior**

I should be able to take a screenshot of the full screen.

**Desktop (please complete the following information):**

- Model: Dell XPS 15 9560
- OS: Manjaro Linux (Gnome Edition)
- Integrated Intel Graphics are used (i7 7700HQ)
- 4K display
- Wayland enabled

SwooshyCueb:
Workaround that Works For Me™:

unset the environment variables `QT_SCREEN_SCALE_FACTORS` and `QT_SCALE_FACTOR`, and set `QT_AUTO_SCREEN_SCALE_FACTOR` to `0` before launching flameshot.  
It may also help to unset `GDK_SCALE`, `GDK_DPI_SCALE`, and/or `ELM_SCALE`.

birdkiwi:
I have same problem in KDE on X.

hoshi411:
I have the same problem.
MS Surface Pro 3
4k display / scaling 
KDE NEON Developer

hoshi411:
> Workaround that Works For Me™:
> 
> unset the environment variables `QT_SCREEN_SCALE_FACTORS` and `QT_SCALE_FACTOR`, and set `QT_AUTO_SCREEN_SCALE_FACTOR` to `0` before launching flameshot.
> It may also help to unset `GDK_SCALE`, `GDK_DPI_SCALE`, and/or `ELM_SCALE`.

How do I implement this?  Is there a way to put this in the .desktop file?  If so, how?

dkowis:
I had this same problem, dual 4k displays on KDE in Ubuntu 18.04. Things wouldn't just look weird tho, the screen would get all super glitchy, like it "lost sync" or something. I would have to switch to another VT and kill the flameshot process to be able to see my desktop. 

Right now I just have it wrapped in a simple shell script and that's how I start it.

linuxkd:
[...]

PR #4872, opened 2026-08-12
Fix GNOME Wayland mixed-DPI capture scaling

### LINKED ISSUE #4871: GNOME Wayland mixed-DPI capture applies HiDPI scaling twice
### Flameshot Version

Flameshot v14.0.0 (da6121bd), compiled with Qt 6.2.4. The affected scaling path is also present on current `master` (`369f9ed9`).

### Installation Type

Linux package manager (`14.0.0-1`) and a build from source for diagnosis/testing.

Ubuntu 22.04.5 LTS, GNOME 42.9, Wayland.

### Description

On a GNOME Wayland mixed-DPI desktop, selecting the HiDPI monitor from the portal-based capture flow produces a blurred preview whose selection coordinates cover only the top-left quarter of the display.

The relevant geometry is:

- `HDMI-2`: 2560x1440, DPR 1, position +2298+0
- `eDP-1`: 3840x2400, DPR 2, position +0+1440
- GNOME portal screenshot: 4858x3840, DPR 1

Here, Qt 6.2 reports the HiDPI screen geometry in native coordinates (3840x2400), and the portal composite uses the same native-coordinate canvas. `ScreenGrabber::cropToMonitor()` therefore correctly crops 3840x2400 pixels, but then treats the portal scale of 1 as a mismatch with the screen DPR of 2 and upscales that crop to 7680x4800. Assigning DPR 2 after that makes the capture surface use the raw 3840x2400 geometry instead of its 1920x1200 device-independent size. This applies HiDPI scaling twice: the preview is blurred and pointer/selection coordinates cover only one quarter of it.

This is the same visible symptom described in #265 and is related to the mixed-DPI reports #227, #2849, and #3444. These reports are closed, several following #4498, but the GNOME Wayland native-coordinate case remains reproducible.

### Steps to reproduce

1. Use GNOME Wayland with one DPR 1 monitor and one DPR 2 monitor.
2. Run `QT_QPA_PLATFORM=wayland flameshot gui`.
3. In GNOME's portal dialog, share the desktop screenshot.
4. In Flameshot's "click to select a screen" preview, select the DPR 2 monitor.
5. Try to make a region selection across that monitor.

Actual result: the preview is lower resolution/blurred, and selection works only in the top-left quarter.

Expected result: the HiDPI monitor is shown at its native sharpness and the full monitor is selectable.

### Proposed fix

When GNOME Wayland returns a portal screenshot whose X and Y scale are both 1, keep the cropped pixels at their native resolution instead of scaling them to `geometry * DPR`. Then, when a Wayland screen's raw geometry size equals the screenshot's pixel size and DPR is greater than 1, size the capture surface and selection area using `geometry / DPR`.

I have tested this behavior on the hardware/configuration above; it restores a sharp preview and full-screen selection. A linked pull request contains the guarded implementation.

The geometry and scaling calculations above were captured with Flameshot's debug capture logging. I can provide additional logs if needed.

### System Information

- Ubuntu 22.04.5 LTS
- GNOME Shell 42.9
- Wayland session
- Qt 6.2.4
- Mixed-DPI layout: external 2560x1440 at DPR 1 plus internal 3840x2400 at DPR 2

lorenzo-di-gregorio:
Fixes #4871.

## Summary

- keep GNOME Wayland portal crops at native resolution when the portal canvas already matches Qt's native screen coordinates
- size the Wayland capture surface and selection area in device-independent coordinates when raw screen geometry matches the screenshot pixel size
- retain the existing scaling path for X11 and for Wayland portal images whose scale differs from 1

## Why

With a mixed-DPI GNOME Wayland layout, Qt 6.2 can report the DPR 2 monitor's geometry in native pixels while the GNOME portal returns a DPR 1 composite using the same native-coordinate canvas. The existing code crops the native-resolution monitor correctly, then scales it by DPR a second time. The result is a blurred preview with selection limited to the top-left quarter.

The guards in this change are based on the observed relationship between portal scale, screen geometry, screenshot pixel size, and DPR so that compositors which already expose logical geometry keep their current behavior.

## Testing

- `git diff --check`
- `cmake --build build --parallel 8`
- `ctest --test-dir build --output-on-failure` (the build defines no tests)
- manually tested on Ubuntu 22.04.5, GNOME 42.9, Wayland, Qt 6.2.4 with:
  - 2560x1440 external monitor at DPR 1
  - 3840x2400 internal monitor at DPR 2
  - 4858x3840 portal composite at DPR 1

The patched build restores native sharpness and allows selection across the complete HiDPI monitor.

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -758,7 +758,16 @@ QPixmap ScreenGrabber::cropToMonitor(const QPixmap& fullScreenshot,
 
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
     // Linux: May need rescaling if scale factors don't match
-    if (qAbs(screenshotScaleX - targetDpr) > 0.01) {
+    // GNOME Wayland's portal image uses the same native-coordinate canvas as
+    // QScreen::geometry() on Qt 6.2. In that case the cropped pixels are
+    // already native resolution; upscaling them by DPR creates a 4x-area
+    // image and limits interaction to the visible top-left quarter.
+    const bool nativeWaylandCanvas =
+      m_info.waylandDetected() &&
+      m_info.windowManager() == DesktopInfo::GNOME &&
+      qAbs(screenshotScaleX - 1.0) < 0.01 &&
+      qAbs(screenshotScaleY - 1.0) < 0.01;
+    if (!nativeWaylandCanvas && qAbs(screenshotScaleX - targetDpr) > 0.01) {
         int targetPhysicalWidth = qRound(targetGeometry.width() * targetDpr);
         int targetPhysicalHeight = qRound(targetGeometry.height() * targetDpr);
         cropped = cropped.scaled(targetPhysicalWidth,
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -194,6 +194,13 @@ CaptureWidget::CaptureWidget(const CaptureRequest& req,
             selectedScreen = QGuiApplication::primaryScreen();
         }
         QRect screenGeom = selectedScreen->geometry();
+        if (DesktopInfo().waylandDetected() &&
+            selectedScreen->devicePixelRatio() > 1.0 &&
+            screenGeom.size() == m_context.screenshot.size()) {
+            const qreal dpr = selectedScreen->devicePixelRatio();
+            screenGeom.setSize(QSize(qRound(screenGeom.width() / dpr),
+                                     qRound(screenGeom.height() / dpr)));
+        }
         move(screenGeom.topLeft());
         resize(screenGeom.size());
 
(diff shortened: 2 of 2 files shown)


borgmanJeremy:
Hello,

Appreciate the attempted fix, but I don't want to merge something that is specific to a specific Qt release (6.2) and a DE (gnome). If you can test this across a wide variety of DE's and Qt releases and post results I would take a second look at it.

lorenzo-di-gregorio:
Thanks—this is a fair concern. I verified the screen metrics with both Qt 5.15.3 and Qt 6.2.4 on the same GNOME Wayland session; both report the HiDPI display as 3840x2400 with DPR 2. Removing my local `QT_SCREEN_SCALE_FACTORS` setting produces the same result, so this is not unique to Qt 6.2. Historical issue #265 also reports the same symptom with Qt 5.11.

However, I cannot provide meaningful coverage across other desktop environments. The portal does not expose enough per-monitor scaling information to distinguish this case reliably from backends that intentionally return a lower-resolution image, so removing the GNOME guard would be unsafe.

If a default-off compatibility option is acceptable, I can narrow the PR to that. Otherwise, I am happy to close it and leave #4871 as documentation of the issue and working workaround.

PR #4874, opened 2026-08-14
Fix Wayland multi-monitor: show capture UI on the correct monitor

DIOR27:
## Summary

On Wayland, `xdg_toplevel` cannot be positioned and Qt's `primaryScreen` is unreliable, so the capture UI could open on the wrong monitor. This PR pins the capture widget to the selected screen and makes the monitor selection deterministic.

## Changes

- **`CaptureWidget`**: force the native window to exist (`winId()`) before calling `QWindow::setScreen()`, which recreates the window on the target output — the only reliable way to choose the monitor on Wayland. Without this the call was a silent no-op and the compositor mapped the fullscreen capture window on the first output.
- **`ScreenGrabber`**: when no screen is passed, pick the monitor under the cursor on X11; on Wayland (where the cursor position is not exposed to clients) fall back to the monitor the desktop reports as primary, parsed from the `org.freedesktop.portal.Desktop` monitor descriptions.

| File | Change |
| --- | --- |
| `src/utils/screengrabber.cpp` | Portal monitor-description parsing, cursor/primary monitor helpers |
| `src/utils/screengrabber.h` | Public helpers for monitor targeting |
| `src/widgets/capture/capturewidget.cpp` | Pin window to the selected screen via `QWindow::setScreen()` |

## Test plan

- [x] `cmake --build build-codebuff --target flameshot` passes
- [x] Manual, Wayland multi-monitor: capture UI opens on the monitor under the cursor

---

### Chain context

This is the first PR of a stacked chain (3 PRs):

1. 📍 **This PR** — fix Wayland monitor placement
2. → `feat/monitor-picker-autoselect` — auto-select monitor under cursor + compact picker on every screen
3. → `feat/hybrid-monitor-switching` — "Capture this monitor" strips + reuse stored screenshots

Out of scope: capture UX (picker/strips), screenshot handling — those land in the following PRs.

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -7,6 +7,7 @@
 #include "utils/systemnotification.h"
 
 #include <QApplication>
+#include <QDBusArgument>
 #include <QEventLoop>
 #include <QGuiApplication>
 #include <QHBoxLayout>
@@ -18,8 +19,10 @@
 #include <QPixmap>
 #include <QProcess>
 #include <QScreen>
+#include <QStringList>
 #include <QTimer>
 #include <QWidget>
+#include <QWindow>
 #include <algorithm>
 
 #ifdef FLAMESHOT_DEBUG_CAPTURE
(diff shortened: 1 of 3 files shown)


borgmanJeremy:
This all looks gnome specific which won't work for us since we try to be DE agnostic.

PR #4875, opened 2026-08-14
Wayland: auto-select monitor under cursor and compact picker on every screen

DIOR27:
## Summary

Improves the Wayland monitor picker: it auto-selects the monitor under the cursor (matching the X11 "capture active monitor" behavior) and renders compactly on every screen so multi-monitor setups stay readable.

## Changes

- **Auto-select**: the picker now pre-selects the monitor under the cursor instead of always the first one. On Wayland the cursor position is not exposed to clients, so the selection is derived from the portal monitor descriptions introduced in the base PR.
- **Compact picker**: preview boxes, margins and fonts scale down so the picker fits comfortably on each screen.
- **Click-only picker**: keyboard-based monitor selection (number keys, arrows, Enter) is removed. On Wayland the picker surfaces never receive keyboard focus, so those shortcuts were dead weight; selection is mouse-only, Escape still cancels.

| File | Change |
| --- | --- |
| `src/utils/monitorpreview.cpp` / `.h` | `compact` mode with smaller layout and fonts, click-only hint, no number key labels |
| `src/utils/screengrabber.cpp` / `.h` | Monitor-under-cursor selection, compact picker layout across screens, click-only input handling |
| `src/widgets/capture/capturewidget.cpp` | Picker sizing / auto-selection wiring |

## Test plan

- [x] `cmake --build build-codebuff --target flameshot` passes
- [x] Manual, Wayland multi-monitor: picker opens compact on every screen and pre-selects the monitor under the cursor
- [x] Manual, Wayland multi-monitor: click selects the monitor, Escape cancels, number keys no longer shown

---

### Chain context

Second PR of a stacked chain (3 PRs):

1. ← `fix/wayland-monitor-placement` (#4874) — base: fix Wayland monitor placement
2. 📍 **This PR** — auto-select monitor under cursor + compact picker on every screen
3. → `feat/hybrid-monitor-switching` — "Capture this monitor" strips + reuse stored screenshots

Out of scope: capture-time monitor switching — that lands in the next PR.

### DIFF
--- a/src/utils/monitorpreview.cpp
+++ b/src/utils/monitorpreview.cpp
@@ -13,18 +13,24 @@
 MonitorPreview::MonitorPreview(int monitorIndex,
                                QScreen* screen,
                                const QPixmap& thumbnail,
-                               QWidget* parent)
+                               QWidget* parent,
+                               bool compact)
   : QWidget(parent)
   , m_monitorIndex(monitorIndex)
   , m_selected(false)
   , m_mouseHovered(false)
+  , m_compact(compact)
   , m_imageLabel(nullptr)
   , m_keyLabel(nullptr)
   , m_textLabel(nullptr)
 {
+    const int margin = compact ? 6 : 10;
+    const int spacing = compact ? 6 : 10;
+    const int keyLabelSize = compact ? 22 : 28;
+
     QVBoxLayout* layout = new QVBoxLayout(this);
-    layout->setContentsMargins(10, 10, 10, 10);
-    layout->setSpacing(10);
+    layout->setContentsMargins(margin, margin, margin, margin);
+    layout->setSpacing(spacing);
 
     m_imageLabel = new QLabel(this);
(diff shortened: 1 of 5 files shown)


borgmanJeremy:
Im leaning towards not merging this but want to hear from @mmahmoudian  my main concern is it seems like the kind of hack that will come around and bite us later.

On tiling DE's like cosmic the window also disrupts the windows as everything resizes.

borgmanJeremy:
I thought about this more and i'm conceptually okay with it as long as its gated behind the same setting as X11 and windows rather than always capturing the active screen. I'm still skeptical it will work on every DE but I think enough users will be interested that its worth a try. 

I still need to review the code.

mmahmoudian:
My main concern is reverting #4811.

borgmanJeremy:
This should not revert #4811 once its behind the "Capture Active Monitor (skip monitor selection)" right?

Issue #3008, opened 2022-12-09
Remove strange black and white border around counter numbers


fabiochelly:
### Feature Description

When I insert counter numbers, there is strange triple border (white/black/white) around the circle. Is there a way 
![sandwich](https://user-images.githubusercontent.com/4060753/206737596-019b5e78-0a4b-41c5-8825-507c761a5afe.png)
to remove it?

mmahmoudian:
No, that is our design. Perhaps it can be set to be customizable at some point if you create a well-formatted feature request. The current format is not suitable to be considered as a feature request.

PR #4882, opened 2026-08-18
Add option to disable the Circle Counter outline

### LINKED ISSUE #3580: options of not draw the outline of the Circle Counter
### Feature Description

image

bekirsoylu:
Closes #3580

### Problem

The Circle Counter bubble is always drawn with a contrasting ring around it, and there is currently no way to turn it off.

The ring appears black for both light and dark draw colors, but by two different mechanisms in `CircleCountTool::process()`:

- an outer ellipse is **filled** with `antiContrastColor` — this is what shows on *dark* draw colors (e.g. red)
- the pen is set to `contrastColor` and **strokes both** ellipses — this is what shows on *light* draw colors (e.g. green)

So disabling only one of the two would still leave a dark border for half the palette.

### Change

Adds a `drawCircleCounterOutline` option plus a matching checkbox in *Configuration > General*, and documents it in `flameshot.example.ini`.

When the option is off, the outer ellipse is skipped and the pen is set to `Qt::NoPen` before the bubble is drawn, which removes both sources of the ring. The text pen is unaffected — it is reassigned to `contrastColor` before `drawText()` — so the number keeps its automatic contrast color.

### Default behaviour

The option defaults to `true`, so the rendering is byte-for-byte unchanged for existing users, and nothing is written to the config file unless the option is actually toggled.

### Notes

This was previously raised in #3008, where the response was that the ring is intended design and could be made customizable via a feature request. #3580 is that request; this PR implements it.

### Testing

Built and run on Windows 11 (MSVC 19.44, Qt 6.9.3, CMake 4.4). Verified that the checkbox is checked by default, that toggling it persists to `flameshot.ini`, and that counter bubbles render without the ring for both dark and light draw colors. `clang-format` reports no changes to the modified regions.

### DIFF
--- a/flameshot.example.ini
+++ b/flameshot.example.ini
@@ -73,6 +73,9 @@
 ;; Last used Circle Counter size (int)
 ;drawCircleCounterSize=1
 ;
+;; Draw a contrasting outline ring around the Circle Counter bubble (bool)
+;drawCircleCounterOutline=true
+;
 ;; Last used Pixelate pixel size (int)
 ;drawPixelateSize=2
 ;
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -79,6 +79,7 @@ GeneralConf::GeneralConf(QWidget* parent)
     initSquareMagnifier();
     initJpegQuality();
     initReverseArrow();
+    initDrawCircleCounterOutline();
     // this has to be at the end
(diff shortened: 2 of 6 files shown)


mmahmoudian:
LGTM

borgmanJeremy:
Thank you!

PR #4885, opened 2026-08-18
Add per-location save hotkeys, Imgur uploader, and reskin support

nicitaacom:
## Added
- **Save to loc1/2/3** — 3 new capture-tool actions, each saves straight to a pre-configured folder with no dialog, bindable to any hotkey from Configuration > Shortcuts
- **Imgur uploader on Ctrl+D** — enabled Flameshot's existing built-in uploader (was disabled by default), default-bound to Ctrl+D, with a new "Show upload result window" toggle (off by default — link just copies to clipboard silently)
- **Reskinnable UI** — app-wide QSS stylesheet support, swappable at runtime via `~/.local/share/flameshot-theme/current.qss`

## Fixed
- Side panel close animation was cut off before rendering (open animated, close didn't)
- `TYPE_SAVE_LOCATION_1/2/3` weren't registered as recognized shortcuts, which blocked the app with a config-error dialog on every launch once a key got bound to one of them

## Docs
- `dev_readme-hotkeys-shortcuts.md` — incident writeup + SOP for adding future hotkeys without repeating the registration gap above

## Screenshots
Shortcuts tab (new loc1/2/3 + Imgur entries):
image

Reskinned Configuration window:

image

Side panel (Ctrl+E / click target):
image

### DIFF
--- a/contrib/flameshot-theme/assets/check.svg
+++ b/contrib/flameshot-theme/assets/check.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
+  <path d="M3 8.5 L6.5 12 L13 4" fill="none" stroke="#ffffff" stroke-width="2.4" stroke-linecap="round" stroke-linejoin="round"/>
+</svg>
--- a/contrib/flameshot-theme/crazy-mechanics.conf
+++ b/contrib/flameshot-theme/crazy-mechanics.conf
@@ -1,0 +1,7 @@
+# crazy-mechanics :: default theme (active outside Halloween/New Year windows)
+# Source: nicitaacom/14_portfolio app/styles/theme-crazy-mechanics.css
+#   cta 277deg 100% 68% | brass 40deg 45% 55% | primary 0deg 0% 13%
+MAIN=#c05cff
+CONTRAST=#c09d59
+DRAW=#c05cff
+USER_COLORS="picker, #c05cff, #c09d59, #212121, #7c6892, #5f8c74"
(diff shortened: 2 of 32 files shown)


nicitaacom:
No AI that would tell if any changes close some issue - so I kept it as is - IDK if any of my commits close existing issues

___

## ✨ Added

- **Filterable UI language picker** — the "UI language" dropdown in Interface settings is now an editable, searchable combobox instead of listing every translation at once (which overflowed off-screen); typing narrows the list.

## 🐛 Fixed

**Better UI**
image

**No modal on every time I press a key on my keyboard**
18 08 2026 at 17-09

**No more list that overflows settings**

IMG_20260818_160704




- **General tab was cramped and clipped** — restructured into bordered, titled, scrollable sections (Startup, Notifications & Behavior, Save Path, Save Path Locations, Selection Geometry Display, etc). A `QGroupBox` populated via a redirected layout pointer was rendering clipped to its first child once nested inside the resizable scroll area.
- **Configuration window could grow past the screen** — capped to ~50% of the current screen's height on open.
- **Language picker saved on every keystroke** — switching the search box to editable meant `currentTextChanged` fired per character, spamming the "must restart" dialog and saving whatever partial text was typed; switched to `textActivated`, which only fires on an actual selection.
- **Mouse wheel over an unfocused dropdown/spin box in Configuration could silently change the setting** instead of scrolling the page — added an app-wide event filter (`WheelFocusGuard`) that ignores wheel events while unfocused. Known gap: doesn't catch every case where the field already holds keyboard focus — filed as a follow-up issue rather than chasing further, since attempts to also keep the page scrolling past the widget caused unreliable event redelivery and were reverted.

## 🔧 Changed

- **JPEG quality now defaults to 100** (was 75).

borgmanJeremy:
Please discuss feature changes before submitting a large PR. Each PR should address 1 change, not combine everything into 1. Of all of these items I would be interested in discussing the new theming options and cosmic changes to the settings, but likely not the other ones.

Issue #3730, opened 2024-09-19
Accessibility for color and size


Scary-Guy:
### Feature Description

So I have autism, and bright colors are a bit triggering.  On Linux I try to make everything follow a black theme or "dark mode" and have been doing that for 20 years.  Unfortunately not everything respects that and trying to get every graphics toolkit which wants to have its own custom theme engine is a real pain.  Anyway, the right click menu on the icon and the settings window are no exception, and I didn't see an option to enable a black theme.

I'm also pretty old and my eyes aren't what they used to be.  I could decrease my monitor's resolution but then the windows would be too big, so I try to increase the size of the fonts as much as possible.  Your app is a bit hard to read and if there were an option to increase the size of the UI elements that would really improve my UX.

GregariousJB:
After using this program for about 10 minutes, this was my only concern - the text size.

I'm at 2560x1440 and +125% Windows UI scale. I can read the text OK but I bet I'd have issues if I dropped the UI scaling back down to 100%. Something to consider for future updates.

PR #4889, opened 2026-08-19
Feat/qss theming - Add support for a user-swappable QSS theme

nicitaacom:
## Added
- **App-wide QSS stylesheet support** — on startup, Flameshot now loads `~/.local/share/flameshot-theme/current.qss` (if present) and applies it as the app-wide stylesheet, layered on top of the existing `uiColor`/`contrastUiColor` settings. If the file doesn't exist, nothing changes — fully opt-in.
- **`contrib/flameshot-theme/`** — example theme files (4 palettes: crazy-mechanics, halloween, new-year, green) plus two small shell scripts (`flameshot-theme` to apply one by name, `flameshot-seasonal-theme` to auto-pick one by month) that generate/install the QSS from a set of named color tokens (room/surface/border/accent/text/etc).

## Why
Loading is deferred one event-loop tick (`QTimer::singleShot(0, ...)`) because `qt6ct`-style platform theme integrations inject their own app-wide stylesheet shortly after startup, which would otherwise clobber this one if applied synchronously.

## Screenshots
Reskinned Configuration window:
image

### DIFF
--- a/contrib/flameshot-theme/assets/check.svg
+++ b/contrib/flameshot-theme/assets/check.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
+  <path d="M3 8.5 L6.5 12 L13 4" fill="none" stroke="#ffffff" stroke-width="2.4" stroke-linecap="round" stroke-linejoin="round"/>
+</svg>
--- a/contrib/flameshot-theme/crazy-mechanics.conf
+++ b/contrib/flameshot-theme/crazy-mechanics.conf
@@ -1,0 +1,7 @@
+# crazy-mechanics :: default theme (active outside Halloween/New Year windows)
+# Source: nicitaacom/14_portfolio app/styles/theme-crazy-mechanics.css
+#   cta 277deg 100% 68% | brass 40deg 45% 55% | primary 0deg 0% 13%
+MAIN=#c05cff
+CONTRAST=#c09d59
+DRAW=#c05cff
+USER_COLORS="picker, #c05cff, #c09d59, #212121, #7c6892, #5f8c74"
--- a/contrib/flameshot-theme/crazy-mechanics.qss
+++ b/contrib/flameshot-theme/crazy-mechanics.qss
@@ -1,0 +1,210 @@
+/* flameshot-theme :: crazy-mechanics
+ * Structure borrowed from the github-desktop-accent box-in-box panel
+ * language (surface/border/hover/selected), colors from the Crazy
+ * Mechanics design system (nicitaacom/14_portfolio).
+ *   room #1b1b1b | surface #242424 | raised #3a3542 | border #7d6f92
+ *   accent #c05cff | accent-bright #d68bff | on-accent #210633
+ *   text #e8e8e8 | secondary #8f8698
+ */
+
+QWidget {
+    background-color: #1b1b1b;
+    color: #e8e8e8;
+    selection-background-color: #c05cff;
+    selection-color: #210633;
+}
+
+QDialog, QMainWindow {
+    background-color: #1b1b1b;
+}
+
+QScrollArea, QAbstractScrollArea::viewport {
+    background-color: #1b1b1b;
+    border: none;
+}
+
+QGroupBox {
+    background-color: #242424;
+    border: 1px solid #7d6f92;
+    border-radius: 6px;
+    margin-top: 14px;
+    padding: 14px 10px 12px 10px;
+    font-weight: 600;
+}
+QGroupBox::title {
+    subcontrol-origin: margin;
+    left: 10px;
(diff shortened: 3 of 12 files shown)


mmahmoudian:
Thanks for the contributionband PR. Two small comments regarding the PR:

I would appreciate it if such major changes are discussed in a thread before implantation.

I see that the code-change to Flameshot core file is very minimal, and most of the 900+ diff changes are new files. I believe it would have been more logical to have one example theme in this repo (e.g., a dark theme), and we have other themes as a separate repo. Considering that there are bash files and etc. these are not cross-platform, plus these are optional things that can be considered as "bloat". I personally would rather keep the Flameshot core minimal and expansions (plugins, themes, etc.) separately, so that those who want them can take them but they are not shipped with the Flameshot directly.

I suggest not spending time on changing the PR until we also get feedback of @borgmanJeremy , other maintainers, and contributors.

I'll tag some folks here that I'm curious to know their opinion on this:

@hoiset @holazt @jack9603301 @panpuchkov @ElTh0r0 @veracioux @FelixJochems  @serebryakov02

serebryakov02:
Thanks for the ping, @mmahmoudian, and thanks @nicitaacom for separating this work from #4885. I reviewed both commits and the full diff.

I mostly agree with the suggested split. The general QSS support and the supplied themes/scripts are two separate concerns.

I think optional QSS support could be useful in Flameshot. It is not only a visual extra: custom colors, contrast, and font sizes can help with accessibility issues such as https://github.com/flameshot-org/flameshot/issues/3730 . However, the current core implementation needs more design work:

- The path `~/.local/share/flameshot-theme/current.qss` is Linux-specific and ignores standard locations on macOS, Windows, portable installations, and sandboxed packages.
- Flameshot already sets an application stylesheet. The new call replaces it instead of clearly combining the two stylesheets.
- `QTimer::singleShot(0, ...)` relies on timing to override platform themes. That order is not clearly guaranteed.
- Asset paths, error reporting, and compatibility with future Flameshot versions are not defined.
- There is no documentation or test coverage for the new behavior.

[Qt already supports a `-stylesheet <file>` option](https://doc.qt.io/qt-6/qapplication.html#QApplication). Flameshot’s current startup and stylesheet handling mean that this is not a complete solution today, but we should consider supporting Qt’s existing mechanism before adding a separate hidden file convention.

I would not keep the supplied theme collection and scripts in the main repository in their current form. The four 210-line QSS files have the same structure and differ only in colors, but there is no generator. Each file also contains the absolute path `/home/kali/.local/share/flameshot-theme/assets/check.svg`, so the checkbox icon will not work for other users.

The scripts also depend on one specific Linux setup. They use fixed configuration, data, and executable paths; require Bash, Python, `pkill`, and `setsid`; rewrite the user’s configuration without a backup; and load theme files as executable shell code. The `--revert` option does not restore the user’s previous settings - it applies another custom theme.

One small correction regarding “bloat”: these files are not included by the current CMake installation rules, so they do not increase the normal binary packages. The larger concern is maintenance. Flameshot would become responsible for about 960 lines of platform-specific scripts and repeated theme data, while installed users would not receive them through a documented installation method.

My preference would be:

- First agree on the design through the project’s [discussion/RFC process](https://github.com/flameshot-org/flameshot/blob/master/docs/RFC.md).
- If accepted, create a focused core PR with a standard or configurable path, clear stylesheet order, portable asset handling, useful error messages, documentation, and tests.
- Keep at most one portable and self-contained example theme in the main repository.
- Keep the full theme collection, seasonal switching, generators, and platform-specific installation scripts in a separate repository linked from the documentation. Any included styling should also have clear source and license information.

So my position is that the main idea is worth pursuing, but I would not merge the current implementation and scope as-is.

ElTh0r0:
> My preference would be:
> 
>     * First agree on the design through the project’s [discussion/RFC process](https://github.com/flameshot-org/flameshot/blob/master/docs/RFC.md).
> 
>     * If accepted, create a focused core PR with a standard or configurable path, clear stylesheet order, portable asset handling, useful error messages, documentation, and tests.
> 
>     * Keep at most one portable and self-contained example theme in the main repository.
> 
>     * Keep the full theme collection, seasonal switching, generators, and platform-specific installation scripts in a separate repository linked from the documentation. Any included styling should also have clear source and license information.

I nearly agree 100% with @serebryakov02 comment. Just my additional 2 cents:

- Seasonal switching is overkill for Flameshot in my opinion
- Install scripts should be completely avoided. Installation could/should be: Copy/extract folder with theme qss + additionally required files into a new to be defined Flameshot (user or global) sub folder. Flameshot scans this folder at application start and let the user select between all found themes through its config dialog.

Personally I don't need Flameshot specific themes, as I prefer consistent theming from my operating system for all applications, but if there is a _high_ users demand for theming, then the source code to be created and maintained for the Flameshot core should be as simple and short as possible!

PR #4891, opened 2026-08-19
Make the UI language picker a filterable search box

nicitaacom:
## Fixed
- **The UI language dropdown overflowed off-screen** — it lists every available translation at once, which is a long enough list to run past the bottom of the screen as a plain dropdown.

Made the combobox editable with a popup-filtering `QCompleter` (case-insensitive, match-contains), so typing narrows the list instead of showing everything.

Also switched the save handler from `currentTextChanged` to `textActivated` — the former fires on every keystroke once the box is editable, which would otherwise spam the "must restart" dialog and save whatever partial text was typed on each key press; `textActivated` only fires once an actual selection is made.

## Screenshot
image

### DIFF
--- a/src/config/visualseditor.cpp
+++ b/src/config/visualseditor.cpp
@@ -8,6 +8,7 @@
 #include "config/uicoloreditor.h"
 #include "utils/confighandler.h"
 
+#include <QCompleter>
 #include <QDirIterator>
 #include <QHBoxLayout>
 #include <QLabel>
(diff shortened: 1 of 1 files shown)


mmahmoudian:
Thanks for the contribution. I have just read the code and am yet to test it, but so far it looks good to me. I'm in favor of this change, although I must add that I don't know how often one would change their application language.

-------

On similar note but unrelated to this PR, I believe we would also need to implement a fuzzy-finder for the Configurations window. Flameshot is very customizable, and finding the relevant setting sometimes can get overwhelming for users who doesn't frequently visit the Configurations window. Perhaps it needs an RFC first as most probably the elements in the Configurations window should be moved around or rearranged.

nicitaacom:
> On similar note but unrelated to this PR, I believe we would also need to implement a fuzzy-finder for the Configurations window. Flameshot is very customizable, and finding the relevant setting sometimes can get overwhelming for users who doesn't frequently visit the Configurations window. Perhaps it needs an RFC first as most probably the elements in the Configurations window should be moved around or rearranged.


Check: https://github.com/flameshot-org/flameshot/pull/4898

ElTh0r0:
I can confirm the huge language combobox on Windows, even though on my Linux machine (Manjaro with KDE) the list is limited to 10 items and scroll-able by default - maybe the operating system style has some influence on this behavior?

I tested the PR on Windows and it works fine and doesn't break anything on my Linux machine. The text filter is "nice to have" and I could have already though about it when I added the language combobox 🙂

PR #4893, opened 2026-08-19
Cap the Configuration window to half the screen height

nicitaacom:
## Fixed
- **Configuration window had no maximum height** — tabs whose content didn't fit inside a `QScrollArea` let the window grow to fit every setting, sometimes exceeding the screen entirely.

Caps the window's maximum height to 50% of the current screen's available height on open, using the same screen-detection helper (`QGuiAppCurrentScreen`) already used elsewhere in the codebase for centering. Each tab's own scroll area handles overflow within that budget.


nicitaacom:
actually this is part of UI change - not reproducabe if no UI changed from other PR

PR #4898, opened 2026-08-21
Group General settings' loose checkboxes into an "Options" section

nicitaacom:
**Description:**

13 checkboxes in the General settings tab were added directly to the tab's scroll area with no visual grouping — interleaved with unrelated input fields (Imgur Client ID, Selection Geometry Display) in an order that didn't reflect any logical grouping, and "Undo limit"/"Latest Uploads Max Size" rendered as two mini boxes hugging their content with dead space beside them while every other section in the tab was a single full-width titled box.

**Changes:**

Moved the 13 checkbox-only settings into a new "Options" titled box, matching the existing "Startup" / "Notifications & Behavior" pattern
Settings whose init function also builds an input field (Save Path, Imgur Client ID, Selection Geometry Display) were left where they are, since nesting an input-bearing box inside a checkbox-only box would look wrong
Merged "Undo limit" and "Latest Uploads Max Size" into one full-width "Limits" box; removed the now-unused pushCompactSpinBox helper that built the old two-mini-box layout
Folded two settings that rendered as bare, borderless rows into the titled boxes they actually belong to: the selection-geometry hide timeout (into "Selection Geometry Display") and JPEG Quality (into "Save Path", since it only affects the save/upload/clipboard file format set there)
Reordered top-level sections so save-destination settings rank above general Options toggles, which rank above rarely-touched tuning knobs (Undo limit, Imgur Client ID, Selection Geometry Display) — "Configuration File" stays last as before

**Test plan:**

Build and open Configuration → General
Confirm the 13 checkboxes render inside one bordered "Options" box
Confirm "Limits" is a single full-width box with both spinboxes
Confirm the geometry timeout and JPEG Quality no longer float as separate rows
Toggle a few settings and confirm they still persist to flameshot.ini

BEFORE
21 08 2026 at 11-38

AFTER 1
image

AFTER 2
image

This is much cleaner because no mess from checkboxes that needs input and just checkboxes and then again checkbox that just buried between 2 input fields
No more suffering when I scroll then scroll blocked because it decide to update some input instead of continuing scrolling

### DIFF
--- a/src/config/generalconf.cpp
+++ b/src/config/generalconf.cpp
@@ -15,7 +15,6 @@
 #include <QMessageBox>
 #include <QPushButton>
 #include <QSettings>
-#include <QSizePolicy>
 #include <QSpinBox>
 #include <QStandardPaths>
 #include <QStringDecoder>
@@ -69,6 +68,11 @@ GeneralConf::GeneralConf(QWidget* parent)
         m_scrollAreaLayout = outer;
         group->activate();
     }
+    // Where screenshots end up is core to every capture, so the save
+    // settings (which also carry JPEG Quality now, folded into the Save
+    // Path box) rank above the general Options toggles below.
+    initSaveAfterCopy();
+    initSaveLocations();
     {
         QVBoxLayout* outer = m_scrollAreaLayout;
         QVBoxLayout* group = pushGroupBox(tr("Options"));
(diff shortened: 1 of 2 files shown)


src/config/generalconf.cpp
```diff
@@ -695,6 +764,93 @@ void GeneralConf::changeSavePath()
     }
 }
 
+// 3 extra save destinations, each exposed as its own tool
```
borgmanJeremy, line 767: This should not be included in this PR which is intended to just be a graphical change.

PR #4899, opened 2026-08-21
Rename Insecure Pixelate to Enable opacity blur, add hover explainer

nicitaacom:
**Description:**

The "Insecure Pixelate" checkbox in General settings restores the old pre-refactor blur/pixelate effect (the current default is a "pseudo-pixelation" that samples only from the selection's edges so it can't be reversed — see the comment above PixelateTool::process). The label and tooltip described this purely in implementation terms ("insecure", "pixelate") without conveying what the setting actually looks like, so it read as an unrelated, faintly alarming toggle rather than a blur-style option.

**Changes:**

Renamed the checkbox to "Enable opacity blur"
Added an (i) info icon next to it; hovering shows a before/after comparison image (data/img/app/blur-diff.png) via Qt's rich-text tooltip support, so the effect is visible before toggling it
No behavior or config-key change — insecurePixelate in flameshot.ini is untouched, this is a UI-only change

**Test plan:**

Build and open Configuration → General
Confirm the checkbox reads "Enable opacity blur"
Hover the (i) icon and confirm the comparison image renders
Toggle it and confirm it still writes insecurePixelate to the config file as before

BEFORE

21 08 2026 at 11-37

AFTER
image


ElTh0r0:
I fully understand your intention for this PR and that the text of the checkbox may not be self-explaining for some users, but "opacity blur" is more confusing in my opinion, because the tools name is "Pixelate" so the current text references the tools name; with the proposed text one doesn't know that it belongs to that tool.

The proposed sample picture is not optimal in my opinion as well, because the "blur gray" caption of the picture is only valid for your black and white sample picture!

Maybe writing a more extended tool tip (without picture) would be better and explaining why it is insecure and what are the optical differences (blurry vs. pixels).

mmahmoudian:
To add my 2cents to this thread, I believe some of the settings' text are a bit unclear and we have tried to elaborate further in the tooltips. But considering that Flameshot is a "visual application" especially when it comes to annotations, some of the visual modifications would be a bit hard to explain purely via text. I personally think adding pictures in an info button is a great idea when it is combined with some clear text.

We have different types of element in the Configurations. For check boxes we should clarify some sort of "before/after" in the picture and in the text so that the user know the effect of enabling/disabling something.

I agree with @ElTh0r0 that the example picture here is suboptimal. Perhaps having some colored pattern like Flameshot logo and use that to show the effect is more informative to the user.

To sum up, I personally like the info button idea and picture example plus adding good and clear explanation in that box. The PR in the current state is suboptimal and the "Enable opacity blur" renaming is in my opinion less clear compared to previous text.

If the idea is agreed by devs and contributors, I can open a discussion in which I propose text, tooltip, and infobox content, and then we can finetune them there. When thatbwas in a satisfactory state, we can update this PR accordingly.

borgmanJeremy:
@mmahmoudian  I generally agree with you but the images need to either not have text, or have the text drawn over the graphic. Otherwise we won't be able to translate.

mmahmoudian:
@borgmanJeremy 
> @mmahmoudian  I generally agree with you but the images need to either not have text, or have the text drawn over the graphic. Otherwise we won't be able to translate. 

Perhaps my comment was vague. What I meant was that when the user hover/clicks on the info button, an overlay info box opens with a picture and some explanation under it in a notmal text (not inside the picture).

I'm not sure even if this explanation is clear 😅 For the sake of efficiency, perhaps I should sketch a mock-up. I'll do it in some hours.

mmahmoudian:
This is a sketch I made with Inkscape. At the top it has one place to show one single picture, and at the bottom it has text. I chose the pixelation because it is one of those options that toggling on and off has visual effect.

image1

I'm not implying that this is the ultimate or best idea, I'm just illustrating what was in my head when I wrote the comment. I'm very curious to know your take and opinion. I think with this type of overlays (whatever the final design might be) we can improve documentation of the configuration page and make it less  confusing.

PR #4902, opened 2026-08-21
Fix clipboard copy on COSMIC Wayland

### LINKED ISSUE #4823: Copy button and Ctrl+C do not copy to clipboard on Pop!_OS COSMIC Wayland (Flatpak)
### Flameshot Version

Flameshot v13.3.0
Compiled with Qt 6.9.3

### Installation Type

Flatpak from Flathub

Pop!_OS 24.04 - COSMIC Desktop - Wayland

### Flameshot Version

Flameshot v13.3.0
Compiled with Qt 6.9.3

### Installation Type

Flatpak (Flathub)

### Operating System

Pop!_OS 24.04
COSMIC Desktop
Wayland

### Description

Flameshot works correctly when launched with

    flatpak run org.flameshot.Flameshot gui

I can capture and annotate screenshots without problems.

However:

- Ctrl+C does nothing
- the toolbar "Copy" button does nothing

The screenshot is NOT copied to the clipboard.

Interestingly, using the context menu (right click -> Copy to clipboard) works correctly.

Also, launching Flameshot with

    flatpak run org.flameshot.Flameshot gui --raw | wl-copy --type image/png

copies the image successfully after pressing Enter.

Therefore the Wayland clipboard itself appears to work. The issue seems to affect only the normal Copy action (toolbar button / Ctrl+C).

### Steps to reproduce

1. Launch:
   flatpak run org.flameshot.Flameshot gui
2. Select an area.
3. Draw an annotation.
4. Press Ctrl+C or click the Copy button.
5. Paste into another application.

### Expected behavior

The screenshot should be copied to the clipboard.

### Actual behavior

Nothing is copied.

Right-click -> Copy to clipboard works.
The `--raw | wl-copy` workaround also works.

### Additional information

Qt 6.9.3
Wayland
COSMIC
Flatpak

### Steps to reproduce

1. Launch:
   flatpak run org.flameshot.Flameshot gui
2. Select an area.
3. Draw an annotation.
4. Press Ctrl+C or click the Copy button.
5. Paste into another application.

_No response_

### System Information

Pop!_OS 24.04
COSMIC Desktop
Wayland

sawy3r:
Closes #4823.

## Problem

The clipboard-persistence workaround (keeping the capture window alive so Wayland can serve clipboard data after the window would otherwise close) is gated to GNOME only:

```cpp
const bool needGnomeWorkaround =
  desktopInfo.waylandDetected() &&
  desktopInfo.windowManager() == DesktopInfo::GNOME;
```

COSMIC hits the exact same underlying Wayland limitation (a client can only serve clipboard data while it's alive to answer paste requests), but never triggers this workaround, so copy-to-clipboard (toolbar button or Ctrl+C) silently does nothing on COSMIC — matching the symptom in #4823 exactly. The right-click "Copy to clipboard" workaround mentioned there happens to work because it doesn't immediately close the window afterward, incidentally keeping the process alive long enough.

## Fix

- Extend the check to also cover `DesktopInfo::COSMIC`.
- Fix a race in the workaround itself, uncovered while testing this on COSMIC: it previously closed the window on the *first* clipboard read of any kind. If a paste consumer probes available clipboard types before issuing the real data fetch, that first read would close the window before the real fetch arrives. It now closes only once the clipboard's current data is no longer ours (compares against the live `QClipboard::mimeData()` pointer rather than assuming the first read is the real one).
- Raise the safety-net timeout from 500ms to 30s. GNOME's mutter grabs clipboard offers eagerly on copy, so 500ms was enough there, but COSMIC's compositor fetches lazily — only once the user actually pastes — which can reasonably take longer than half a second.

## Testing

Verified on Pop!_OS 24.04 with COSMIC 1.0.0 (`cosmic-comp 1.0.0`), built from current `master`: capture → copy → paste now works reliably across multiple runs, into both GUI apps and `wl-paste`-based tooling.

### DIFF
--- a/src/utils/screenshotsaver.cpp
+++ b/src/utils/screenshotsaver.cpp
@@ -238,11 +238,6 @@ private:
             return;
         m_notified = true;
         AbstractLogger::info() << QObject::tr("Capture saved to clipboard.");
-        QPointer<QWidget> guard = m_owner;
-        QTimer::singleShot(0, [guard]() {
-            if (guard)
-                guard->close();
-        });
     }
 
     QImage m_image;
@@ -258,10 +253,31 @@ bool saveToClipboardGnomeWorkaround(const QPixmap& pixmap, QWidget* keepAlive)
 
     keepAlive->hide();
 
-    // Safety net: force close after 500ms if compositor never fetches
-    QTimer::singleShot(500, keepAlive, [keepAlive]() {
-        qWarning() << "GNOME workaround timed out, compositor did not request "
-                      "clipboard data within 500ms. Force closing.";
+    // Close once something else takes clipboard ownership, rather than
+    // after the first read: some paste consumers probe the clipboard
+    // (e.g. checking available types) before issuing the real fetch, and
+    // closing on that first read would cut them off before it arrives.
+    // dataChanged also fires for our own setMimeData() call above (the
+    // Wayland clipboard claim is confirmed asynchronously by the
+    // compositor), so we only close once the clipboard's current data is
+    // no longer ours, rather than closing on the first emission.
+    QObject::connect(clipboard,
+                     &QClipboard::dataChanged,
+                     keepAlive,
+                     [clipboard, mimeData, keepAlive]() {
+                         if (clipboard->mimeData() == mimeData)
+                             return;
+                         if (keepAlive)
+                             keepAlive->close();
+                     });
+
+    // Safety net: force close if nothing ever fetches the data. GNOME's
+    // mutter grabs clipboard offers eagerly on copy, so 500ms was enough
+    // there, but compositors that fetch lazily (e.g. COSMIC) only request
+    // the data once the user actually pastes, which can take much longer.
+    QTimer::singleShot(30000, keepAlive, [keepAlive]() {
+        qWarning() << "Clipboard workaround timed out, compositor did not "
+                      "request clipboard data within 30s. Force closing.";
         if (keepAlive)
             keepAlive->close();
     });
--- a/src/widgets/capture/capturewidget.cpp
+++ b/src/widgets/capture/capturewidget.cpp
@@ -625,10 +625,10 @@ void CaptureWidget::uncheckActiveTool()
 void CaptureWidget::closeEvent(QCloseEvent* event)
 {
 #if !(defined(Q_OS_MACOS) || defined(Q_OS_WIN))
-    /* GNOME copy problem workaround, copy
+    /* Wayland copy problem workaround: the copy
        operation seems to work only when there
        is a visible window to retrieve the
-       data from. On GNOME, the GUI should
+       data from. On GNOME and COSMIC, the GUI should
        handle the copy operation, not the
        daemon.
     */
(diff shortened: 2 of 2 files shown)


cleytonmendest (commented):
Confirmed on COSMIC — this fixes it for me.

Pop!_OS 24.04, `cosmic-comp 0.1`, Wayland, source builds of `master` (5e5c341) and this PR (e6ae33a), Qt 6.4.2, default CMake options. Same procedure on both: `flameshot gui`, select with the mouse, Ctrl+C, then poll `wl-paste` once a second. On `master`, `image/png` never shows up. On this PR it appears ~4s in and stays available for the whole window — 464 KB, `PNG image data, 1029 x 614`.

(Tip for anyone reproducing: `--accept-on-select` fails on both builds — auto-accept produces no input event, and Wayland's `set_selection` needs a serial from a real one. Has to be an actual Ctrl+C.)

One thing worth a look: **the 30s timeout is the normal exit path, not a fallback.** Every successful run ended with `Clipboard workaround timed out...`. The new `dataChanged` close never fires, because reading the selection doesn't change ownership — `clipboard->mimeData() == mimeData` stays true unless something else *claims* the clipboard, which doesn't happen in a normal copy-then-paste. So every copy holds the process for the full 30s, and during that time a second `flameshot gui` is rejected by the single-instance guard in `main.cpp` with exit code 1 and no message. Releasing the mutex before the keep-alive phase would avoid that.

Also, since #4823 is against the Flatpak: that build sets `-DUSE_WAYLAND_CLIPBOARD=1` (→ `KSystemClipboard`), but `saveToClipboardGnomeWorkaround()` uses `QGuiApplication::clipboard()` unconditionally. I only tested with the default (OFF), so that path is still unverified.

Neither is a blocker from my side — big improvement over copy doing nothing at all.

borgmanJeremy:
Why is this needed on Cosmic and Gnome but not other wayland DE's like KDE or Hyprland?

sawy3r:
Good question — worth splitting into what I can confirm vs. what I can't:

- **GNOME**: eager-fetch. Mutter grabs clipboard offers immediately at copy time, which is exactly why the original GNOME workaround only needed a 500ms grace window before this PR.
- **COSMIC**: lazy-fetch, confirmed by direct testing. The compositor only requests clipboard data once an actual paste happens, which is why the window needs to stay alive much longer (see #4917 for a related fix to a side effect of that).
- **KDE**: doesn't need this specific workaround, but not because its plain Wayland clipboard behaves like GNOME's — it has its own dedicated solution already in this codebase (`USE_WAYLAND_CLIPBOARD` / `KSystemClipboard`), which solves the same underlying "process must stay alive to serve clipboard data" problem through KDE's own clipboard infrastructure instead of this GUI-window trick.
- **Hyprland/Sway (wlroots)**: I haven't verified their exact fetch timing myself, so I don't want to assert a mechanism I can't back up. What I can say: the docs (`docs/UsageHyprlandSwayWlroots.md`) describe Flameshot working "out of the box" there with no clipboard caveat, and there's no equivalent bug report for those DEs in the tracker — so either they don't hit this particular timing issue, or nobody's run into it yet. If someone with a Hyprland/Sway setup wants to check, the way to tell is: `flameshot gui`, select, Ctrl+C, then poll `wl-paste --type image/png` — if it succeeds within ~100ms with the app still running, that compositor is eager-fetching like GNOME; if it only succeeds once you actually paste somewhere, it's lazy-fetching like COSMIC and would need the same fix.

borgmanJeremy:
the USE_WAYLAND_CLIPBOARD flag makes cosmic work fine without this change. I think I'm going to roll this back. I need to repeat testing on gnome to see if I can remove this hack there too.

PR #4905, opened 2026-08-23
feat(plugins): introduce modular Qt plugin architecture and plugins manager UI

SniperRavan:
## Summary

This pull request introduces a modular, lightweight C++ dynamic plugin system built around Qt's `QPluginLoader`.

The system allows third-party developers and users to build, install, and load custom capture tools, post-processing utilities, and community integrations—such as OCR, AI vision, QR code scanners, custom uploaders, and image filters—without adding third-party dependencies to the core Flameshot codebase.

---

### 1. Plugin Interface SDK (`FlameshotPluginInterface`)

* Introduces a clean, abstract plugin interface.
* Provides plugin metadata including:

  * `id`
  * `name`
  * `version`
  * `author`
  * `description`
  * `icon`
* Exposes a `createTools()` factory method for creating plugin-provided tools.

### 2. Plugin Manager (`PluginManager`)

* Automatically discovers and loads plugins with platform-specific extensions:

  * `.so` on Linux
  * `.dll` on Windows
  * `.dylib` on macOS
* Searches for plugins in:

  * User directory: `~/.local/share/flameshot/plugins`
  * Windows user directory: `%APPDATA%/flameshot/plugins`
  * System directory: `/usr/lib/flameshot/plugins`
  * Paths specified through the `FLAMESHOT_PLUGIN_PATH` environment variable
* Persists plugin enable/disable state through Flameshot's configuration system.

### 3. Capture Toolbar Integration

* Tools provided by enabled plugins are dynamically added to the capture toolbar.
* Plugin tools work alongside Flameshot's built-in tools without requiring changes to the core tool implementation.

### 4. Plugin Configuration UI

* Adds a dedicated **Plugins** tab to Flameshot Configuration (`flameshot config`).
* Displays installed plugins and their metadata.
* Provides enable/disable switches for individual plugins.
* Includes a one-click button to open the plugins directory.

### 5. Example Plugin & Documentation

* Adds a reference implementation in `examples/sample-plugin/`.
* Adds a developer guide at `docs/PLUGINS.md`.
* Documents the plugin interface, discovery paths, installation, and development workflow.

---

## Benefits

* Keeps the Flameshot core lightweight and dependency-free.
* Enables community-driven integrations without modifying the core codebase.
* Makes it easier to experiment with new capture and post-processing functionality.
* Provides a clear and stable foundation for third-party plugin development.


mmahmoudian:
Thanks for the PR. I just wonder if you have incorporated the ideas fro  this RFC into the design:

SniperRavan:
> Thanks for the PR. I just wonder if you have incorporated the ideas fro this RFC into the design:
> 
> #2529

Thanks for reviewing! Yes, this PR is inspired by the ideas and maintainer discussion in
  [#2529](https://github.com/flameshot-org/flameshot/discussions/2529).

docs/PLUGINS.md
```diff
@@ -0,0 +1,108 @@
...
+## 3. Creating a Custom Plugin
+
+### A. Define the Plugin Class
+Implement `FlameshotPluginInterface` and provide custom `CaptureTool` instances:
+
+```cpp
+#include "plugins/flameshotplugininterface.h"
+#include "tools/abstractactiontool.h"
+
+class MyCustomTool : public AbstractActionTool
```
jack9603301: Please don't use such a confusing name.

docs/PLUGINS.md
```diff
@@ -0,0 +1,108 @@
...
+    }
+    CaptureTool::Type type() const override { return CaptureTool::NONE; }
+    CaptureTool* copy(QObject* parent = nullptr) override { return new MyCustomTool(parent); }
+    void process(QPainter& painter, const QPixmap& pixmap) override {
+        // Custom drawing or image processing
+    }
+    bool closeOnButtonPressed() const override { return false; }
+};
+
+class MyPlugin : public QObject, public FlameshotPluginInterface
```
jack9603301: Please don't use names that are clearly generated by AI and confusing.

jack9603301:
I didn't go through it all, I went through it roughly, I'll go through it later today, and I left some initial feedback.

jack9603301:
@SniperRavan Did you personally compile and submit this PR?

I am not opposed to using AI, but I do not want you to rely on it entirely.

src/widgets/capture/capturewidget.cpp
```diff
@@ -407,6 +408,39 @@ void CaptureWidget::initButtons()
             vectorButtons << b;
         }
     }
+
+    // Add dynamic buttons from active plugins
```
jack9603301, line 412: It appears that the tool interface is being called directly here; in other words, your plugin interface does not support pipeline or, uh, piping workflows—is that correct?

Also, we have considered supporting other languages, such as Python and Lua; how do you plan to handle this?

jack9603301:
Actually, we already have a draft: https://github.com/flameshot-org/flameshot/discussions/3953

I’m not saying your idea isn't good, but personally, I think it would be fantastic if we could support interoperability and piping.

SniperRavan:
Hi @jack9603301, thank you for the review and for linking RFC #3953!

  On the naming in docs/PLUGINS.md — you're completely right. I left AI-generated placeholder names which I missed when reviewing. I've pushed an update that replaces them with a concrete `/HighlightPlugin` example using a proper reverse-domain ID (`org.flameshot.highlight`), real method implementations, a build-and-walkthrough, a CLI management section, and a conventions table.
  I also removed a hardcoded source-tree dev path that had slipped into `::pluginSearchPaths()`.

  On whether I compiled and tested this — yes, the entire change set (`core binary`, `OCR plugin`, `QR plugin`, and `sample plugin`) builds cleanly on my Linux with `GCC 13` / `Qt6`. The `flameshot --list-plugins` CLI command, the Plugins config tab, enable/disable toggles, and install/remove flow are manually tested. Though as a newcomer I may have missed things, so further feedback is welcome.

  I use AI tools for drafting and initial code generation, but I review, compile, and test everything locally and take full responsibility for the implementation.

  ---

 About the piping workflow & other languages (RFC #3953):
- The current PR implements what I'd call Tier 1: Interactive GUI Tool Plugins — tools that need direct canvas interaction within the capture window, such as drawing a region to extract text (OCR) or scanning a captured area for QR codes. These need to live in-process and `CaptureTool` is the right interface for them.

- RFC #3953 describes a Tier 2: Post-Capture Pipeline Plugins —` watermarking`, `compression`, upload to `Imgur`/`XBackBone`,` webhooks`, `OCR-to-file` export. I agree these should not go through `CaptureTool`. The right model for them is a `QProcess` subprocess runner:

  1. Flameshot writes the captured PNG to a temp file (or passes it on stdin).
  3. Spawns the plugin script asynchronously: `python watermark.py`, `./uploader`, `lua hook.lua`.
  4. Reads the result (modified image path or JSON `{"url": "..."})` from stdout on exit 0.
  5. Core handles `Wayland`/`X11` clipboard write and desktop notifications natively — plugin scripts don't need to touch the clipboard themselves.

  For this PR I'd suggest keeping the scope to Tier 1. I can follow up with a separate PR for the Tier 2 subprocess pipeline runner once this lands and we've aligned on the directory structure from RFC #3953.

While checking I also found and fixing a few bugs while testing and going to push them— `hardcoded ZBar` library path in `CMakeLists` (won't build on other distros), a Tesseract API memory leak when tessdata init fails, the UI freezing during OCR scan (added wait cursor), the Pin button being a stub, and both  result windows always centering on the primary monitor regardless of which screen the capture was on.

jack9603301:
I've had a brief look at your implementation, and I don't think there are any other problems with the first level of implementation. We can experimentally incorporate some minimal features, and then, as you said, gradually refine it.

jack9603301:
Thank you very much for your willingness to implement the plugin RFC

SniperRavan:
Thanks

mmahmoudian:
Thank you @SniperRavan and @jack for the fruitful discussion and being active.

I believe it is a good idea to get feedback from other devs and see if we can push this forward considering the good condition of this PR. I'll tag some folks here, but feel free to tag more.

@borgmanJeremy @panpuchkov @veracioux @hosiet @holazt @ElTh0r0

Issue #4851, opened 2026-07-28
[Bug] Regression: Double Scaling / UI & Pin offset on Windows 11 returns in v14.0.0 (reoccurrence of #4610)


silfsy:
### Flameshot Version

[Flameshot v14.0.0 (3207dfeb)](https://github.com/flameshot-org/flameshot/releases/tag/v14.0.0)

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

Windows 11 Pro, version 24H2

### Description

This is a regression report for issue #4610, which was previously fixed by PR #4614.

With the release of Flameshot [v14.0.0](https://github.com/flameshot-org/flameshot/releases/tag/v14.0.0), the fractional scaling conflict on Windows 11 (e.g., 150% scaling) has returned:

1. Qt internal scaling conflicts with Windows DWM scaling, resulting in oversized UI elements.
2. Coordinate math for the **Pin** tool is broken—pinned screenshots shift away from the selected capture area and render significantly enlarged.

Since PR #4614 resolved this during testing, it appears the fix may have regressed or was not carried over into the final v14.0.0 release.

### Steps to reproduce

1. Set Windows 11 display scaling to any fractional value above 100% (e.g., 150%).
2. Open Flameshot v14.0.0.
3. Capture any region of the screen.
4. Click the "Pin" tool.
5. Observe that the pinned snapshot shifts position and scales up incorrectly.

### Screenshots or screen recordings

Refer to the original recording in #4610 (the behavior in v14.0.0 is identical).

### System Information

1. Operating System: Windows 11 Pro, version 24H2
2. Display Configuration: 150% Scaling
3. Flameshot Version: v14.0.0

silfsy:
**Additional finding regarding activation method:**

I noticed the issue seems to depend on how the capture process is invoked:

* **Triggered via System Tray Icon:** Clicking the Flameshot tray icon to capture and then using the **Pin** tool works as expected (pinned image is normally sized and positioned correctly).
* **Triggered via Global Shortcut (e.g., `PrintScreen`):** Activating the capture using a keyboard shortcut and then using the **Pin** tool triggers the bug (pinned image is shifted and scaled incorrectly).

It looks like the DPI scaling context might be evaluated differently depending on whether Flameshot is invoked through the UI tray context versus the global keyboard hook.

eugenesvk:
[...]

PR #4907, opened 2026-08-24
Fix pin size on Windows for scaled screen

4RH1T3CT0R7:
Pinning a capture on a fractionally scaled display gives a window about 1.5x the region you selected. #4614 fixed the position of that window; this is the size half of the same report.

A capture from the tray runs inside the daemon, so the pixmap is handed straight over. Anything in its own process (`flameshot gui`, a shortcut, D-Bus) serializes it, and `QDataStream` drops the device pixel ratio, so the daemon rebuilds the pixmap at ratio 1. Same capture, right from the tray, wrong from a shortcut.

The fix sends the ratio alongside the pixmap and restores it on read. On Windows 11 at `QT_SCALE_FACTOR=1.5`, a 400x300 region pinned as 621x471 before and 422x321 after. Mixed-DPI pin positioning is still broken. Refs #4851.

@silfsy, if you can build this branch, could you confirm the pin looks right for you?

### DIFF
--- a/data/dbus/org.flameshot.Flameshot.xml
+++ b/data/dbus/org.flameshot.Flameshot.xml
@@ -11,7 +11,8 @@
 
     <!--
         attachPin:
-        @data: Byte array containing the screenshot and geometry information.
+        @data: Byte array containing the screenshot, geometry information and
+        the screenshot's device pixel ratio.
 
         Attach a pinned screenshot widget to the daemon.
     -->
--- a/src/core/flameshotdaemon.cpp
+++ b/src/core/flameshotdaemon.cpp
@@ -39,6 +39,26 @@
 #include "core/globalshortcutfilter.h"
 #endif
 
+namespace {
+/**
+ * @brief Read a pin message written by FlameshotDaemon::createPin.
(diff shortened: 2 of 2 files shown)


mmahmoudian:
@ElTh0r0 I'm interested in having your opinion on this PR as well.

ElTh0r0:
Good catch!
I can confirm the issue and that this PR fixes it! Tested on Windows 11 and Manjaro (KDE with Wayland). Honestly I never tested pinning from cli together with scaling - there are just too many possible combinations to be tested that some are slipping through.
Thanks for the PR - green light from me 😉

mmahmoudian:
@4RH1T3CT0R7 Thanks for the PR and the explanation. The code looks good to me. Let's wait for @borgmanJeremy opinion as well. In the meantime and considering the diff size, I approved the code-change so that the CIs get run.

@ElTh0r0 Thank you for trying it out. Your insights are always valuable. Feel free to review the PRs if you see any that you find interesting :)

borgmanJeremy:
Thank you for the contribution!

PR #4917, opened 2026-08-29
Release gui mutex before clipboard keep-alive wait

sawy3r:
Follow-up to #4902, addressing a bug found in review by @cleytonmendest.

## Problem

The COSMIC/GNOME clipboard workaround keeps the capture window alive (hidden) for up to 30s while waiting for a paste consumer to fetch the clipboard data. As pointed out in [review on #4902](https://github.com/flameshot-org/flameshot/pull/4902#pullrequestreview-5003060055), that 30s wait is actually the *normal* path, not a rare fallback — the `dataChanged`-based early close only fires when something else claims clipboard ownership, which doesn't happen in an ordinary copy-then-paste. So every copy holds the process alive for the full 30 seconds.

The problem: that process still holds `flameshot gui`'s single-instance lock (`QSharedMemory`-based, `guiMutexLock()` in `main.cpp`) for the entire wait. A second `flameshot gui` invocation during that window fails to acquire the lock and exits immediately with no visible UI or error — reported independently as "Ctrl+Shift+S sometimes just does nothing" after a recent capture.

## Fix

Moves `guiMutexLock()` out of `main.cpp` into its own `utils/guimutex.{h,cpp}`, and adds `releasePendingGuiMutex()` — an idempotent release, callable both from the normal `aboutToQuit` cleanup and, now, from the clipboard workaround itself. The mutex is released as soon as we hand off to the keep-alive wait: the actual GUI portion of the capture is already done at that point (selection made, copy requested), so a new capture shouldn't be blocked by a previous one that's only lingering in the background to serve a pending clipboard read.

## Testing

Verified on Pop!_OS 24.04 with COSMIC 1.0.0, built on current `master`: rapid back-to-back `Ctrl+Shift+S` captures now both work immediately. Previously the second one silently did nothing until the first's 30s timeout (or an incidental clipboard-ownership change elsewhere) released the lock.

Doesn't address the other point raised in review (the `USE_WAYLAND_CLIPBOARD`/`KSystemClipboard` build path also going through `saveToClipboardGnomeWorkaround()`, which uses `QGuiApplication::clipboard()` unconditionally) — I don't have a way to build/test against `KF6GuiAddons` here, so left that as a known follow-up rather than guessing at a fix I can't verify.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -18,6 +18,7 @@
 #include "utils/abstractlogger.h"
 #include "utils/confighandler.h"
 #include "utils/filenamehandler.h"
+#include "utils/guimutex.h"
 #include "utils/pathinfo.h"
 #include "utils/valuehandler.h"
 
@@ -31,7 +32,6 @@
 #include <QDir>
 #include <QLibraryInfo>
 #include <QNetworkProxyFactory>
-#include <QSharedMemory>
 #include <QTimer>
 #include <QTranslator>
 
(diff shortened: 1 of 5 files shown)


borgmanJeremy:
Thank you but this is not needed on gnome. I've reverted the patch that allows cosmic to use this workaround because it is not needed on Cosmic if flameshot is compiled with the USE_WAYLAND_CLIPBOARD feature flag.

sawy3r:
Thanks @borgmanJeremy that's a much cleaner solution!

Issue #4910, opened 2026-08-25
Feature proposal: ShareX-style hover snapping to windows and UI elements before selecting (working macOS patch)


nuuxcode:
## The feature

In ShareX on Windows, when you start a capture and move the mouse before clicking, the prospective selection automatically snaps to what is under the cursor: a whole app window, or a sub-region of it (a browser's page area, a toolbar, a panel). One click accepts it; dragging still makes a free selection.

## Working implementation (macOS)

I have this running as a patch against v14.0.0:

- While no selection exists, `CGWindowListCopyWindowInfo` finds the frontmost window under the cursor (skipping Flameshot's own overlay, the menu bar and the Dock) and the selection follows it live.
- Inside that window, `AXUIElementCopyElementAtPosition` scoped to the owning app's Accessibility element refines the rect to the UI element under the cursor, climbing to the parent while the element is smaller than a usable region (120x80). Falls back to the whole window when Accessibility permission is missing.
- A press ends hover mode: plain click accepts the snapped region, drag creates a normal free selection.

Patch + prebuilt build: https://github.com/nuuxcode/flameshot-inside-toolbar (`hover-snap-window.patch`)

The same design should map to Windows (EnumWindows/UIA) and Linux (varies by compositor), which is why I'm filing it as a proposal rather than a PR. Happy to contribute the macOS side as a PR behind a config option if there's interest.

Related: #4909 (toolbar inside the selection, same patch repo).

mmahmoudian:
Seems like duplicate of #5.

Does similar proposed solution also work on Linux (Wayland and X11)?
I'm personally not in favor of platform-specific solutions. The feature request of #5 and all it's duplicates are fine, but if the proposed solution is discriminatory, I don't believe it is a viable solution.

Issue #4239, opened 2025-09-13
Enhanced window/panel selector


a-b:
### Feature Description

How about adding a selector mode that works like the Mac's native screenshot tool when you press the `Space` key? It lets you select the entire UI panel or window element. 
It'd be great if Flameshot had a similar feature, so I could use a modifier key to keep the selection stuck to the perimeter of the panel, window, or dialog under my cursor, rather than just manually selecting the area to match the GUI element.

For instance, in the screenshot below, we can see that the panel is highlighted in light blue because I pressed the `Space` key while selecting it using the native Mac screenshot tool using a key combination: Ctrl-Command-Shift-4  and space

<img width="689" height="613" alt="Image" src="https://github.com/user-attachments/assets/1eb7a876-0fc1-4d6e-8055-6062a007fd2e" />

borgmanJeremy:
It's a great idea but unfortunately can't be implemented across all platforms. Each platform has its own API for that, and on Wayland applications cannot get that kind of information through a standard API.

a-b:
What do you consider feasible in terms of building or adopting a lightweight image recognition algorithm via a CV library that can identify the closest rectangles and stick to them?

borgmanJeremy:
Someone attempted this before but there are too many random lines to make it effective

a-b:
Thank you for the context. 
I'm now wondering if there have been any advancements in library since the last attempt. 
I'm also curious to explore the possibility of offsetting deficit of algorithm precision with product requirements. For example, instead of strict rectangle detection, we could still rely on user guidance as a main driver. When the modifier key is pressed, the cursor could jump to the next line or corner to fast forward users movement.

PR #4918, opened 2026-08-29
RFC: Add smart region selection

feketegabor:
## Summary

Propose an optional smart-region workflow for GUI capture:

- hover to preview the window or UI element under the pointer;
- click to commit that region into the normal editable Flameshot selection;
- drag to keep the existing manual rectangle workflow;
- scroll before selection to move between child and parent candidates;
- retain normal tool-size scrolling after a selection is committed.

This pull request contains only the RFC, following the repository's RFC-first process for substantial UI features.

## Platform strategy

The RFC defines one platform-neutral detector contract owned by the capture UI. It proposes native providers for macOS, Windows, and X11, while leaving unsupported backends as strict no-ops. It also calls out the current Wayland limitation explicitly: a backend should remain disabled until compositor APIs can provide reliable foreign window and control geometry.

The first working reference implementation is macOS-only and opt-in. I would particularly value maintainer feedback on whether staged native backends are acceptable, or whether multiple backends should be required in the first implementation pull request.

## Prior work and proof of concept

Issue #4910 and its linked patch demonstrate the feasibility of macOS Window Server and Accessibility hover hit-testing. This RFC credits that work and extends the design with a shared backend boundary, leaf-to-window navigation, throttling, explicit permission handling, and unchanged behavior unless the user opts in.

A tested proof-of-concept implementation is available on the fork's [`feature/macos-smart-region-selection`](https://github.com/feketegabor/flameshot/tree/feature/macos-smart-region-selection) branch. I will not open the implementation pull request unless and until the RFC is accepted, as requested by `docs/RFC.md`.

Related to #5, #4239, and #4910.

## Validation

- RFC is the only file changed in this pull request.
- Markdown diff passes `git diff --check`.
- The separate proof-of-concept branch builds on macOS arm64, passes targeted click-vs-drag and detector smoke tests, and has a valid ad-hoc app-bundle signature.

### DIFF
--- a/docs/RFC/0000-smart-region-selection.md
+++ b/docs/RFC/0000-smart-region-selection.md
@@ -1,0 +1,154 @@
+# Smart region selection
+
+### To be Reviewed By
+
+* Flameshot maintainers
+* macOS maintainers and users
+* Contributors interested in equivalent Windows and Linux backends
+
+### Authors
+
+* Gábor Fekete and contributors
+
+### Status: Draft
+
+### Superseded by
+
+N/A
+
+### Related
+
+* Issue #5
+* Issue #1814
+* Issue #2775
+
+## Problem
+
+Flameshot's GUI capture mode starts with no selected area. Capturing a whole
+window, dialog, toolbar, or other rectangular interface element therefore
+requires carefully drawing its bounds by hand. This is slower and less precise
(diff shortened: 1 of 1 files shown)


feketegabor:
Closing this RFC for now. We will continue the cross-platform design and implementation work in the fork before asking upstream to review the proposal.

PR #4932, opened 2026-09-03
Fall back to a same-language translation for unmatched locales

### LINKED ISSUE #4921: Italian localization doesn't apply to it_CH
### Flameshot Version

Flameshot v14.0.0 ()
Compiled with Qt 6.11.1

### Installation Type

Arch Linux

### Description

Flameshot defaults to English when the locale is `it_CH.UTF-8`, as the localization only exists for `it_IT`

### Steps to reproduce

1. run `flameshot config`
2. make usre Lanauge is set to "auto"
3. run `LANG="it_CH.UTF-8" flameshot`

Notice the output `"No Flameshot translation found for it-Latn-CH, it-CH, it-Latn, it"` (twice for some reason)

### Proposed Fix
Rename the locale from `it_IT` (which is supported) to just `it`. There is no linguistic difference between the two. This would make the fallback `it` hit

4RH1T3CT0R7:
With the UI language set to auto, configureTranslation in src/main.cpp relies on QTranslator's locale lookup. For it_CH that tries Internationalization_it_CH and Internationalization_it, but the Italian file is shipped as it_IT, so Swiss Italian users get English and the "No Flameshot translation found" warning. de_AT and de_CH miss de_DE the same way, and sv_FI misses sv_SE.

When the lookup fails, the loader now looks for any Internationalization_<lang>_*.qm in the same translations directory and loads the first one by name (so a language with several regional files, like zh, falls back to the alphabetically first), leaving the .ts file names alone since Weblate manages them. Compiled on Windows with Qt 6.10.1 and checked with a small program against the built .qm files: it_CH picks it_IT, de_AT picks de_DE, nl_BE still gets nl, an unknown language still gets nothing.

Fixes #4921

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -121,6 +121,19 @@ void configureTranslation(QTranslator& translator, QTranslator& qtTranslator)
                               QStringLiteral("Internationalization"),
                               QStringLiteral("_"),
                               path);
+            if (!foundTranslation) {
+                // Neither the exact locale nor the bare language is shipped
+                // (it_CH with only it_IT available): take any translation
+                // of the same language rather than falling back to English.
+                QString language = QLocale().name().section('_', 0, 0);
+                QStringList sameLanguage = QDir(path).entryList(
+                  { QStringLiteral("Internationalization_%1_*.qm")
+                      .arg(language) },
+                  QDir::Files);
+                foundTranslation =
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Thank you!

PR #4935, opened 2026-09-06
build(windows): enable Control Flow Guard for MSVC builds

### LINKED ISSUE #4934: Windows release binary ships without Control Flow Guard (/guard:cf)
### Feature Description

The Windows release binary ships without Control Flow Guard. Checked against the published `Flameshot-14.0.0-win64.msi`: the extracted `flameshot.exe` has `DYNAMICBASE`, `HIGHENTROPYVA` and `NXCOMPAT` set and is Authenticode-signed, but the `IMAGE_DLLCHARACTERISTICS_GUARD_CF` flag is absent - reproducible with `winchecksec` or `dumpbin /headers /loadconfig` on the extracted binary.

For an MSVC build this is a compiler-and-linker flag pair (`/guard:cf`), typically gated on MSVC in CMake so the MinGW/Linux/macOS builds are untouched. Qt desktop apps ship with it routinely - KeePassXC 2.7.12, also Qt on MSVC, has it enabled alongside the same mitigation set Flameshot already carries - so the compatibility ground is well trodden.

Since everything else in the hardening set is already present, this looks like the one missing piece rather than a policy choice - but is there a known reason `/guard:cf` was left out, or would a CMake change enabling it for MSVC builds be accepted?

Dashtid:
Fixes #4934.

Flameshot's Windows build already turns on ASLR, high-entropy ASLR and DEP. Control Flow Guard is the one Windows mitigation it is missing. CFG makes a memory-corruption bug -- in an image decoder, in Qt, anywhere in the process -- much harder to turn into code execution, because an indirect call through a corrupted pointer is stopped before it lands. Windows' own DLLs are compiled with it, but it only switches on for a process whose main executable is marked, so Flameshot gets none of it today.

The change is two MSVC flags in the top-level CMakeLists.txt: `/guard:cf` when compiling and `/GUARD:CF` when linking. Both are needed. They sit above the `add_subdirectory()` calls so the vendored Qt-Color-Widgets and KDSingleApplication builds pick them up too.

I could not build Flameshot itself here (no Qt 6.9), so I checked the construct on a minimal project with the same layout: with the flags the linked exe shows `Control Flow Guard` in `dumpbin /headers`, without them it does not. The AppVeyor MSVC build of this branch passed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,21 @@ endif()
 #Needed due to linker error with QtColorWidget
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+# Control Flow Guard, an MSVC exploit mitigation that validates the target of
+# every indirect call at runtime. Set here, above the add_subdirectory() calls
+# below, so that it also covers the vendored dependencies that get compiled
+# into the executables.
+#
+# Both flags are needed and they do different jobs: /guard:cf instruments the
+# code, and /GUARD:CF marks the image. The mark is the part that matters most,
+# because CFG is a per-process policy read from the main image at process
+# creation -- an unmarked flameshot.exe leaves CFG off for the whole process,
+# including the checks already compiled into the Windows DLLs it loads. Setting
+# only the compile flag would pay the cost and gain nothing.
+if(MSVC)
+  add_compile_options(/guard:cf)
+  add_link_options(/GUARD:CF)
+endif()
 
 # Dependency can be fetched via flatpak builder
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Nothing in this very long, AI generated PR explains _why_ we want this. Please rework this to be actually human readable.

Then adjust the comments in the code to not be an entire paragraph.

Dashtid:
Reworked: the description now opens with what CFG buys Flameshot and is a third of the length, and the CMake comment is two lines. The change itself is unchanged -- two flags, top level, `if(MSVC)` -- rebased onto current master.

GitHub would not let me reopen this one after the rebase, so it continues in #4967.

PR #3166, opened 2023-04-15
Use pixmap's width and desktop's width to calculate DPR on Wayland

easyteacher:
Without fractional scaling support in Qt5 and without a surface in Qt6, getting the real scale factor will require using Wayland protocol. However using Wayland protocol will add additional dependencies in flameshot. This fix uses pixmap's width and desktop's width to calculate the real DPR on Wayland.

Depends on https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/186

Fixes #3164

PR #4941, opened 2026-09-08
Add KWin ScreenShot2 fast path for KDE Wayland

uzqw:
### Description

On KDE Plasma Wayland, full-desktop capture goes through the
freedesktop screenshot portal. The KDE backend encodes the frame as
PNG before returning a file URI, which dominates capture time on
large or high-DPI screens (~2.8s measured on a 4K Plasma Wayland
setup, `flameshot gui` until the overlay appears).

This adds a fast path through KWin's
`org.kde.KWin.ScreenShot2.CaptureWorkspace`: KWin writes raw pixels
to a pipe over D-Bus, so there is no PNG round-trip (~260ms on the
same screen). The request uses `native-resolution:true`.

The portal stays the fallback. KWin is tried only on Wayland and only
if `org.kde.KWin.ScreenShot2` is on the session bus. Any failure
(service missing, rejected request, timeout, short read) falls back
to the portal. GNOME, wlroots, X11, Flatpak, and Snap are unchanged.

Related to #2495 and #3165.

- #2495 moved KDE to the freedesktop portal. In that review Jeremy
  asked about using ScreenShot2; the PR took the simpler portal-only
  route. This adds that path as a fast lane, with the portal as
  fallback.
- #3165 used the same interface for the half-resolution (DPR) bug and
  was dropped in favor of the smaller #3166. This PR is about capture
  latency from the portal PNG encode, which that DPR fix does not
  address.

### Notes for self-built installs

KWin only authorizes callers whose executable (`/proc/PID/exe`)
matches the `Exec=` of a `.desktop` file that declares
`X-KDE-DBUS-Restricted-Interfaces=org.kde.KWin.ScreenShot2`.

The packaged `.desktop` already has that key. Distro packages are
fine. If you build with a custom prefix or run the binary out of
`build/`, either install so `Exec=` matches, or leave it: KWin
denies the call and we fall back to the portal. Capture still works,
just without the speedup.

### Testing

- KDE Plasma Wayland, 4K: ~2.8s (portal) → ~260ms (KWin)
- Service absent / `.desktop` mismatch: falls back to portal

### DIFF
--- a/src/utils/screengrabber.cpp
+++ b/src/utils/screengrabber.cpp
@@ -31,9 +31,14 @@
 #include <QDBusInterface>
 #include <QDBusMessage>
 #include <QDBusReply>
+#include <QDBusUnixFileDescriptor>
 #include <QDir>
 #include <QUrl>
 #include <QUuid>
+#include <cerrno>
+#include <fcntl.h>
+#include <poll.h>
+#include <unistd.h>
 #endif
 
 bool ScreenGrabber::m_monitorSelectionActive = false;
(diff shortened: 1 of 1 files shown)


borgmanJeremy:
Thank you but I've intentionally unified all of the Wayland backend to the portal implementation. If the slowdown is in the KWIN portal or compositor, a fix should be looked at there.

uzqw:
> Thank you but I've intentionally unified all of the Wayland backend to the portal implementation. If the slowdown is in the KWIN portal or compositor, a fix should be looked at there. 

Ack. Thanks for the clarification.

PR #4944, opened 2026-09-10
fix: reduce Wayland capture and clipboard latency

Aziffo:
On supported Wayland compositors, opening a capture and copying it with Ctrl+C can be substantially faster while keeping the normal editor and annotation workflow. This change makes the faster capture and clipboard backends independently opt-in; existing defaults remain in place.

Enable with `FLAMESHOT_USE_GRIM=1 FLAMESHOT_USE_WL_COPY=1 flameshot gui`, with `grim` and `wl-clipboard` installed. Direct PPM capture avoids PNG compression before opening the editor. The clipboard helper encodes the image once and keeps serving it after Flameshot exits, including repeated pastes. Helper failures fall back to the existing backends, with a three-second timeout per helper.

On one Hyprland 0.56.2 desktop with Qt 6.11.2 and two monitors at scales 1 and 1.6, three interleaved runs per mode on the same monitor gave these medians for an 800×500 selection:

| Measurement | Default backends | Both options enabled |
| --- | ---: | ---: |
| Launch to editor detected | 2,834 ms | 247 ms |
| Ctrl+C to first image read | 227 ms | 33 ms |

These are a small local sample on a live desktop, not a cross-compositor benchmark. The included `tests/wayland_performance.py` script reproduces the measurement and checks the copied dimensions and repeated pastes.

PPM increases temporary memory use. The clipboard path uses a private temporary file until `wl-copy` consumes it and preserves the JPEG and save-after-copy preferences. Desktop testing covered Hyprland only; support elsewhere depends on the protocols provided by the compositor. Existing monitor selection and mixed-DPI cropping are retained.

Related: #3859, #4498

Validation:

- Release builds succeeded with `USE_WAYLAND_CLIPBOARD=ON` and `OFF`.
- Changed C++ files passed clang-format 11.1.0 checks.
- Ctrl+C produced an 800×500 image and identical repeated pastes after editor exit on both monitors, including fractional scaling.
- Escape preserved the previous clipboard contents.
- Injected failures of `grim` and `wl-copy` successfully used the fallback backends.

For comparison, run `python3 tests/wayland_performance.py build/src/flameshot --screen 1` with and without `--fast`. Add `--cancel` or `--fast --fail-tool grim` / `--fast --fail-tool wl-copy` for the regression checks. The test replaces the clipboard and uses the current user preferences.

### DIFF
--- a/docs/UsageHyprlandSwayWlroots.md
+++ b/docs/UsageHyprlandSwayWlroots.md
@@ -37,6 +37,46 @@ hl.window_rule({
 ```
 
 
+## Optional faster capture and clipboard backends
+
+On compositors supported by `grim` and `wl-clipboard`, these opt-in backends
+avoid PNG compression before the editor opens and repeated image encoding
+during clipboard transfers:
+
+```sh
+FLAMESHOT_USE_GRIM=1 FLAMESHOT_USE_WL_COPY=1 flameshot gui
+```
+
+Install `grim` and `wl-clipboard` first. The normal editor stays available:
+select a region, annotate it, and press **Ctrl+C**. `--raw` is not needed.
+You can enable either backend independently. To apply the settings to captures
+started from the tray, start the Flameshot tray process with the same variables.
+
+`FLAMESHOT_USE_GRIM=1` captures an uncompressed PPM image at grim's default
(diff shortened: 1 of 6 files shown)


borgmanJeremy:
We intentionally use the portal backend for commonality

Issue #4182, opened 2025-08-19
Flameshot Windows 11 - Not saving last-picked color or tool size


NightHawkATL:
### Flameshot Version

<img width="178" height="293" alt="Image" src="https://github.com/user-attachments/assets/cf8702eb-5351-47b8-94ea-d310eb1ce927" />

### Installation Type

Using the ready-made package from Github Releases

### Operating System type and version

Windows 11 Pro 24H2

### Description

I just updated to the latest version of 13.1.0 in the hopes that it would be fixed since others are complaining about similar issues, but it hasn't.

As seen in this screenshot, I have chosen a green color for my tool color and I clicked the "Update" button. Then I used the tool to take a screenshot and it immediately chose the first color again (white) for the tool color.

<img width="438" height="565" alt="Image" src="https://github.com/user-attachments/assets/adaeba1f-599b-4cbf-a8fc-4764ccae6375" />

As with the tool size setting also, It will revert back after switching tools when the setting is changed.

<img width="222" height="391" alt="Image" src="https://github.com/user-attachments/assets/8ca87b32-ec20-4f3a-9b34-70b672c6e565" />

You can see that I had the tool set for "8" and then when I saved the screenshot for use, and then used flameshot to draw an arrow to the setting, it was back to "3".

<img width="257" height="378" alt="Image" src="https://github.com/user-attachments/assets/a053f44e-1bb4-4f7f-a64a-1b6dcbaa3679" />

### Steps to reproduce

install Flameshot on Windows 11, open configuration and make changes, go to use the tools and see the config changes reverted.

### Screenshots or screen recordings

See in-line images above.

### System Information

<img width="835" height="525" alt="Image" src="https://github.com/user-attachments/assets/f5133da2-d9f3-4445-876a-5ec831c3585a" />

NightHawkATL:
I was told to open my own issue since others couldn't replicate it in the "duplicated" issue.

borgmanJeremy:
Oh sorry, the color is covered in that other issue. This is for the size only? Did you read through the intended behavior of how the size is saved?

NightHawkATL:
This is for size and color.

The latest version and the one just before have changed and I never had to do anything before to save color and size.

mmahmoudian:
Is this partial duplicate of #4148?

borgmanJeremy:
The size part is unique to this issue . I asked for a new issue because I wasn't able to replicate the size issue.

TCROC:
[...]

Issue #4148, opened 2025-08-11
Tools no longer remember last used color


tintinsilva:
### Flameshot Version

Flameshot v13.0.1 ()
Compiled with Qt 6.9.1

### Installation Type

Linux (pacman)

### Operating System type and version

EndeavourOS (X11)

### Description

Tools used to remember the last color used in the Tool Settings. Now they just default to red, requiring manual input every time you'd want to use a specific color. I noticed this after updating to the latest version. 

### Steps to reproduce

1. Run `flameshot gui`
2. Go to Tool Settings, use the picker to pick a color
3. Use the rectangle tool (or text, or circle, etc.)
4. Save the edited screenshot, exit
5. When next running `flameshot gui` it will revert to the default color (red) instead of the color used in the previous session.

### System Information

EndeavourOS x86_64 (up to date); Qtile Wm (X11)

borgmanJeremy:
This bug was introduced in this bug fix: https://github.com/flameshot-org/flameshot/commit/a41137707ce2ebdc2a0be536f5cf872f35c950ce

The proper color is first loaded from the config file, then overridden with a default value when initPanel() is called. More work to find a fix.

NightHawkATL:
Can we also have it save the tool size as well? (Windows 11 Pro)

borgmanJeremy:
@NightHawkATL  that feature was added in v13 (also try to post feature requests on their own post not other bugs :-) )

NightHawkATL:
> [@NightHawkATL](https://github.com/NightHawkATL) that feature was added in v13 (also try to post feature requests on their own post not other bugs :-) )

Well the "save state" of the tool size isn't working after 13.0.1 and it is a part of the tool color not working as expected as well.

borgmanJeremy:
Huh okay let me look into that too

borgmanJeremy:
@NightHawkATL  I cannot replicate your issue on linux. If you are still having a problem please open a separate issue since it is not related (in root cause) to this issue. 

The saved size is shared across "sets" of tools. See this PR and this issue for more details:

https://github.com/flameshot-org/flameshot/pull/4011

https://github.com/flameshot-org/flameshot/issues/3812

tomtastic:
I see this too on MacOS with the latest `v13.1.0` release, color is defaulting to red (`#800000`) and ignoring last chosen color.

[...]

Issue #3079, opened 2023-01-31
MacOS: Active tool size is not saved


Dr4hcir:
### Flameshot Version

Flameshot v12.1.0 (96c2c82e)
Compiled with Qt 5.15.5
darwin: 22.2.0
osx: 13.1

### Installation Type

Linux, MacOS, or Windows Package manager (apt, pacman, eopkg, choco, brew, ...)

### Operating System type and version

MacOS 13.1

### Description

I draw a lot of squares around the stuff I want to be noticed in my screenshot. I want the borders to be more than 1 pixel thick, so I move the slider of 'Active tool size'. It looks like the size is saved, but every time I draw a new one, the thickness is back to 1 pixel. 

### Steps to reproduce

1. Set tool size large, so that a drawing of a rectangle is very thick
2. make a screenshot
3. Now, make a second screenshot
4. Select rectangle tool (s)
5 .hit 'space' to check settings

Expected: Active tool size is saved. It's still thick, so I expect my rectangle to be thick also

Result: Very narrow drawing. I have to actively move the 'Active tool size' every time I make a screenshot. As you see here the setting is thick, but the actual drawing is thin.

### Screenshots or screen recordings

![image](https://user-images.githubusercontent.com/26461357/215697855-359bb7dd-484a-4145-a9a1-058a847b4ba3.png)
![2023-01-31 08 51 23](https://user-images.githubusercontent.com/26461357/215699155-8bca6ca4-b5a1-482f-ac6b-9a1bc9c91d13.gif)

### System Information

Macbook Air M2 MacOS 13.1
![image](https://user-images.githubusercontent.com/26461357/215700035-c9a29d65-2390-43cd-9ab6-aa01512401eb.png)

geeksesi:
same here. what should we do?!

i'm using arch linux

jankatins:
Same here with a newer version of MacOSX:

Flameshot v12.1.0 (96c2c82e)
Compiled with Qt 5.15.5
darwin: 23.6.0
osx: 14.6

I had a very thin (1px) draw width for all annotation tools. I found out that I had a setting in `~/.config/flameshot/flameshot.ini` which hardcoded some draw size (don't remember the actual ini key) that to '1'. After removing it at least the default size is not 1px anymore. But it is still not updated / saved upon creating a new screenshot (create screenshot, annotate, adjust size, copy to clipboard) :-(

bterps:
[...]

PR #4961, opened 2026-09-19
Persist tool size when it changes, not when the notifier hides

bterps:
## Problem

Tool size is only written to the config file by `CaptureWidget::onToolSizeSettled()`, and its sole caller is the lambda connected to `NotifierBox::hidden` (`capturewidget.cpp:261-267`). `NotifierBox` hides on a 600 ms single-shot timer (`notifierbox.cpp:16-19`), so the size is persisted only if the capture window outlives that timer.

That produces two distinct failures:

**1. A timing race on the wheel / keyboard / button paths.** `wheelEvent`, the digit-key handler and the increase/decrease tool buttons all go through `CaptureWidget::setToolSize()`, which shows the notifier. Change the size and complete the capture within 600 ms — pressing Copy, for instance — and the change is discarded.

**2. The side panel never persists at all.** `SidePanelWidget::toolSizeChanged` is connected to `CaptureWidget::onToolSizeChanged` rather than `setToolSize` (`capturewidget.cpp:1274-1277`). `onToolSizeChanged` never touches `m_notifierBox`, so that path cannot reach `onToolSizeSettled()` no matter how long the user waits.

This is a regression from 534f05b2 ("Refactor tool size handling", #1969). Before it, the write was unconditional in `CaptureWidget::setDrawThickness()` on every size change, with no notifier involved:

```cpp
void CaptureWidget::setDrawThickness(int t)
{
    m_context.thickness = qBound(1, t, maxDrawThickness);
    // save draw thickness for text and other tool separately
    if (m_activeButton) {
        if (activeButtonToolType() == CaptureTool::TYPE_TEXT) {
            m_config.setDrawFontSize(m_context.thickness);
        } else {
            m_config.setDrawThickness(m_context.thickness);
        }
    }
    ...
}
```

That refactor split the function into `onToolSizeChanged()` (UI only) and a new `onToolSizeSettled()` (persistence only), then wired the latter into a lambda that #1841 had added purely to call `updateCursor()`.

## Fix

Move the write into `onToolSizeChanged()`, which every input path reaches — the wheel/keyboard/button paths via the `toolSizeChanged` self-connect at `capturewidget.cpp:254-258`, and the side panel directly at `1274-1277` — and drop `onToolSizeSettled()`. This restores the persist-on-change behaviour that predates #1969 while keeping that refactor's separation of UI updates from configuration.

`ConfigHandler::setToolSize()` also has to accept `CaptureTool::NONE`. Its final branch excluded it, so a write for a capture with no active tool button silently did nothing — while `ConfigHandler::toolSize()` already returns `drawThickness()` for `NONE`. Making the setter symmetric with the getter is what fixes the side panel case, since no tool button need be selected in order to move the slider.

No debounce is needed: the `CONFIG_SETTER` macro only calls `setValue()` when the value actually differs (`confighandler.h:41-50`), and `setValue()` does not call `sync()`, so `QSettings` batches the writes. This is the same write pattern the pre-#1969 code used.

## Reproduction

Side panel (never saved):
1. Take a screenshot, select the pencil.
2. Set the tool size with the side panel slider or spinbox.
3. Complete the capture, take another screenshot. The size has reverted.

Wheel/keyboard (saved only if you wait):
1. Take a screenshot, select the pencil.
2. Change the size with the mouse wheel.
3. Press Copy within 600 ms. The size has reverted. Waiting ~1 s before Copy saves it.

On a system where the config is an ini file, `~/.config/flameshot/flameshot.ini` can be watched directly — `drawThickness` and the file's mtime are both unchanged after the failing sequences above, including after quitting the application.

Reported in #3079 (macOS) and #4182 (Windows 11). #4148 covers the same shape for colour.

## Testing

Built and verified manually on macOS 26.6.2 (arm64), Qt 6.11.2, against an isolated `XDG_CONFIG_HOME` so the observation was unambiguous.

Instrumenting `onToolSizeChanged()` showed it being entered with each new size and with `activeButtonToolType()` returning `CaptureTool::NONE` (-1) on every call for the side panel path — which is how the `NONE` case above was found. With both changes in place:

| case | `drawThickness` before | after |
|---|---|---|
| side panel slider, no tool button selected, then Copy | 3 | 10 |
| tool selected, wheel, Copy immediately (inside the 600 ms window) | 3 | 11 |

Both were silently discarded before the change.

Not covered: the per-tool branches that route to a different key (e.g. Text → `drawFontSize`). Those branches are untouched by this patch, but I have not exercised them.

There are no existing tests covering `CaptureWidget`, tool size handling or `NotifierBox` (`tests/` contains two CLI-flag shell scripts), so I have not added one rather than introduce a new harness in this PR — happy to if you'd like it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


borgmanJeremy:
Thanks for the contribution, but we need patch authors to build and test thier commits.

bterps:
> Thanks for the contribution, but we need patch authors to build and test thier commits.

@borgmanJeremy No problem, do you have a formalized process for that or should I just do it and say it worked?

bterps:
I tested a local build and the fixes worked

